PR #4, opened 2021-06-04
Add support to numbers

alvalentini:


upf/action.py
```diff
@@ -88,5 +88,9 @@ def add_effect(self, fluent: Union[FNode, 'upf.Fluent'],
                    value: Union[FNode, 'upf.Fluent', 'upf.Object', ActionParameter, bool]):
         """Adds the given action effect."""
         [fluent_exp, value_exp] = self._env.expression_manager.auto_promote(fluent, value)
-        assert self._env.type_checker.get_type(fluent_exp) == self._env.type_checker.get_type(value_exp)
+        t_left = self._env.type_checker.get_type(fluent_exp)
+        t_right = self._env.type_checker.get_type(value_exp)
+        assert (t_left == t_right or (t_left.is_int_type() and t_right.is_int_type()) or
```
mikand, line 93: Maybe we can create a predicate `is_compatible_type(a, b)` that encapsulates this?
Also, instead of an assertion we can raise a `UPFTypeError` (inherited from python `TypeError`)

upf/expression.py
```diff
@@ -171,18 +188,61 @@ def Bool(self, value: bool) -> FNode:
...
+        """ Creates an expression of the form: left - right."""
+        [left, right] = self.auto_promote(left, right)
+        return self.create_node(node_type=op.MINUS, args=(left, right))
 
-    def ObjectExp(self, obj: 'upf.Object') -> FNode:
-        """Returns an expression for the given object."""
-        return self.create_node(node_type=op.OBJECT_EXP, args=tuple(), payload=obj)
+    def Times(self, left: Expression, right: Expression) -> FNode:
+        """ Creates an expression of the form: left * right."""
+        [left, right] = self.auto_promote(left, right)
```
mikand, line 217: No need to create a list:
```suggestion
        left, right = self.auto_promote(left, right)
```

upf/problem.py
```diff
@@ -93,7 +93,11 @@ def set_initial_value(self, fluent: Union[FNode, upf.Fluent],
                           value: Union[FNode, upf.Fluent, upf.Object, bool]):
         """Sets the initial value for the given fluent."""
         [fluent_exp, value_exp] = self._env.expression_manager.auto_promote(fluent, value)
-        assert self._env.type_checker.get_type(fluent_exp) == self._env.type_checker.get_type(value_exp)
+        t_left = self._env.type_checker.get_type(fluent_exp)
+        t_right = self._env.type_checker.get_type(value_exp)
+        assert (t_left == t_right or (t_left.is_int_type() and t_right.is_int_type()) or
```
mikand, line 98: Here you can reuse the `is_compatible_type` as before

upf/type_checker.py
```diff
@@ -82,3 +64,62 @@ def walk_object_exp(self, expression: FNode, args: List[upf.typing.Type]) -> upf
...
+        return self.env.type_manager().RealType(expression.value(), expression.value())
+
+    @walkers.handles(op.INT_CONSTANT)
+    def walk_identity_int(self, expression, args):
+        assert expression is not None
+        assert len(args) == 0
+        return self.env.type_manager().IntType(expression.value(), expression.value())
+
+    @walkers.handles(op.PLUS, op.MINUS, op.TIMES, op.DIV)
+    def walk_realint_to_realint(self, expression, args):
```
mikand, line 88: Here you are not propagating the lower- and upper- bounds.

mikand (changes requested):
Excellent work. Have a look at the few comments I left.

mikand (approved):

PR #8, opened 2021-06-10
Add problem tostring

alvalentini:


mikand (changes requested):
The PR is excellent, but we need some tests to stimulate these printing routines...

mikand (approved):

PR #12, opened 2021-06-15
Add a parallel solver

### LINKED ISSUE #9: Parallel Solving
### User Story
As a **TSB developer** I want to use multiple planners or more than one configuration of the same planner in parallel to solve a planning problem.

The parallel solving should have the same interface as the normal solving: each call to the solver is a sort of "voting system" that returns the result of the solver that answers first and the other solvers are stopped.

### Acceptance Criteria

- [ ] Easy to switch from using a single solver or a parallel solver
- [ ] One can solve a planning problem using two different planners
- [ ] One can solve a planning problem  using a planner in two configurations (e.g. 2 random seeds)

### Additional Material

- A similar tool is present in pysmt: https://github.com/pysmt/pysmt/blob/master/pysmt/solvers/portfolio.py#L50
- We can either use the python standard library (e.g. `async` keyword) or a dedicated library such as pebble: https://pebble.readthedocs.io/en/latest

### Attention Points

- Fragmentation issues: we shall find a way to avoid having too many classes and constructors (e.g. `OneShotSolver`, `ParallelOneShotSolver`, `Validator`, `ParallelValidator`).

alvalentini:
Closes #9

### DIFF
--- a/upf/parallel.py
+++ b/upf/parallel.py
@@ -1,0 +1,70 @@
+# Copyright 2021 AIPlan4EU project
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import upf
+from upf.solver import Solver
+from typing import Dict, List, Tuple
+from multiprocessing import Process, Queue
+
+
+class Parallel(Solver):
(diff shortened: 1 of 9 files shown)


upf/parallel.py
```diff
@@ -0,0 +1,93 @@
...
+# limitations under the License.
+#
+
+import upf
+from upf.solver import Solver
+from typing import Dict, List, Tuple
+from multiprocessing import Process, Queue
+
+
+class Parallel(Solver):
```
mikand, line 22: No docstring

upf/parallel.py
```diff
@@ -0,0 +1,93 @@
...
+    def __init__(self, solvers: List[Tuple[type, Dict[str, str]]]):
+        self.solvers = solvers
+
+    def is_oneshot_planner(self) -> bool:
+        return True
```
mikand, line 27:
```suggestion
        return all(Solver(name=s).is_oneshot_planner() for s in self.solvers)
```

This suggests that maybe is_xxx() could be a static method?

mikand (changes requested):
LGTM. Re-align with master and we are set.

PR #23, opened 2021-07-08
PDDL reader

### LINKED ISSUE #14: PDDL parsing
### User Story
As a **planning expert** I want to import existing PDDL benchmarks and models to augment them via code and/or use integrated planning techniques.

The parser should be able to read just a domain or a domain+problem and generate a `Problem` instance.

### Acceptance Criteria

- [ ] IPC classical and numerical planning examples can be successfully parsed
- [ ] A suite of erroneous model is created and added to CI
- [ ] Proper errors are reported to the user

### Additional Material

- We could exploit `tarski` that does offer a PDDL parser
- https://github.com/pucrs-automated-planning/pddl-parser
- https://github.com/pyparsing/pyparsing/
- https://github.com/swayf/grako#id3

### Attention Points

- PDDL has several dialects/fragments depending on the requirements (https://planning.wiki/ref/pddl21/domain) a reasonable subset needs to be identified

alvalentini:
Closes #14


upf/io/pddl_reader.py
```diff
@@ -0,0 +1,207 @@
...
+                return self.em.ObjectExp(objects[formula.name])
+            else:
+                raise UPFProblemDefinitionError(symbol + ' not supported!')
+        elif isinstance(formula, Variable):
+            assert formula.symbol in action_parameters
+            return self.em.ParameterExp(action_parameters[formula.symbol])
+        else:
+            raise UPFProblemDefinitionError(formula + ' not supported!')
+
+    def _convert_tarski_problem(self, tarski_problem: tarski.fstrips.Problem) -> upf.Problem:
```
mikand, line 109: Suggestion: maybe we can put this function into a dedicated package like `upf.interop.tarski` because it provides one way of supporting the interoperability with tarski. At that point, with an inverse function (i.e., upf.Problem to Tarski problem) we can support (and the user can exploit) various kinds of high-level interoperation

upf/io/pddl_reader.py
```diff
@@ -0,0 +1,207 @@
...
+        self.env = get_env(env)
+        self.em = self.env.expression_manager
+        self.tm = self.env.type_manager
+
+    def _convert_formula(self, fluents: Dict[str, upf.Fluent],
+                         objects: Dict[str, upf.Object],
+                         action_parameters: Dict[str, upf.ActionParameter],
+                         formula: Union[Formula, Term]) -> upf.fnode.FNode:
+        if is_and(formula):
+            children = [self._convert_formula(fluents, objects, action_parameters, f)
```
mikand, line 44: Not a fan of recursion over formulae, but OK for now

mikand (changes requested):
LGTM, but no tests are provided and the code can be organized as suggested in a comment.

mikand (approved):

PR #25, opened 2021-07-28
Closed world assumption

### LINKED ISSUE #17: Closed World Assumption
As a PDDL user, I would find it nice to have a way to express the initial state in CWA through UPF. Maybe we could make it optional, but, at least for propositional variables, this should be quite straightforward, and very handy especially for problems with a large number of objects. For instance, when there is the need to model when two positions are (not) connected.

alvalentini:
Closes #17

### DIFF
--- a/upf/problem.py
+++ b/upf/problem.py
@@ -229,9 +229,79 @@ class Problem:
         else:
             raise UPFProblemDefinitionError('Initial value not set!')
 
+    def domain_size(self, typename: upf.typing.Type) -> Optional[int]:
+        """Returns the domain size of the given type."""
+        if typename.is_bool_type():
+            return 2
+        elif typename.is_user_type():
+            return len(self.objects(typename))
+        elif typename.is_real_type():
+            lb = typename.lower_bound() # type: ignore
+            ub = typename.upper_bound() # type: ignore
+            if lb is not None and lb == ub:
+                return 1
+        elif typename.is_int_type():
+            lb = typename.lower_bound() # type: ignore
+            ub = typename.upper_bound() # type: ignore
+            if lb is not None and ub is not None:
+                return ub - lb
+        return None
+
+    def domain_item(self, typename: upf.typing.Type, idx: int) -> FNode:
+        """Returns the ith domain item of the given type."""
+        if typename.is_bool_type():
+            return self._env.expression_manager.Bool(idx == 0)
+        elif typename.is_user_type():
+            return self._env.expression_manager.ObjectExp(self.objects(typename)[idx])
+        elif typename.is_real_type():
+            lb = typename.lower_bound() # type: ignore
+            ub = typename.upper_bound() # type: ignore
+            assert (lb is not None and lb == ub)
+            return self._env.expression_manager.Real(lb)
+        elif typename.is_int_type():
+            lb = typename.lower_bound() # type: ignore
+            ub = typename.upper_bound() # type: ignore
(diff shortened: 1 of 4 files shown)


upf/problem.py
```diff
@@ -176,13 +197,88 @@ def set_initial_value(self, fluent: Union[FNode, upf.Fluent],
...
+            return self._initial_value[fluent_exp]
+        elif fluent_exp.fluent() in self._fluents_defaults:
+            return self._fluents_defaults[fluent_exp.fluent()]
+        elif fluent_exp.fluent().type() in self._initial_defaults:
+            return self._initial_defaults[fluent_exp.fluent().type()]
+        else:
             raise UPFProblemDefinitionError('Initial value not set!')
-        return self._initial_value[fluent_exp]
+
+    def domain_size(self, typename: upf.typing.Type) -> Optional[int]:
```
mikand, line 209: maybe make this private with an `_`?

upf/problem.py
```diff
@@ -176,13 +197,88 @@ def set_initial_value(self, fluent: Union[FNode, upf.Fluent],
...
+            if lb is not None and lb == ub:
+                return 1
+        elif typename.is_int_type():
+            lb = typename.lower_bound() # type: ignore
+            ub = typename.upper_bound() # type: ignore
+            if lb is not None and ub is not None:
+                return ub - lb
+        return None
+
+    def domain_item(self, typename: upf.typing.Type, idx: int) -> FNode:
```
mikand, line 227: same

upf/problem.py
```diff
@@ -176,13 +197,88 @@ def set_initial_value(self, fluent: Union[FNode, upf.Fluent],
...
+            return self._env.expression_manager.Real(lb)
+        elif typename.is_int_type():
+            lb = typename.lower_bound() # type: ignore
+            ub = typename.upper_bound() # type: ignore
+            assert (lb is not None and ub is not None)
+            return self._env.expression_manager.Int(lb + idx)
+        else:
+            raise
+
+    def get_ith_fluent_exp(self, fluent: upf.Fluent, domain_sizes: List[int], idx: int) -> FNode:
```
mikand, line 246: same

upf/problem.py
```diff
@@ -176,13 +197,88 @@ def set_initial_value(self, fluent: Union[FNode, upf.Fluent],
...
+                f_exp = self._env.expression_manager.FluentExp(f)
+                res[f_exp] = self.initial_value(f_exp)
+            else:
+                ground_size = 1
+                domain_sizes = []
+                is_groundable = True
+                for p in f.signature():
+                    ds = self.domain_size(p)
+                    if ds is None:
+                        is_groundable = False
```
mikand, line 273: Maybe we can simply assume that all the parameters of all the fluents will be groundable. Otherwise, this way of defining initial values does not work. No?

mikand: Also, the portions of the code dealing with INT and REAL parameters are currently totally untested, decreasing the coverage and complicating the code

mikand (commented):
All in all this is good: not sure about real parameters, I think we can safely remove them

mikand:
@alvalentini When you have time, can you rebase and have a look at my comments?

upf/problem.py
alvalentini: Do you mean to assume that all the fluents parameters have type UserType?

alvalentini: I updated the code assuming groundable parameters.

PR #26, opened 2021-08-04
Implemented arithmetic simplification

Framba-Luca:
Added plus, minus, times and div simplification with some basic testing


mikand (commented):
Still on ecomment missing :smile:

upf/simplifier.py
```diff
@@ -186,8 +189,138 @@ def walk_lt(self, expression: FNode, args: List[FNode]) -> FNode:
...
+        else: 
+            if len(new_args_times) == 0:
+                return self.manager.Int(1)
+            elif len(new_args_times) == 1:
+                return new_args_times[0]
+            else:
+                fnode_acc = new_args_times.pop(0)
+                for a in new_args_times:
+                    fnode_acc = self.manager.Times(a, fnode_acc)
+                return fnode_acc
```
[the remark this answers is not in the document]
Framba-Luca: Yes, it was into "hidden conversations" so I missed it. No it should be ok

mikand (approved):
LGTM! Excellent increase in coverage, btw!

upf/simplifier.py
```diff
@@ -13,11 +13,14 @@
 # limitations under the License.
 #
 
+from fractions import Fraction
+from collections import OrderedDict
 import upf.environment
 import upf.walkers as walkers
 import upf.operators as op
+from upf.shortcuts import *
```
alvalentini, line 21: This import should be not necessary.

upf/simplifier.py
```diff
@@ -28,6 +31,13 @@ def __init__(self, env: 'upf.environment.Environment'):
         self.env = env
         self.manager = env.expression_manager
 
+    def _number_to_fnode(self, value: Union[int, float, Fraction]) -> FNode:
+            if isinstance(value, int):
```
alvalentini, line 35: Fix the indentation.

upf/simplifier.py
```diff
@@ -186,8 +196,115 @@ def walk_lt(self, expression: FNode, args: List[FNode]) -> FNode:
...
+        #if accumulator != 0 create it as a constant FNode and then add all the non-constant FNodes found
+        #else return 0 or all the non-constant FNodes found
+        if accumulator != 0:
+            if isinstance(accumulator, int):
+                fnode_constant_values = self.manager.Int(accumulator)
+            else:
+                fnode_constant_values = self.manager.Real(Fraction(accumulator))
+            fnode_acc = fnode_constant_values
+            for a in new_args_plus:
+                fnode_acc = self.manager.Plus(a, fnode_acc)
```
alvalentini, line 224: Here you can create only a Plus node passing the full list of arguments.

upf/simplifier.py
```diff
@@ -186,8 +196,115 @@ def walk_lt(self, expression: FNode, args: List[FNode]) -> FNode:
...
+            return fnode_acc
+        else: 
+            if len(new_args_plus) == 0:
+                return self.manager.Int(0)
+            else:
+                fnode_acc = self.manager.Plus(new_args_plus)
+                return fnode_acc
```
alvalentini, line 231:
```suggestion
                return self.manager.Plus(new_args_plus)
```

upf/simplifier.py
```diff
@@ -186,8 +196,115 @@ def walk_lt(self, expression: FNode, args: List[FNode]) -> FNode:
...
+                fnode_constant_values = self.manager.Int(accumulator)
+            else:
+                fnode_constant_values = self.manager.Real(Fraction(accumulator))
+            fnode_acc = fnode_constant_values
+            for a in new_args_plus:
+                fnode_acc = self.manager.Plus(a, fnode_acc)
+            return fnode_acc
+        else: 
+            if len(new_args_plus) == 0:
+                return self.manager.Int(0)
```
alvalentini, line 228: Maybe here we can return `self.manager.Real(Fraction(0))` when at least one of the constant args was a real.
@mikand what do you think?

upf/simplifier.py
```diff
@@ -186,8 +196,115 @@ def walk_lt(self, expression: FNode, args: List[FNode]) -> FNode:
...
+
+    
+    def walk_times(self, expression: FNode, args: List[FNode]) -> FNode:
+        new_args_times: List[FNode] = list()
+        accumulator = 1
+        #divide constant FNode and accumulate their value into accumulator
+        for a in args:
+            if a.is_int_constant() or a.is_real_constant():
+                if a.constant_value() == 0:
+                    return self.manager.Int(0)
```
alvalentini, line 260: Same as above

upf/simplifier.py
```diff
@@ -186,8 +196,115 @@ def walk_lt(self, expression: FNode, args: List[FNode]) -> FNode:
...
+                else:
+                    accumulator = accumulator * a.constant_value()
+            elif a.is_times():
+                for s in a.args():
+                    if s.is_int_constant() or s.is_real_constant():
+                        if s.constant_value() == 0:
+                            return self.manager.Int(0)
```
alvalentini, line 267: Same as above

upf/simplifier.py
```diff
@@ -186,8 +196,115 @@ def walk_lt(self, expression: FNode, args: List[FNode]) -> FNode:
...
+            else:
+                new_args_times.append(a)
+        #if accumulator != 1 create it as a constant FNode and then add all the non-constant FNodes found
+        #else return  or all the non-constant FNodes found
+        if accumulator != 1:
+            fnode_acc = self._number_to_fnode(accumulator)
+            return self.manager.Times(*new_args_times, fnode_acc)
+        else: 
+            if len(new_args_times) == 0:
+                return self.manager.Int(1)
```
alvalentini, line 281: Same as above

upf/simplifier.py
```diff
@@ -186,8 +196,115 @@ def walk_lt(self, expression: FNode, args: List[FNode]) -> FNode:
...
+                value = int(left.constant_value() / right.constant_value())
+            else:
+                value = Fraction(left.constant_value(), right.constant_value())
+        elif (left.is_int_constant() or left.is_real_constant()) and (right.is_int_constant() or right.is_real_constant()):
+            assert(right.constant_value() != 0)
+            value = Fraction(left.constant_value(), right.constant_value())
+        else:
+            return self.manager.Div(left, right)
+        fnode_constant_values = self._number_to_fnode(value)
+        return fnode_constant_values
```
alvalentini, line 302:
```suggestion
        return self._number_to_fnode(value)
```

upf/simplifier.py
```diff
@@ -186,8 +196,115 @@ def walk_lt(self, expression: FNode, args: List[FNode]) -> FNode:
...
+        left, right = args
+        value : Union[Fraction, int, float] = 0
+        if left.is_int_constant() and right.is_int_constant():
+            if (left.constant_value() % right.constant_value()) == 0:
+                value = int(left.constant_value() / right.constant_value())
+            else:
+                value = Fraction(left.constant_value(), right.constant_value())
+        elif (left.is_int_constant() or left.is_real_constant()) and (right.is_int_constant() or right.is_real_constant()):
+            assert(right.constant_value() != 0)
+            value = Fraction(left.constant_value(), right.constant_value())
```
alvalentini, line 298: The Fraction arguments must be integers, but here at least one of the two is a Fraction too.
```suggestion
            value = Fraction(left.constant_value()) / Fraction(right.constant_value())
```

upf/simplifier.py
```diff
@@ -186,8 +196,115 @@ def walk_lt(self, expression: FNode, args: List[FNode]) -> FNode:
...
+        else:
+            return self.manager.Div(left, right)
+        fnode_constant_values = self._number_to_fnode(value)
+        return fnode_constant_values
+                
+        
+        
+
```
alvalentini, line 306: Use always the same number of line spaces between methods.

upf/test/test_simplifier.py
```diff
@@ -0,0 +1,560 @@
+    
+
+    # Copyright 2021 AIPlan4EU project
```
alvalentini, line 3:
```suggestion
# Copyright 2021 AIPlan4EU project
```

upf/test/test_simplifier.py
```diff
@@ -0,0 +1,560 @@
...
+        r4 = s.simplify(e4)
+        self.assertEqual(r4, t)
+
+    
+
+class TestArithmeticOperators(TestCase):
+    def setUp(self):
+        TestCase.setUp(self)
+
+
```
alvalentini: Remove the extra line spaces in this file.

upf/test/test_simplifier.py
```diff
@@ -0,0 +1,560 @@
...
+        x_2 = Div(x, fnode2)
+        data_list = [Fraction(5), Fraction(1, 5)]
+
+        fnode_of_data_list = Int(1)
+        for a in data_list:
+            fnode_of_data_list = Div(fnode_of_data_list, Real(Fraction(a)))
+        fnode_of_data_list = Div(x_2, fnode_of_data_list)
+        fnode_simplified = s.simplify(fnode_of_data_list)
+        #self.assertEqual(fnode_simplified, Div( Div(x, Int(3s)), Int(1)))
+        #self.assertEqual(fnode_simplified, Div( Div(x, Real(Fraction(3))), Real(Fraction(1))))
```
alvalentini, line 542: Remove commented code if not needed

upf/type_checker.py
```diff
@@ -46,7 +46,7 @@ def is_compatible_type(self, fluent_exp: FNode, value_exp: FNode) -> bool:
                 (t_left.is_real_type() and t_right.is_real_type()) or
                 (t_left.is_real_type() and t_right.is_int_type())):
             return False
-        left_lower = -float('inf') if t_left.lower_bound() is None else t_left.lower_bound() # type: ignore
+        left_lower = -float('inf') if t_left.lower_bound() is None else t_left.lower_bound() #type: ignore
```
alvalentini, line 49: Revert this.

upf/test/test_pddl_io.py
```diff
@@ -128,7 +128,7 @@ def test_robot_loader_adv_writer(self):
         self.assertTrue('l1 l2 l3 - Location' in pddl_problem)
         self.assertTrue('c1 - Container' in pddl_problem)
         self.assertTrue('(:init (robot_at r1 l1) (cargo_at c1 l2))' in pddl_problem)
-        self.assertTrue('(:goal (and (cargo_at c1 l3) (robot_at r1 l1)))' in pddl_problem)
+        self.assertIn('(:goal (and (cargo_at c1 l3) (robot_at r1 l1)))', pddl_problem)
```
alvalentini, line 131: Apply the same change to the assertion above

upf/type_checker.py
alvalentini, line 49: Revert this.

alvalentini (changes requested):
I added a few comments.
It is also needed to update the `walk_plus` and `walk_times` in the `ConverterToPDDLString` in file: https://github.com/aiplan4eu/upf/blob/master/upf/io/pddl_writer.py

upf/simplifier.py
mikand: Given the current typing system, that auto-promotes integers to reals, it should be safe to return an integer, no?

upf/simplifier.py
mikand: Nope, I like @Framba-Luca solution better, there is indeed a `Fraction` constructor with two arguments: https://docs.python.org/3/library/fractions.html#fractions.Fraction

alvalentini: You are right!

upf/simplifier.py
alvalentini: Yes, with the current typing system it is correct.
I was wondering if the proposed suggestion would be more robust to possible future changes in the typing system.

upf/io/pddl_writer.py
```diff
@@ -77,16 +79,16 @@ def walk_int_constant(self, expression, args):
         return str(expression.constant_value())
 
     def walk_plus(self, expression, args):
-        assert len(args) == 2
-        return f'(+ {args[0]} {args[1]})'
+        assert len(args) > 1
+        return f'(+ {" ".join(args)})'
```
alvalentini, line 83: In PDDL `+ - * /` are all binary operators, then if the args are greater than 2 you need to build something like:
`(+ args[0] (+ args[1] ..))`

upf/io/pddl_writer.py
```diff
@@ -77,16 +79,16 @@ def walk_int_constant(self, expression, args):
...
 
     def walk_minus(self, expression, args):
         assert len(args) == 2
         return f'(- {args[0]} {args[1]})'
 
     def walk_times(self, expression, args):
-        assert len(args) == 2
-        return f'(* {args[0]} {args[1]})'
+        assert len(args) > 1
+        return f'(* {" ".join(args)})'
```
alvalentini, line 91: Same here

alvalentini (changes requested):

upf/io/pddl_writer.py
mikand: @Framba-Luca tip: you can use the `reduce` function of python: https://docs.python.org/3/library/functools.html#functools.reduce

mikand: @alvalentini Nice catch!

upf/simplifier.py
alvalentini: @mikand so do you think it is better to keep `Int(0)`?

upf/io/pddl_writer.py
Framba-Luca: @mikand  Thanks for the suggestion:
I came up with something like this
`return reduce(lambda x, y: f'(+ {y} {x})', iter(args))
`

This should work, right? The test are ok but I guess this isn't really tested yet since the tests are made with binary + and *

alvalentini: Yes, it works!

mikand: @Framba-Luca Yes! (You can omit the `iter()` though): `return reduce(lambda x, y: f'(+ {y} {x})', args)` should work just fine

upf/simplifier.py
mikand: For the time being, I would keep `Int()`. If the type system changes we will adapt.

alvalentini (approved):

mikand:
@Framba-Luca This looks perfect now, but you added a `substituter.py` file by mistake :smiley_cat:

PR #28, opened 2021-08-05
Modified fnode printing method

Framba-Luca:
Modified get_nary_string method of fnode

### DIFF
--- a/upf/fnode.py
+++ b/upf/fnode.py
@@ -37,18 +37,15 @@ class FNode(object):
     # issue since, by default, equality coincides with identity.
     def __hash__(self) -> int:
         return self._node_id
-
+    
     def get_nary_expression_string(self, op: str, args: List['FNode']) -> str:
         p = []
-        first = True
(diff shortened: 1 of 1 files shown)


upf/fnode.py
```diff
@@ -37,18 +37,15 @@ def __init__(self, content: FNodeContent, node_id: int):
     # issue since, by default, equality coincides with identity.
     def __hash__(self) -> int:
         return self._node_id
-
+
```
alvalentini, line 40: Remove these trailing whitespaces.

To prevent them you can enable an option in your vscode (https://stackoverflow.com/questions/30884131/remove-trailing-spaces-automatically-or-with-a-shortcut).

alvalentini (changes requested):
I added only a minor comment.

upf/fnode.py
Framba-Luca: Thank you very much for the tip.

PR #29, opened 2021-08-05
Added an identity dag walker, a Substituter and some testing

Framba-Luca:
Added identity dagWalker as a support to the substituter


upf/walkers/identitydag.py
```diff
@@ -0,0 +1,78 @@
...
+    but the structure of the formula has to be kept.
+    """
+
+    def __init__(self, env: 'upf.environment.Environment', invalidate_memoization=False):
+        walkers.DagWalker.__init__(self, invalidate_memoization)
+        self.env = env
+        self.manager = env.expression_manager
+
+
+    def walk_and(self, args, **kwargs):
```
alvalentini, line 35: The following methods have the following sign:
```suggestion
    def walk_and(self, expression, args, **kwargs):
```

upf/walkers/identitydag.py
```diff
@@ -0,0 +1,78 @@
...
+#
+
+
+import upf.environment
+import upf.walkers as walkers
+import upf.operators as op
+from upf.fnode import FNode
+
+class IdentityDagWalker(walkers.DagWalker):
+    """This class traverses a formula and rebuilds it recursively
```
alvalentini, line 23:
```suggestion
    """This class traverses an expression and rebuilds it recursively
```

upf/walkers/identitydag.py
```diff
@@ -0,0 +1,78 @@
...
+    identically.
+    This could be useful when only some nodes needs to be rewritten
+    but the structure of the formula has to be kept.
```
alvalentini, line 26:
```suggestion
    but the structure of the expression has to be kept.
```

upf/walkers/identitydag.py
```diff
@@ -0,0 +1,78 @@
...
+        return self.manager.Div(args[0], args[1])
+
+    def walk_fluent_exp(self, expression: FNode, args: List[FNode]) -> FNode:
+        return self.manager.FluentExp(expression.fluent(), tuple(args))
+
+    @walkers.handles(op.IRA_OPERATORS)
+    @walkers.handles(op.CONSTANTS)
+    @walkers.handles(op.PARAM_EXP, op.OBJECT_EXP)
+    def walk_identity(self, expression: FNode, args: List[FNode]) -> FNode:
+        return expression
```
alvalentini, line 78: You have to define the walk method for all the operators: https://github.com/aiplan4eu/upf/blob/master/upf/operators.py

alvalentini (changes requested):

Framba-Luca:
I just saw I also have to change the walk_param_exp and the walk_object_exp of the IdWalker

upf/substituter.py
```diff
@@ -0,0 +1,78 @@
...
+        if len(substitutions) == 0:
+            return expression
+        new_substitutions: typing.Dict[FNode, FNode] = {}
+        new_keys = self.manager.auto_promote(substitutions.keys())
+        new_values = self.manager.auto_promote(substitutions.values())
+        new_substitutions = dict(zip(new_keys, new_values))
+        to_remove: typing.List[FNode] = []
+        for k, v in new_substitutions.items():
+            if not self.type_checker.is_compatible_type(k, v):
+                to_remove.append(k)
```
alvalentini, line 68: Here you can raise an exception.

upf/substituter.py
```diff
@@ -0,0 +1,78 @@
...
+        subs = {a -> c, (c & b) -> d, (a & b) -> c}
+        substitute(f, subs) = c
+
+        f = a
+        subs = {a -> c, c -> d}
+        substitute(f, subs) = c
+
+        f = a & b
+        subs = {a -> 5, b -> c}
+        substitute(f, subs) = a & c
```
alvalentini, line 51: Should be something like this, right?
```suggestion
        substitute(f, subs) = 5 & c
```

Framba-Luca: Actually in this case, since "a" must be a boolean, assigning 5 to it is just removed from the substitution dict.

However, since we are raising an exception when this happens and not simply removing it from the subst dictionary anymore, this documentation must also be changed.

alvalentini: Ah ok, yes!

upf/substituter.py
```diff
@@ -0,0 +1,77 @@
...
+import typing
+from upf.type_checker import TypeChecker
+
+class Substituter(IdentityDagWalker):
+    """Performs substitution into an expression """
+    def __init__(self, env: 'upf.environment.Environment'):
+        IdentityDagWalker.__init__(self, env, True)
+        self.env = env
+        self.manager = env.expression_manager
+        self.type_checker = TypeChecker(env)
```
mikand, line 33: Can't we use the type checker within the environment?

upf/substituter.py
```diff
@@ -0,0 +1,77 @@
...
+        new_substitutions = dict(zip(new_keys, new_values))
+        for k, v in new_substitutions.items():
+            if not self.type_checker.is_compatible_type(k, v):
+                raise UPFTypeError(
+                    f"The expression type of {str(k)} is not compatible with the given substitution {str(v)}")
+        return self.walk(expression, subs = new_substitutions)
+
+    @walkers.handles(op.ALL_TYPES)
+    def walk_replace_or_identity(self, expression: FNode, args: List[FNode], subs: dict = {}, **kwargs) -> FNode:
+        if expression in subs:
```
mikand, line 74: This can be more efficient:

```suggestion
        res = subs.get(expression, None)
        if res is not None:
            return res
```

upf/walkers/identitydag.py
```diff
@@ -0,0 +1,88 @@
...
+        return self.manager.Equals(args[0], args[1])
+
+    def walk_le( self, expression: FNode, args: List[FNode], **kwargs):
+        return self.manager.LE(args[0], args[1])
+
+    def walk_lt( self, expression: FNode, args: List[FNode], **kwargs):
+        return self.manager.LT(args[0], args[1])
+
+    def walk_plus( self, expression: FNode, args: List[FNode], **kwargs):
+        return self.manager.Plus(*args)
```
mikand, line 61: no need for `*` here: it works, but it forces an unneeded unpacking + repacking. Just `args` is enough

upf/walkers/identitydag.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def walk_times( self, expression: FNode, args: List[FNode], **kwargs):
+        return self.manager.Times(*args)
```
mikand, line 64: Same.

upf/substituter.py
```diff
@@ -0,0 +1,77 @@
...
+            return expression
+        new_substitutions: typing.Dict[FNode, FNode] = {}
+        new_keys = self.manager.auto_promote(substitutions.keys())
+        new_values = self.manager.auto_promote(substitutions.values())
+        new_substitutions = dict(zip(new_keys, new_values))
+        for k, v in new_substitutions.items():
+            if not self.type_checker.is_compatible_type(k, v):
+                raise UPFTypeError(
+                    f"The expression type of {str(k)} is not compatible with the given substitution {str(v)}")
+        return self.walk(expression, subs = new_substitutions)
```
mikand, line 70: This part is a bit too convoluted: why not simply iterate over `substitutions` and create `new_substitution and do the type check on-the-fly?

mikand (changes requested):
@Framba-Luca Excellent work! I left a few minor comments, but all in all this is almost perfect.

upf/substituter.py
Framba-Luca: @mikand  This idea is because auto_promote can take a list and gives always back a list, so calling it on one single argument would mean to create a list for just one argument.

I don't really know if this has a highly negative impact or it's something negligible.

Framba-Luca: One clever solution I can find is:
`
        new_keys = self.manager.auto_promote(substitutions.keys())
        new_values = self.manager.auto_promote(substitutions.values())
        for k, v in zip(new_keys, new_values):
               if self.type_checker.is_compatible_type(k, v):
                        new_substitutions[k] = v
                else:
                        raise UPFTypeError(...)
`

Framba-Luca: I actually missed the part "iterate over `substitution`", my brain red "iterate over `new_substitution`".

Ok!

alvalentini (approved):

upf/substituter.py
```diff
@@ -0,0 +1,80 @@
...
+        subs = {a -> b, a -> c}
+        substitute(f, subs) = c
+        """
+
+        if len(substitutions) == 0:
+            return expression
+        new_substitutions: Dict[FNode, FNode] = {}
+        for k, v in substitutions.items():
+            new_k, = self.manager.auto_promote(k)
+            new_v, = self.manager.auto_promote(v)
```
alvalentini, line 66:
```suggestion
            new_k, new_v = self.manager.auto_promote(k, v)
```

upf/substituter.py
```diff
@@ -0,0 +1,80 @@
...
+            if self.type_checker.is_compatible_type(new_k, new_v):
+                new_substitutions[new_k] = new_v
+            else:
+                raise UPFTypeError(
+                    f"The expression type of {str(k)} is not compatible with the given substitution {str(v)}")
+        return self.walk(expression, subs = new_substitutions)
+
+    @walkers.handles(op.ALL_TYPES)
+    def walk_replace_or_identity(self, expression: FNode, args: List[FNode], subs: dict = {}, **kwargs) -> FNode:
```
alvalentini, line 75:
```suggestion
    def walk_replace_or_identity(self, expression: FNode, args: List[FNode], subs: Dict[Expression, Expression] = {}, **kwargs) -> FNode:
```

upf/substituter.py
mikand: :+1:

alvalentini (changes requested):

mikand (approved):
Excellent. There is one minor leftover comment from @alvalentini, and then we can merge.

alvalentini (approved):

PR #30, opened 2021-08-06
Changed upf.typing with upf.types

Framba-Luca:
Due to name conflict from upf.typing and python typing library, we decided to change the name to upf.types.


mikand (approved):
LGTM. @Framba-Luca Can you rebase the branch on top of master to make sure it works with the latest changes on the substituter?

alvalentini (approved):

PR #32, opened 2021-08-09
ConditionalEffect & ConditionalEffectRemover

Framba-Luca:


upf/action.py
```diff
@@ -114,9 +132,15 @@ def add_precondition(self, precondition: Union[FNode, 'upf.Fluent', ActionParame
         self._preconditions.append(precondition_exp)
 
     def add_effect(self, fluent: Union[FNode, 'upf.Fluent'],
```
alvalentini, line 134: Here we can create `add_increase_effect` and `add_decrease_effect` methods

upf/action.py
```diff
@@ -114,9 +132,15 @@ def add_precondition(self, precondition: Union[FNode, 'upf.Fluent', ActionParame
...
-                   value: Union[FNode, 'upf.Fluent', 'upf.Object', ActionParameter, bool]):
+                   value: Union[FNode, 'upf.Fluent', 'upf.Object', ActionParameter, bool],
+                   condition: Union[FNode, BoolExpression] = True, kind: int = ASSIGNMENT):
         """Adds the given action effect."""
-        fluent_exp, value_exp = self._env.expression_manager.auto_promote(fluent, value)
+        fluent_exp, value_exp, condition_exp = self._env.expression_manager.auto_promote(fluent, value, condition)
+        if not (self._env.type_checker.get_type(condition_exp) == self._env.type_manager.BoolType()):
```
alvalentini, line 139:
```suggestion
        if not self._env.type_checker.get_type(condition_exp).is_bool_type():
```

upf/action.py
```diff
@@ -114,9 +132,15 @@ def add_precondition(self, precondition: Union[FNode, 'upf.Fluent', ActionParame
...
+            raise UPFTypeError('Action condition is not a Boolean condition!')
         if not self._env.type_checker.is_compatible_type(fluent_exp, value_exp):
             raise UPFTypeError('Action effect has not compatible types!')
-        self._effects.append((fluent_exp, value_exp))
+        if condition_exp == self._env.expression_manager.TRUE():
+            self._effects.append(Effect(fluent_exp, value_exp, condition = None, kind = kind))
```
alvalentini, line 144: We shuold use the same default for the condition, `True` or `None`.

upf/effect.py
```diff
@@ -0,0 +1,77 @@
...
+from upf.fnode import FNode
+from upf.exceptions import UPFTypeError
+from upf.expression import BoolExpression
+from collections import OrderedDict
+from typing import List, Union, Tuple, Optional
+
+KINDS = list(range(0, 3))
+
+(
+    ASSIGNMENT, INCREASE, DECREASE
```
alvalentini, line 33:
```suggestion
    ASSIGN, INCREASE, DECREASE
```

upf/io/pddl_writer.py
```diff
@@ -182,13 +184,31 @@ def _write_domain(self, out: IO[str]):
...
+                        if e.value().is_true():
+                            out.write(f' {converter.convert(e.fluent())}')
+                        elif e.value().is_false():
+                            out.write(f' (not {converter.convert(e.fluent())})')
+                        elif e.is_increase():
+                            out.write(f' (increase {converter.convert(e.fluent())} {converter.convert(e.value())})')
+                        elif e.is_decrease():
+                            out.write(f' (decrease {converter.convert(e.fluent())} {converter.convert(e.value())})')
+                        else:
+                            out.write(f' (assign {converter.convert(e.fluent())} {converter.convert(e.value())})')
```
alvalentini, line 211: Duplicated code

upf/shortcuts.py
```diff
@@ -23,10 +23,15 @@
...
 from upf.fnode import FNode
+from upf.effect import INCREASE, DECREASE
 from typing import Iterable, List, Union, Dict, Tuple, Optional
 from fractions import Fraction
 
 
+def Increase() -> int:
+    return INCREASE
+def Decrease() -> int:
+    return DECREASE
```
alvalentini, line 34: This is not needed

alvalentini (changes requested):
I added a few comments

upf/action.py
Framba-Luca: You are totally right. For now I am using True so that if a simple tautology is given as a condition, it can be simplified to True and the effect becomes unconditional.

upf/io/pddl_writer.py
```diff
@@ -182,13 +184,22 @@ def _write_domain(self, out: IO[str]):
...
             if len(a.effects()) > 0:
                 out.write('\n  :effect (and')
-                for k, v in a.effects():
-                    if v.is_true():
-                        out.write(f' {converter.convert(k)}')
-                    elif v.is_false():
-                        out.write(f' (not {converter.convert(k)})')
+                for e in a.effects():
+                    if e.is_conditional():
+                        out.write(f' (when ({converter.convert(e.condition())}')
```
alvalentini, line 189: Is it missing a bracket here?
```suggestion
                        out.write(f' (when ({converter.convert(e.condition())})')
```

Framba-Luca: I think it's the opposite, like so:
```suggestion
                        out.write(f' (when {converter.convert(e.condition())}')
```

Because the converter already puts brackets around every expression.

Framba-Luca: Very nice catch btw, you are so meticulous!!

upf/action.py
```diff
@@ -95,10 +98,32 @@ def preconditions(self) -> List[FNode]:
...
+        for e in self._effects:
+            if e.is_conditional():
+                conditional_effects.append(e)
+        return conditional_effects
+
+    def has_conditional_effects(self) -> bool:
+        for e in self._effects:
+            if e.is_conditional():
+                return True
+        return False
```
mikand, line 117: Simpler:
```suggestion
        return any(e.is_conditional() for e in self._effects)
```

upf/action.py
```diff
@@ -95,10 +98,32 @@ def preconditions(self) -> List[FNode]:
...
+
+    def unconditional_effects(self) -> List[Effect]:
+        """Returns the list of the action unconditional effects."""
+        unconditional_effects: List[Effect] = []
+        for e in self._effects:
+            if not e.is_conditional():
+                unconditional_effects.append(e)
+        return unconditional_effects
```
mikand, line 125: Simpler:
```suggestion
        return [e for e in self._effects if not e.is_conditional()]
```

upf/action.py
```diff
@@ -114,9 +139,39 @@ def add_precondition(self, precondition: Union[FNode, 'upf.Fluent', ActionParame
         self._preconditions.append(precondition_exp)
 
     def add_effect(self, fluent: Union[FNode, 'upf.Fluent'],
-                   value: Union[FNode, 'upf.Fluent', 'upf.Object', ActionParameter, bool]):
+                   value: Union[FNode, 'upf.Fluent', 'upf.Object', ActionParameter, bool],
+                   condition: Union[FNode, BoolExpression] = True):
```
mikand, line 143: Isn't `BoolExpresison` already a Union with `FNode`?

upf/action.py
```diff
@@ -114,9 +139,39 @@ def add_precondition(self, precondition: Union[FNode, 'upf.Fluent', ActionParame
...
         """Adds the given action effect."""
-        fluent_exp, value_exp = self._env.expression_manager.auto_promote(fluent, value)
+        fluent_exp, value_exp, condition_exp = self._env.expression_manager.auto_promote(fluent, value, condition)
+        assert fluent_exp.is_fluent_exp()
+        if not self._env.type_checker.get_type(condition_exp).is_bool_type():
+            raise UPFTypeError('Action condition is not a Boolean condition!')
```
mikand, line 148: Wrong message. This is an effect condition

upf/action.py
```diff
@@ -114,9 +139,39 @@ def add_precondition(self, precondition: Union[FNode, 'upf.Fluent', ActionParame
...
+        if not self._env.type_checker.is_compatible_type(fluent_exp, value_exp):
+            raise UPFTypeError('Action effect has not compatible types!')
+        self._effects.append(Effect(fluent_exp, value_exp, condition_exp))
+
+    def add_increase_effect(self, fluent: Union[FNode, 'upf.Fluent'],
+                   value: Union[FNode, 'upf.Fluent', 'upf.Object', ActionParameter, bool],
```
mikand, line 154: Isn't value an `Expression` type?

upf/action.py
```diff
@@ -114,9 +139,39 @@ def add_precondition(self, precondition: Union[FNode, 'upf.Fluent', ActionParame
...
+                   condition: Union[FNode, BoolExpression] = True):
+        """Adds the given action increase effect."""
+        fluent_exp, value_exp, condition_exp = self._env.expression_manager.auto_promote(fluent, value, condition)
+        assert fluent_exp.is_fluent_exp()
+        #Note that this assert is in conflict with the possibility of assigning x+3 with a fresh variable y while the substituter supports this.
```
mikand, line 159: OK, but I do not see the point of this comment. An effect always assigns a variable

upf/action.py
```diff
@@ -114,9 +139,39 @@ def add_precondition(self, precondition: Union[FNode, 'upf.Fluent', ActionParame
...
+            raise UPFTypeError('Action effect has not compatible types!')
+        self._effects.append(Effect(fluent_exp, value_exp, condition_exp, kind = INCREASE))
+
+    def add_decrease_effect(self, fluent: Union[FNode, 'upf.Fluent'],
+                   value: Union[FNode, 'upf.Fluent', 'upf.Object', ActionParameter, bool],
+                   condition: Union[FNode, BoolExpression] = True):
+        """Adds the given action increase effect."""
+        fluent_exp, value_exp, condition_exp = self._env.expression_manager.auto_promote(fluent, value, condition)
+        assert fluent_exp.is_fluent_exp()
+        #Note that this assert is in conflict with the possibility of assigning x+3 with a fresh variable y while the substituter supports this.
```
mikand, line 172: Same as above

upf/action.py
```diff
@@ -114,9 +139,39 @@ def add_precondition(self, precondition: Union[FNode, 'upf.Fluent', ActionParame
...
+        if not self._env.type_checker.get_type(condition_exp).is_bool_type():
+            raise UPFTypeError('Action condition is not a Boolean condition!')
+        if not self._env.type_checker.is_compatible_type(fluent_exp, value_exp):
+            raise UPFTypeError('Action effect has not compatible types!')
+        self._effects.append(Effect(fluent_exp, value_exp, condition_exp, kind = INCREASE))
+
+    def add_decrease_effect(self, fluent: Union[FNode, 'upf.Fluent'],
+                   value: Union[FNode, 'upf.Fluent', 'upf.Object', ActionParameter, bool],
```
mikand, line 167: Isn't value an `Expression` type?

upf/effect.py
```diff
@@ -0,0 +1,77 @@
...
+from typing import List, Union, Tuple, Optional
+
+KINDS = list(range(0, 3))
+
+(
+    ASSIGN, INCREASE, DECREASE
+) = KINDS
+
+class Effect:
+    def __init__(self, fluent: FNode, value: FNode, condition: FNode, kind: int = ASSIGN):
```
mikand, line 37: Maybe add a default value for `condition`

upf/problem.py
```diff
@@ -135,6 +135,13 @@ def add_action(self, action: upf.Action):
                 self._kind.set_conditions_kind('NEGATIVE_CONDITIONS') # type: ignore
             if op.OR in ops:
                 self._kind.set_conditions_kind('DISJUNCTIVE_CONDITIONS') # type: ignore
+        for e in action.effects():
```
mikand, line 138: You also need to update the taxonomy wiki: https://github.com/aiplan4eu/upf/blob/master/docs/wiki/Planning-Taxonomy.md

mikand (changes requested):
LGTM. I left a few suggestions and changes

upf/effect.py
Framba-Luca: Actually here, to add a default value we would need the environment. Since an Effect is always created in an Action, the Action already auto-promotes to FNode, but giving a default-FNode True value would mean to import the environment also in the Effect class to use the TRUE FNode of the ExpressionManager.

Now, this is not a problem, it can be easily imported, it's just a design choice. Let me know which option you like better.

(This matters only if the user creates an Effect without "using" the class Action, but the action doesn't have a method `
add_effect(Effect)
` so this doesn't seem quite useful to do.)

upf/problem.py
Framba-Luca: I opened another pull-request on the docs/wiki repository

upf/effect.py
mikand: Ah, right! OK, let's leave it like this.

Framba-Luca:
I merged the cond-effect-remover branch in this one so in 1 review you can check both files and then we can change this or merge.

upf/operators.py
```diff
@@ -14,17 +14,18 @@
 #
 """This module defines all the operators used by the UPF."""
 
-ALL_TYPES = list(range(0, 18))
+ALL_TYPES = list(range(0, 20))
 
 (
 AND, OR, NOT, IMPLIES, IFF,
+EXISTS, FORALL,
```
mikand, line 21: This does not belong to this PR :smile:

upf/conditional_effects_remover.py
```diff
@@ -0,0 +1,171 @@
...
+            new_problem.add_fluent(f)
+        for o in self._problem.all_objects():
+            new_problem.add_object(o)
+        for fl, v in self._problem.initial_values().items():
+            new_problem.set_initial_value(fl, v)
+        for g in self._problem.goals():
+            new_problem.add_goal(g)
+        for n, a in self._problem.actions().items():
+            if a.has_conditional_effects():
+                action_stack.append((n, a))
```
mikand, line 150: In python, it's more common to return a list rather than filling a list taken in input

mikand (changes requested):
@Framba-Luca thanks for the great work. Have a look at the algorithmic suggestion I left, it seems to me that the code can be greatly simplified.

upf/conditional_effects_remover.py
```diff
@@ -0,0 +1,171 @@
...
+                    new_problem.add_action(new_action_t)
+                    self._action_mapping[new_action_t.name()] = original_action_name
+                if self._check_and_simplify_preconditions(new_action_f):
+                    if len(new_action_f.effects()) > 0:
+                        new_problem.add_action(new_action_f)
+                        self._action_mapping[new_action_f.name()] = original_action_name
+            else:
+                action_stack.append((original_action_name, new_action_t))
+                action_stack.append((original_action_name, new_action_f))
+        return new_problem
```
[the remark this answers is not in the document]
Framba-Luca: Tomorrow I will implement this.

upf/conditional_effects_remover.py
```diff
@@ -46,43 +53,29 @@ def get_rewritten_problem(self) -> Problem:
         if self._unconditional_problem is not None:
             return self._unconditional_problem
         #cycle over all the actions
-        #Note that a different environment might be needed when multy-threading
-        action_stack: List[Tuple[str, Action]] = []
-        new_problem = self._create_problem_copy(action_stack)
+        #NOTE that a different environment might be needed when multy-threading
+        action_stack: List[Action] = []
```
mikand, line 57: Minor: no need for this initialization

upf/conditional_effects_remover.py
```diff
@@ -46,43 +53,29 @@ def get_rewritten_problem(self) -> Problem:
...
+        new_problem, action_stack = self._create_problem_copy()
 
         while len(action_stack) > 0:
```
mikand, line 60: Why not a `for` on the list of conditional actions?

mikand (approved):
LGTM! I left just a couple of minor comments!

upf/conditional_effects_remover.py
Framba-Luca: Right now we don't have a method that gives the conditional/unconditional actions of a problem.

I will resolve this asap

upf/conditional_effects_remover.py
```diff
@@ -0,0 +1,158 @@
...
+        return new_problem
+
+    def rewrite_back_plan(self, unconditional_sequential_plan: SequentialPlan) -> SequentialPlan:
+        '''Takes the sequential plan of the non-conditional problem (created with
+        the method "self.get_rewritten_problem()" and translates the plan back
+        to be a plan of the original problem.'''
+        uncond_actions = unconditional_sequential_plan.actions()
+        cond_actions = []
+        for ai in uncond_actions:
+            if ai.action().name() in self._action_mapping:
```
alvalentini, line 149:
```suggestion
            if ai.action() in self._action_mapping:
```

alvalentini (changes requested):

alvalentini (approved):

PR #34, opened 2021-08-10
Added ConditionalEffectRemover

Framba-Luca:


upf/action.py
```diff
@@ -163,3 +159,6 @@ def add_decrease_effect(self, fluent: Union[FNode, 'upf.Fluent'],
         if not self._env.type_checker.is_compatible_type(fluent_exp, value_exp):
             raise UPFTypeError('Action effect has not compatible types!')
         self._effects.append(Effect(fluent_exp, value_exp, condition_exp, kind = DECREASE))
+
+    def add_effect_class(self, effect: Effect):
```
alvalentini, line 163: I would rename this method `add_effect` and the original one in `add_assign_effect`

upf/conditional_effects_remover.py
```diff
@@ -0,0 +1,92 @@
...
+            new_action_f = self._create_action_copy(action, original_action_name, number, unchanged_effects)
+            number = number + 1
+            new_action_t.add_precondition(effect_changed.condition())
+            new_effect = Effect(effect_changed.fluent(), effect_changed.value(),
+                    self._env.expression_manager.TRUE(), effect_changed.kind())
+            new_action_t.add_effect_class(new_effect)
+            new_action_f.add_precondition(self._env.expression_manager.Not(effect_changed.condition()))
+            if new_action_t.has_conditional_effects():
+                action_stack.append((original_action_name, new_action_t, number))
+                action_stack.append((original_action_name, new_action_f, number+2))
```
alvalentini, line 55: Here I suppose is missing an `else` branch where you should add the action to the `new_problem`

upf/conditional_effects_remover.py
```diff
@@ -0,0 +1,92 @@
...
+            number = number + 1
+            new_action_f = self._create_action_copy(action, original_action_name, number, unchanged_effects)
+            number = number + 1
+            new_action_t.add_precondition(effect_changed.condition())
+            new_effect = Effect(effect_changed.fluent(), effect_changed.value(),
+                    self._env.expression_manager.TRUE(), effect_changed.kind())
+            new_action_t.add_effect_class(new_effect)
+            new_action_f.add_precondition(self._env.expression_manager.Not(effect_changed.condition()))
+            if new_action_t.has_conditional_effects():
+                action_stack.append((original_action_name, new_action_t, number))
```
alvalentini, line 54: Is it missing the `+1` here?
```suggestion
                action_stack.append((original_action_name, new_action_t, number+1))
```

upf/conditional_effects_remover.py
```diff
@@ -0,0 +1,92 @@
...
+
+        return new_problem
+
+    def _create_action_copy(self, action, original_action_name, number, unchanged_effects) -> Action:
+        new_action_name = original_action_name+ "_" +str(number)
+        new_parameters = OrderedDict()
+        for ap in action.parameters():
+            new_parameters[ap.name()] = ap.type()
+        new_action = Action(new_action_name, new_parameters, self._env)
+        self._action_mapping[new_action_name] = original_action_name
```
alvalentini, line 65: The action mapping can be created only for actions added to the new problem.

alvalentini (changes requested):
I added a few comment. After fixing them, you can add a few tests to check if the compiler works as expected.

upf/conditional_effects_remover.py
Framba-Luca: No it should be right, the number is incremented right after usage (after the `
_create_action_copy()` function. Actually the number increment could be done inside the function but then the function would have to return a tuple (int, Action) so I preferred to replicate 1 line of code instead of making it uselessly complex.

alvalentini: Ok, and in line 55 why is it needed to add 2 instead of 1?

upf/action.py
```diff
@@ -163,3 +159,6 @@ def add_decrease_effect(self, fluent: Union[FNode, 'upf.Fluent'],
         if not self._env.type_checker.is_compatible_type(fluent_exp, value_exp):
             raise UPFTypeError('Action effect has not compatible types!')
         self._effects.append(Effect(fluent_exp, value_exp, condition_exp, kind = DECREASE))
+
+    def add_effect(self, effect: Effect):
+        self._effects.append(effect)
```
alvalentini, line 164: After thinking a little about that, I think it is better to remove this method and keep only the original 3 methods to add new effects.

PR #35, opened 2021-08-10
Update Planning-Taxonomy

Framba-Luca:
Update Planning-Taxonomy to include conditional action effects


Framba-Luca:
Hi @mikand , I hope this is what you ment; otherwise let me know.

PR #36, opened 2021-08-12
Added quantifier support and implemented a QuantifierRemover class to "compile"away quantifiers from a complete problem instance

Framba-Luca:


upf/walkers/identitydag.py
```diff
@@ -48,6 +48,12 @@ def walk_implies( self, expression: FNode, args: List[FNode], **kwargs):
     def walk_iff( self, expression: FNode, args: List[FNode], **kwargs):
         return self.manager.Iff(args[0], args[1])
 
+    def walk_exists( self, expression: FNode, args: List[FNode], **kwargs):
+        return self.manager.Exists(args[0], *expression.variables())
+
+    def walk_forall( self, expression: FNode, args: List[FNode], **kwargs):
+        return self.manager.Exists(args[0], *expression.variables())
```
alvalentini, line 55:
```suggestion
        return self.manager.Forall(args[0], *expression.variables())
```

alvalentini (changes requested):
I added only one small comment. But now you need to update all the walkers.

upf/walkers/identitydag.py
Framba-Luca: Yep, with the next commit walkers should be ok!

alvalentini: In the last commit you didn't update the type checker and the pddl writer

upf/variable.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def type(self) -> upf.types.Type:
+        """Returns the variable type."""
+        return self._typename
+
+class FreeVarsOracle(walkers.DagWalker):
+    # We have only few categories for this walker.
+    #
+    # - Simple Args simply need to combine the cone/dependencies
+    #   of the children.
```
Framba-Luca, line 51:
```suggestion
   
```

upf/walkers/identitydag.py
```diff
@@ -48,6 +48,12 @@ def walk_implies( self, expression: FNode, args: List[FNode], **kwargs):
     def walk_iff( self, expression: FNode, args: List[FNode], **kwargs):
         return self.manager.Iff(args[0], args[1])
 
+    def walk_exists( self, expression: FNode, args: List[FNode], **kwargs):
+        return self.manager.Exists(args[0], *expression.variables())
+
+    def walk_forall( self, expression: FNode, args: List[FNode], **kwargs):
+        return self.manager.Forall(args[0], *expression.variables())
```
alvalentini, line 55:
```suggestion
    def walk_exists( self, expression: FNode, args: List[FNode], **kwargs) -> FNode:
        return self.manager.Exists(args[0], *expression.variables())

    def walk_forall( self, expression: FNode, args: List[FNode], **kwargs) -> FNode:
        return self.manager.Forall(args[0], *expression.variables())
```

upf/variable.py
```diff
@@ -0,0 +1,77 @@
...
+   
+    # - Quantifiers need to exclude bounded variables
+    # - Constants have no impact
+
+    def get_free_variables(self, expression: FNode) -> Set[Variable]:
+        """Returns the set of Symbols appearing free in the expression."""
+        return self.walk(expression)
+
+    @walkers.handles(op.VARIABLE_EXP)
+    def walk_variable_exp(self, expression: FNode, args: Set[Variable], **kwargs) -> Set[Variable]:
```
alvalentini, line 59: Here and below
```suggestion
    def walk_variable_exp(self, expression: FNode, args: List[Set[Variable]], **kwargs) -> Set[Variable]:
```

upf/variable.py
```diff
@@ -0,0 +1,77 @@
...
+        #pylint: disable=unused-argument
+        return {expression.variable()}
+
+    @walkers.handles(op.EXISTS, op.FORALL)
+    def walk_quantifier(self, expression: FNode, args: Set[Variable], **kwargs) -> Set[Variable]:
+        #pylint: disable=unused-argument
+        return args.difference(expression.variables())
```
alvalentini, line 66:
```suggestion
        return args[0].difference(expression.variables())
```

upf/variable.py
```diff
@@ -0,0 +1,77 @@
...
+
+    @walkers.handles(op.CONSTANTS)
+    def walk_constant(self, expression: FNode, args: Set[Variable], **kwargs) -> Set[Variable]:
+        #pylint: disable=unused-argument
+        return Set()
+
+    @walkers.handles(set(op.ALL_TYPES) - {op.VARIABLE_EXP, op.CONSTANTS, op.EXISTS, op.FORALL})
+    def walk_all(self, expression: FNode, args: Set[Variable], **kwargs) -> Set[Variable]:
+        return args
```
alvalentini, line 75: `args` is a list of set

alvalentini (changes requested):

upf/variable.py
Framba-Luca: Why is it a list? Isn't a set enough?

upf/io/pddl_writer.py
```diff
@@ -34,6 +34,34 @@ def convert(self, expression):
...
+    def walk_exists(self, expression, args):
+        assert len(args) == 1
+        vars = expression.variables()
+        vars_string = "("
+        for v in vars:
+            vars_string = vars_string + "?"
+            vars_string = vars_string + v.name()
+            vars_string = vars_string + " - "
+            vars_string = vars_string + str(v.type())
+        vars_string = vars_string + ")"
```
alvalentini, line 46: Use f-strings instead of string concatenation

upf/io/pddl_writer.py
```diff
@@ -34,6 +34,34 @@ def convert(self, expression):
...
+    def walk_forall(self, expression, args):
+        assert len(args) == 1
+        vars = expression.variables()
+        vars_string = "("
+        for v in vars:
+            vars_string = vars_string + "?"
+            vars_string = vars_string + v.name()
+            vars_string = vars_string + " - "
+            vars_string = vars_string + str(v.type())
+        vars_string = vars_string + ")"
```
alvalentini, line 58: Same here.

upf/problem_kind.py
```diff
@@ -21,9 +21,10 @@
 FEATURES = {
     'TIME' : ['CONTINUOUS_TIME', 'DISCRETE_TIME'],
     'NUMBERS' : ['CONTINUOUS_NUMBERS', 'DISCRETE_NUMBERS'],
-    'CONDITIONS_KIND' : ['NEGATIVE_CONDITIONS', 'DISJUNCTIVE_CONDITIONS', 'EQUALITY'],
+    'CONDITIONS_KIND' : ['NEGATIVE_CONDITIONS', 'DISJUNCTIVE_CONDITIONS', 'EQUALITY', 'EXISTENTIAL_PRECONDITIONS', 'UNIVERSAL_PRECONDITIONS'],
     'EFFECTS_KIND' : ['CONDITIONAL_EFFECTS', 'INCREASE_EFFECTS', 'DECREASE_EFFECTS'],
-    'TYPING' : ['FLAT_TYPING']
+    'TYPING' : ['FLAT_TYPING'],
+    'PRECONDITIONS_KIND' : []
```
alvalentini, line 27:
```suggestion
    'TYPING' : ['FLAT_TYPING']
```

alvalentini (changes requested):

upf/io/pddl_writer.py
Framba-Luca:
```

    def walk_exists(self, expression, args):
        assert len(args) == 1
        vars = expression.variables()
        vars_string_list = []
        for v in vars:
            vars_string = f"?{v.name()} - {str(v.type())}"
            vars_string_list.append(vars_string)
        return f'(exists ({" ".join(vars_string_list)})\n {args[0]})'

```

Let me know if this is ok. Looks more elegant to me.

upf/quantifiers_remover.py
```diff
@@ -0,0 +1,174 @@
...
+            new_problem.set_initial_value(fl, v)
+        return new_problem
+
+    def _action_without_quantifiers(self, action) -> Action:
+        #emulates a do-while loop: searching for an available name
+        is_unavailable_name = True
+        while is_unavailable_name:
+            new_action_name = action.name()+ "_" +str(self._counter)
+            self._counter = self._counter + 1
+            is_unavailable_name = self._problem.has_action(new_action_name)
```
mikand, line 139: Why do you need to rename the actions? The new actions are add to the `new_problem` so keeping their original name should not be a problem, right?

upf/quantifiers_remover.py
```diff
@@ -0,0 +1,174 @@
...
+        new_parameters = OrderedDict()
+        for ap in action.parameters():
+            new_parameters[ap.name()] = ap.type()
+        new_action = Action(new_action_name, new_parameters, self._env)
```
mikand, line 143:
```suggestion
        new_action = Action(new_action_name, OrderedDict((ap.name(), ap.type()) for ap in action.parameters()), self._env)
```

upf/quantifiers_remover.py
```diff
@@ -0,0 +1,174 @@
...
+        '''Takes the sequential plan of the problem (created with
+        the method "self.get_rewritten_problem()" and translates the plan back
+        to be a plan of the original problem.'''
+        uncond_actions = unconditional_sequential_plan.actions()
+        cond_actions = []
+        for ai in uncond_actions:
+            if ai.action() in self._action_mapping:
+                cond_actions.append(self._new_action_instance_original_name(ai))
+            else:
+                cond_actions.append(ai)
```
mikand, line 168: Is this case even possible?

upf/simplifier.py
```diff
@@ -154,6 +154,12 @@ def walk_implies(self, expression: FNode, args: List[FNode]) -> FNode:
         else:
             return self.manager.Implies(sl, sr)
 
+    def walk_exists(self, expression: FNode, args: List[FNode]) -> FNode:
+        return self.manager.Exists(args[0], *expression.variables())
```
mikand, line 158: Maybe handle the case where `args[0]` is true or false. Moreover, we can simplify the cases where `args[0]` does not contain some of the variables in `expression.variables()` by reducing the resulting set of variables.

E.g.
`Exists a, b, c . Or(b, c)` is equivalent to `Exists b, c . Or(b, c) `

upf/simplifier.py
```diff
@@ -154,6 +154,12 @@ def walk_implies(self, expression: FNode, args: List[FNode]) -> FNode:
...
+
+    def walk_forall(self, expression: FNode, args: List[FNode]) -> FNode:
+        return self.manager.Forall(args[0], *expression.variables())
```
mikand, line 161: Same as above

upf/test/test_conditional_effects_remover.py
```diff
@@ -73,18 +73,18 @@ def test_basic_conditional(self):
...
             self.assertEqual(str(plan), str(new_plan))
 
     def test_complex_conditional(self):
         problem = self.problems['complex_conditional'].problem
+        plan = self.problems['complex_conditional'].plan
         cer = ConditionalEffectsRemover(problem)
         unconditional_problem = cer.get_rewritten_problem()
 
-        with OneshotPlanner(name='enhsp') as planner:
+        with OneshotPlanner(problem_kind=unconditional_problem.kind()) as planner:
```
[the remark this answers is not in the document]
mikand: Maybe we can use a Validator to check plan validity instead of asserting that a specific plan is found?

mikand (changes requested):
@Framba-Luca Excellent work! I left some comments mostly to cleanup some minor code style issues. Apart for those this is excellent!

upf/fnode.py
```diff
@@ -39,7 +39,7 @@ def __init__(self, content: FNodeContent, node_id: int):
     def __hash__(self) -> int:
         return self._node_id
 
-    def get_nary_expression_string(self, op: str, args: List['FNode']) -> str:
+    def get_nary_expression_string(self, op: str, args: Union[List['FNode'], List['upf.Variable']]) -> str:
```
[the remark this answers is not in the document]
Framba-Luca: The Lists should be homogeneous, because they are only Variables if they are found as args of a quantifier or FNode (VariableExp) if not

upf/quantifiers_remover.py
Framba-Luca: Yes because the _action_mapping contains only the actions that had a quantifier.

upf/simplifier.py
Framba-Luca: How? The first solution that comes to my mind is that args is a:
List[Tuple[FNode, Set[Variable]]] and the function returns a Tuple[FNode, Set[Variable]]. What do you think about this?

upf/quantifiers_remover.py
```diff
@@ -0,0 +1,174 @@
...
+        type_list = []
+        name_list = []
+        for v in vars:
+            type_list.append(v.type())
+            name_list.append(v.name())
+        possible_objects: List[List[Object]] = []
+        for t in type_list:
+            possible_objects.append(self._problem.objects(t))
+        #product([1,2], [3,4]) = (1,3) (1,4) (2,3) (2,4)
+        possible_combinations = list(product(*possible_objects))
```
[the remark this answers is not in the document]
Framba-Luca: So i can just iterate over the product instead of on the list?

upf/quantifiers_remover.py
Framba-Luca: Yes this is totally true. This would also remove the need of a "write back" structure.

This was a remaining of the copy-paste from the ConditionalEffectsRemover

upf/quantifiers_remover.py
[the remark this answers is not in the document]
mikand: Yep!

Framba-Luca:
Here tests are failing because our plan validator/simplifier/substituter does not know how to simplify quantifiers to booleans. Any suggestion?

mikand:
I would also go for the second option! The right place where to put the support is clearly in the validator, possibly having a dedicated walker

Framba-Luca:
Ok I found a really "short" solution; using an EffectQuantifiersRemover (defined in quantifiers_remover.py) to compile away quantifiers before substitution/simplification. 
This makes me think that, instead of compiling every ActionInstance of the plan we could just call the QuantifiersRemover on the initial problem and then check if the plan is valid. 

Extending the simplifier would be "recursive" because, after the son of the Exists is simplified, we do a substitution with an or and we still have to simplify the created FNode, so this makes a simplifier call inside the simplifier.

upf/plan_validator.py
```diff
@@ -85,16 +230,12 @@ def _get_ground_fluent(self, fluent:FNode, assignments: Dict[Expression, Express
...
-        #This do-while loop is necessary because when we have a FluentExp with
-        #  some parameters, the first substitution substitutes the parameters with
-        #  the object: then every ground fluent is substituted with it's value.
-        #  It is a while loop because sometimes more than 2 substitutions can be
-        #  required.
-            old_exp = new_exp
-            new_exp = self._substituter.substitute(new_exp, assignments)
-        r = self._simplifier.simplify(new_exp)
+        r = self._qsimplifier.qsimplify(expression, assignments, {})
+        print("Inizio")
```
mikand, line 234: Remove these leftover debug prints

upf/simplifier.py
```diff
@@ -154,6 +155,26 @@ def walk_implies(self, expression: FNode, args: List[FNode]) -> FNode:
...
+        return self.manager.Exists(args[0], *vars)
+
+    def walk_forall(self, expression: FNode, args: List[FNode]) -> FNode:
+        assert len(args) == 1
+        if args[0].is_bool_constant():
+            if args[0].bool_constant_value():
+                return self.manager.TRUE()
+            return self.manager.FALSE()
+        free_vars: Set[Variable] = self.env.free_vars_oracle.get_free_variables(args[0])
+        vars = tuple(var for var in expression.variables() if var in free_vars)
```
mikand, line 175: Same as above

mikand (changes requested):
@Framba-Luca Excellent! I left the final comments: after those we can merge

mikand (approved):

PR #40, opened 2021-08-18
Plan validator

Framba-Luca:


haz:
Out of curiosity, why another beyond [any](https://github.com/KCL-Planning/VAL) [of](https://github.com/patrikhaslum/INVAL) [these](https://github.com/guicho271828/ArriVAL)?

Framba-Luca:
Hi @haz . The idea is to have a simple plan validator inside the upf to offer a capability without adding more dependencies on external features.

This the reason I know, when @alvalentini and @mikand return they might add some informations to your question.

upf/plan_validator.py
```diff
@@ -0,0 +1,68 @@
...
+        for ai in self._plan.actions():
+            for p in ai.action().preconditions():
+                ps = self._subs_simplify(p)
+                if not (ps.is_bool_constant() and ps.bool_constant_value()):
+                    return False
+            for e in ai.action().effects():
+                cond = True
+                if e.is_conditional():
+                    ec = self._subs_simplify(e.condition())
+                if not (ec.is_bool_constant() and ec.bool_constant_value()):
```
alvalentini, line 51: fix indentation

upf/plan_validator.py
```diff
@@ -0,0 +1,90 @@
...
+
+class PlanValidator(object):
+    """Performs plan validation."""
+    def __init__(self, env: 'upf.environment.Environment'):
+        self._env = env
+        self.manager = env.expression_manager
+        self.type_checker = env.type_checker
+        self._simplifier = Simplifier(self._env)
+        self._substituter = Substituter(self._env)
+        self._subst: Dict[Expression, Expression] = {}
```
alvalentini, line 36:
```suggestion
        self._assignments: Dict[FNode, FNode] = {}
```

upf/plan_validator.py
```diff
@@ -0,0 +1,90 @@
...
+        self._plan = plan
+        self._subst = self._problem.initial_values()
+        for ai in self._plan.actions():
+            new_subst: Dict[Expression, Expression] = {}
+            for ap, oe in zip(ai.action().parameters(), ai.parameters()):
+                self._subst[ap] = oe
+            for p in ai.action().preconditions():
+                ps = self._subs_simplify(p)
+                if not (ps.is_bool_constant() and ps.bool_constant_value()):
+                    print("a")
```
alvalentini, line 49: remove this print

upf/plan_validator.py
```diff
@@ -0,0 +1,90 @@
...
+        self._simplifier = Simplifier(self._env)
+        self._substituter = Substituter(self._env)
+        self._subst: Dict[Expression, Expression] = {}
+
+    def is_valid_plan(self, problem, plan) -> bool:
+        self._problem = problem
+        self._plan = plan
+        self._subst = self._problem.initial_values()
+        for ai in self._plan.actions():
+            new_subst: Dict[Expression, Expression] = {}
```
alvalentini, line 43:
```suggestion
            new_assignments: Dict[FNode, FNode] = {}
```

upf/plan_validator.py
```diff
@@ -0,0 +1,90 @@
...
+            gs = self._subs_simplify(g)
+            if not (gs.is_bool_constant() and gs.bool_constant_value()):
+                    return False
+        return True
+
+    def _get_ground_fluent(self, fluent:FNode) -> FNode:
+        assert fluent.is_fluent_exp()
+        new_args = []
+        for p in fluent.args():
+            new_args.append(self._subs_simplify(p)) #NOTE Might be better just to substitute once(?)
```
alvalentini, line 78: Remove the comment if not needed anymore

upf/plan_validator.py
```diff
@@ -0,0 +1,90 @@
...
+        return self.manager.FluentExp(fluent.fluent(), tuple(new_args))
+
+    def _subs_simplify(self, expression: FNode) -> FNode:
+        to_subst = True
+        new_exp = expression
+        while to_subst:
```
alvalentini, line 84: Here we can add a comment to explain why we need this loop

upf/test/test_plan_validator.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def test_robot(self):
+        problem, plan = self.problems['robot'].problem, self.problems['robot'].plan
+        pv = PlanValidator(get_env())
+        self.assertTrue(pv.is_valid_plan(problem, plan))
+
+    def test_complex_conditional(self):
+        problem, plan = self.problems['complex_conditional'].problem, self.problems['complex_conditional'].plan
+        pv = PlanValidator(get_env())
+        self.assertTrue(pv.is_valid_plan(problem, plan))
```
alvalentini, line 40: These tests are redundant

alvalentini (changes requested):
It looks good to me. I added just a few comments.

upf/plan_validator.py
Framba-Luca: Actually it was Dict[FNode, FNode], but mypy doesn't understand it is right.

alvalentini (approved):

upf/plan_validator.py
```diff
@@ -0,0 +1,91 @@
...
+        self.type_checker = env.type_checker
+        self._simplifier = Simplifier(self._env)
+
+        self._substituter = Substituter(self._env)
+
+    def is_valid_plan(self, problem, plan) -> bool:
+        assignments = problem.initial_values()
+        for ai in plan.actions():
+            new_assignments: Dict[Expression, Expression] = {}
+            for ap, oe in zip(ai.action().parameters(), ai.parameters()):
```
mikand, line 39: A suggestion: maybe we can rename the `parameters()` method of an action instance into `actual_parameters()` to clarify lines like this one?

```suggestion
            for ap, oe in zip(ai.action().parameters(), ai.actual_parameters()):
```

@alvalentini What do you think?

upf/plan_validator.py
```diff
@@ -0,0 +1,91 @@
...
+        new_exp = expression
+        while to_subst:
+        #This do-while loop is necessary because when we have a FluentExp with
+        #  some parameters, the first substitution substitutes the parameters with
+        #  the object: then every ground fluent is substituted with it's value.
+        #  It is a while loop because sometimes more than 2 substitutions can be
+        #  required.
+            old_exp = new_exp
+            new_exp = self._substituter.substitute(new_exp, assignments)
+            to_subst = not (old_exp == new_exp)
```
mikand, line 88:
```suggestion
        old_exp = None
        new_exp = expression
        while old_exp != new_exp:
        #This do-while loop is necessary because when we have a FluentExp with
        #  some parameters, the first substitution substitutes the parameters with
        #  the object: then every ground fluent is substituted with it's value.
        #  It is a while loop because sometimes more than 2 substitutions can be
        #  required.
            old_exp = new_exp
            new_exp = self._substituter.substitute(new_exp, assignments)
```

upf/plan_validator.py
```diff
@@ -0,0 +1,91 @@
...
+    def is_valid_plan(self, problem, plan) -> bool:
+        assignments = problem.initial_values()
+        for ai in plan.actions():
+            new_assignments: Dict[Expression, Expression] = {}
+            for ap, oe in zip(ai.action().parameters(), ai.parameters()):
+                assignments[ap] = oe
+            for p in ai.action().preconditions():
+                ps = self._subs_simplify(p, assignments)
+                if not (ps.is_bool_constant() and ps.bool_constant_value()):
+                    return False
```
mikand, line 44: This is correct, but we can maybe give out additional information like a string stating at which step and for which precondition the plan failed.

upf/plan_validator.py
```diff
@@ -0,0 +1,91 @@
...
+            for e in ai.action().effects():
+                cond = True
+                if e.is_conditional():
+                    ec = self._subs_simplify(e.condition(), assignments)
+                    if not (ec.is_bool_constant() and ec.bool_constant_value()):
+                        cond = False
```
mikand, line 50:
```suggestion
                    assert ec.is_bool_constant()
                    cond = ec.bool_constant_value()
```

upf/plan_validator.py
```diff
@@ -0,0 +1,91 @@
...
+                    elif e.is_decrease():
+                        new_assignments[ge] = self._subs_simplify(self.manager.Minus(e.fluent(),
+                                                e.value()), assignments)
+            assignments.update(new_assignments)
+            for ap in ai.action().parameters():
+                del assignments[ap]
+        for g in problem.goals():
+            gs = self._subs_simplify(g, assignments)
+            if not (gs.is_bool_constant() and gs.bool_constant_value()):
+                    return False
```
mikand, line 67: Also here we can report that the goal is not reached at the end of the plan.

mikand (changes requested):
@Framba-Luca Excellent. I left some minor suggestions here and there.

upf/plan_validator.py
alvalentini: I agree. `actual_parameters` is a better name.

upf/plan_validator.py
```diff
@@ -0,0 +1,100 @@
...
+        count = 0 #used for better error indexing
+        for ai in plan.actions():
+            count = count + 1
+            new_assignments: Dict[Expression, Expression] = {}
+            for ap, oe in zip(ai.action().parameters(), ai.actual_parameters()):
+                assignments[ap] = oe
+            for p in ai.action().preconditions():
+                ps = self._subs_simplify(p, assignments)
+                if not (ps.is_bool_constant() and ps.bool_constant_value()):
+                    self._last_error = f'Precondition: {p} of action number: {str(count)} of plan: {str(plan)} is not satisfied.'
```
mikand, line 48:
```suggestion
                    self._last_error = f'Precondition {p} of {str(count)}-th action instance {str(ai)} is not satisfied.'
```

upf/plan_validator.py
```diff
@@ -0,0 +1,100 @@
...
+                    elif e.is_decrease():
+                        new_assignments[ge] = self._subs_simplify(self.manager.Minus(e.fluent(),
+                                                e.value()), assignments)
+            assignments.update(new_assignments)
+            for ap in ai.action().parameters():
+                del assignments[ap]
+        for g in problem.goals():
+            gs = self._subs_simplify(g, assignments)
+            if not (gs.is_bool_constant() and gs.bool_constant_value()):
+                    self._last_error = f'Goal: {str(g)} of the problem is not reached.'
```
mikand, line 72:
```suggestion
                    self._last_error = f'Goal {str(g)} is not reached by the plan.'
```

mikand (changes requested):
@Framba-Luca I added some minor rephrasing, but note that tests are failing due to a typing error.

PR #41, opened 2021-08-19
Added a dnf-transformer

Framba-Luca:


upf/dnf.py
```diff
@@ -0,0 +1,202 @@
...
+    Negational Normal Form expression.
+
+    This is done first by removing all the Implications and Equalities,
+    then by pushing all the not to the leaves of the Tree representing the expression."""
+
+    def __init__(self, env: 'upf.environment.Environment'):
+        self.env = env
+        self.manager = env.expression_manager
+
+    def get_nnf_expression(self, expression: FNode) -> FNode:
```
alvalentini, line 38: Something like this should work and it is a little more clear:
```python
    def get_nnf_expression(self, expression: FNode) -> FNode:
        stack = []
        statuses = []
        stack.append((True, expression))
        statuses.append(False)
        solved = []
        while len(stack) > 0:
            p, e = stack.pop()
            status = statuses.pop()
            if status:
                if e.is_not():
                    pass
                elif e.is_and():
                    args = [solved.pop() for i in range(len(e.args()))]
                    if p:
                        new_e = self.manager.And(args)
                    else:
                        new_e = self.manager.Or(args)
                    solved.append(new_e)
                elif e.is_or():
                    args = [solved.pop() for i in range(len(e.args()))]
                    if p:
                        new_e = self.manager.Or(args)
                    else:
                        new_e = self.manager.And(args)
                    solved.append(new_e)
                else:
                    raise
            else:
                if e.is_not():
                    stack.append((p, e))
                    statuses.append(True)
                    stack.append((not p, e.arg(0)))
                    statuses.append(False)
                elif e.is_and() or e.is_or():
                    stack.append((p, e))
                    statuses.append(True)
                    for arg in e.args():
                        stack.append((p, arg))
                        statuses.append(False)
                elif e.is_implies():
                    stack.append((p, self.manager.Or(self.manager.Not(e.arg(0)), e.arg(1))))
                    statuses.append(False)
                elif e.is_iff():
                    stack.append((p, self.manager.Or(self.manager.And(e.arg(0), e.arg(1)),
                                                     self.manager.And(self.manager.Not(e.arg(0)),
                                                                      self.manager.Not(e.arg(1))))))
                    statuses.append(False)
                else:
                    if p:
                        solved.append(e)
                    else:
                        solved.append(self.manager.Not(e))
        return solved.pop()

```

Framba-Luca: Yes it should work! It is also possible to "skip" iterations for Implies and Iff, by decomposing in place the or, the and and the not, at the cost of readability. Like for Implies:

```suggestion
    def get_nnf_expression(self, expression: FNode) -> FNode:
        ....
            elif e.is_implies():
                    na1 = self.manager.Not(e.arg(0))
                    new_e = self.manager.Or(na1, e.arg(1))
                    stack.append((p, new_e, True))
                    stack.append((not p, e.arg(0), False))
                    stack.append((p, e.arg(1), False))
```

What do you think about that? It is also possible for Iff, based on what you like better we can choose the best option.

upf/dnf.py
```diff
@@ -0,0 +1,162 @@
...
+                    solved.append(new_e)
+                elif e.is_or():
+                    args = [solved.pop() for i in range(len(e.args()))]
+                    if p:
+                        new_e = self.manager.Or(args)
+                    else:
+                        new_e = self.manager.And(args)
+                    solved.append(new_e)
+                else:
+                    assert False
```
Framba-Luca, line 67: Here I don't know which Exception to raise. Should I use one we already have or create a new one?

alvalentini (approved):

upf/dnf.py
```diff
@@ -0,0 +1,179 @@
...
+from upf.exceptions import UPFUnraisableError
+from upf.walkers.dag import DagWalker
+from upf.fnode import FNode
+from typing import List, Tuple
+from itertools import product
+
+
+class Nnf():
+    """Class used to transform a logic expression into the equivalent
+    Negational Normal Form expression.
```
mikand, line 29:
```suggestion
    Negation Normal Form expression.
```

upf/dnf.py
```diff
@@ -0,0 +1,179 @@
...
+        For example, the form: !(a => (b && c)) becomes:
+        a && (!b || !c), therefore it is a NNF."""
+        stack: List[Tuple[bool, FNode, bool]] = []
+        stack.append((True, expression, False))
+        solved: List[FNode] = []
+        while len(stack) > 0:
+            p, e, status = stack.pop()
+            if status:
+                if e.is_and():
+                    args = [solved.pop() for i in range(len(e.args()))]
```
mikand, line 54:
```suggestion
                    args = [solved.pop() for _ in range(len(e.args()))]
```

mikand: @Framba-Luca this is very minor, but is useful to avoid nasty bugs when you use variable `i` after this line without re-initializing it (unfortunately it is valid in python). Instead `_` marks an unneeded value and cannot be reused

upf/dnf.py
```diff
@@ -0,0 +1,179 @@
...
+                    if p:
+                        new_e = self.manager.And(args)
+                    else:
+                        new_e = self.manager.Or(args)
+                    solved.append(new_e)
+                elif e.is_or():
+                    args = [solved.pop() for i in range(len(e.args()))]
```
mikand, line 61: same

upf/dnf.py
mikand: Just for information, there is even another way to do the same thing, but probably the repeated `pop()` is much clearer:
```suggestion
                    arity = len(e.args())
                    args = list(reversed(solved[-arity:]))
                    del solved[-arity:]
```

upf/dnf.py
```diff
@@ -0,0 +1,179 @@
...
+                    stack.append((p, e.arg(1), False))
+                    stack.append((p, e2, True))
+                    stack.append((not p, e.arg(0), False))
+                    stack.append((not p, e.arg(1), False))
+                else:
+                    if p:
+                        solved.append(e)
+                    else:
+                        solved.append(self.manager.Not(e))
+        return solved.pop()
```
mikand, line 108:
```suggestion
        assert len(solved) == 1 # Sanity check
        return solved.pop()
```

upf/dnf.py
```diff
@@ -0,0 +1,179 @@
...
+
+        For example, the form: !(a => (b && c)) becomes:
+        a && (!b || !c), in NNF form, and then:
+        (a && !b) || (a && !c), therefore a DNF expression."""
+        nnf_exp = self._nnf.get_nnf_expression(expression)
+        tuples = self.walk(nnf_exp)
+        and_list: List[FNode] = []
+        for and_args in tuples:
+            and_list.append(self.manager.And(and_args))
+        return self.manager.Or(and_list)
```
mikand, line 145: An iterator is an Iterable and both `And` and `Or` accept an Iterable as first argument, so you can simplify the code:
```suggestion
        return self.manager.Or(self.manager.And(and_args) for and_args in tuples)
```

upf/dnf.py
```diff
@@ -0,0 +1,179 @@
...
+                    break
+            if to_add:
+                res.append(nl)
+        return res
+
+    def walk_or(self, expression: FNode, args: List[List[List[FNode]]], **kwargs) -> List[List[FNode]]:
+        res: List[List[FNode]] = []
+        for a1 in args:
+            res.extend(a1)
+        return res
```
mikand, line 175: Simpler:
```suggestion
        return [conjunction for disjunction in args for conjunction in disjunction]
```

upf/dnf.py
```diff
@@ -0,0 +1,179 @@
...
+                            break
+                    else:
+                        if self.manager.Not(ne) in nl:
+                            to_add = False
+                            break
+                    if ne not in nl:
+                        nl.append(ne)
+                if not to_add:
+                    break
+            if to_add:
```
mikand, line 167: Too much cognitive complexity for 6pm :smiley: 
@Framba-Luca What is this piece of code trying to achieve?

General Feedback: this code is hard to read because of the nesting of ifs/fors and because variable names are not too descriptive.

mikand (changes requested):
LGTM except the and case of the Dnf class, what is the highlighted piece of code trying to achieve? For the rest, good work, I appreciated the iterative NNF a lot :smile:

upf/dnf.py
Framba-Luca: @mikand Ok, i tried to improve variables names and to add some comments with one example, let me know if it is clearer now.

mikand (changes requested):

upf/exceptions.py
```diff
@@ -24,3 +24,6 @@ class UPFProblemDefinitionError(UPFException):
 
 class UPFTypeError(UPFException, TypeError):
     pass
+
+class UPFUnraisableError(UPFException):
```
mikand, line 28: "There are only 10 hard things in computer science... naming things, binary numbers, and off-by-one errors." (adapted)
```suggestion
class UPFUnreachableCodeError(UPFException):
```

mikand (approved):
@Framba-Luca : Excellent! just rename the error as indicated and we can merge!

PR #42, opened 2021-09-01
Pyperplan

Framba-Luca:


alvalentini:
@Framba-Luca if you want to add pyperplan in the CI you need to install it like we did with tamer-upf. Take a look here: https://github.com/aiplan4eu/upf/blob/pyperplan/.github/workflows/main.yml

PR #46, opened 2021-09-13
Negative Preconditions Remover

Framba-Luca:


mikand:
@Framba-Luca Can you fix the tests?

PR #48, opened 2021-09-28
Protobuf & gRPC Support

uwe-koeckemann:
Added:
- protobuf message definitions mirroring upf objects
- factory to convert between protobuf and upf objects
- grpc server
- grpc client
- test cases


upf/test/test_grpc_server.py
```diff
@@ -0,0 +1,53 @@
...
+        self.server.start()
+        self.client = UpfGrpcClient("localhost", 30000)
+
+    def test_basic(self):
+        problem = self.ps['basic'].problem
+        plan_a = self.client(problem)
+        with OneshotPlanner(name='tamer', params={'weight': 0.8}) as planner:
+            plan_b = planner.solve(problem)
+
+        assert(str(plan_a) == str(plan_b)) # TODO: Should this work without str()?
```
mikand, line 27:
```suggestion
        self.assertEqual(str(plan_a), str(plan_b)) # TODO: Should this work without str()?
```

upf/test/test_grpc_server.py
```diff
@@ -0,0 +1,53 @@
...
+
+        self.ps = get_example_problems()
+        self.server = UpfGrpcServer(30000)
+        self.server.start()
+        self.client = UpfGrpcClient("localhost", 30000)
+
+    def test_basic(self):
+        problem = self.ps['basic'].problem
+        plan_a = self.client(problem)
+        with OneshotPlanner(name='tamer', params={'weight': 0.8}) as planner:
```
mikand, line 24: Maybe use a `Validator` instead of a `OneshotPlanner`?

upf/grpc/upf_pb2_grpc.py
```diff
@@ -0,0 +1,66 @@
+# Generated by the gRPC Python protocol compiler plugin. DO NOT EDIT!
```
mikand, line 1: Does it make sense to put auto-generated stuff in a dedicated package? Then we can import them with `__init__.py` to make this transparent

mikand (changes requested):
@uwe-koeckemann I left some comments,let me know if they make sense

upf/grpc/upf_pb2_grpc.py
uwe-koeckemann: Sounds reasonable. One issue I have with the protobuf generation right now is that it actually has to be modified for the imports to work.

upf/test/test_grpc_server.py
uwe-koeckemann: Does `Validator` create a plan? If there is a way to run tests without tamer it may be preferable.

mikand: It can check if a plan is valid (and we have a pure-python plan validator). See here: https://github.com/aiplan4eu/upf/blob/master/upf/test/test_validator.py#L32

upf/grpc/upf_pb2_grpc.py
uwe-koeckemann: Moved to sub-module called `generated`.

uwe-koeckemann:
I have two issues:
- mypy works locally (fresh python 3.7, 3.8, 3.9 environments) but keeps failing in the check above. 
- test cases work locally only for python 3.7. Which python version should I assume for my virtual environment?

mikand:
@uwe-koeckemann I should have fixed the issues. Basically, we had a problem with the CI/CD using an "external" `mypy` that was not sensitive to the `dev-requirements.txt` and I needed to rebase the branch to fix some minor issues caused by the UPF refactoring

uwe-koeckemann:
> @uwe-koeckemann I left some comments,let me know if they make sense

Thank you! I think I addressed all of them.

uwe-koeckemann:
> @uwe-koeckemann I should have fixed the issues. Basically, we had a problem with the CI/CD using an "external" `mypy` that was not sensitive to the `dev-requirements.txt` and I needed to rebase the branch to fix some minor issues caused by the UPF refactoring

Ah, good to know it was not on my end. It still says "change requested" above. I think I addressed all the things you pointed out, but I am not sure how to convince github of this... :)

upf/grpc/upf.proto
```diff
@@ -0,0 +1,63 @@
...
+}
+
+message Payload {
+    string type = 1;
+    string value = 2;
+}
+
+message Action {
+    string name = 1;
+    repeated string parameters = 2;
```
arbimo, line 32: Instead of two fields `parameters` and `parameterTypes`, shouldn't this be a single sequence 
`repeated Parameter parameters`, where `Parameter contains the `name` and the `type`. Note that `Parameter` would be the same as `Object` but with a different semantic. It might evolve into something different in the future.

This enforces the fact that each parameter is given a type.

upf/grpc/upf.proto
```diff
@@ -0,0 +1,63 @@
...
+message ActionInstance {
+    Action action = 1;
+    repeated Expression parameters = 2;
+}
+
+message SequentialPlan {
+    repeated ActionInstance actions = 1;
+}
+
+message Answer {
```
arbimo, line 56: With this formulation, we are forced to provide a plan regardless of the status.
`Answer` should probably be a `Oneof` where we have one possibility for each kind of answer : Sat(Plan), Unsat, Aborted(Cause)...

Here `Cause` could be an enum with the different sources of interuption (timeout, interrepted, internal error, ...)

upf/grpc/upf.proto
```diff
@@ -0,0 +1,63 @@
...
+    repeated string signature = 3;
+}
+
+message Object {
+    string name = 1;
+    string type = 2; 
+}
+
+message Expression {
+    int64 type = 1;
```
arbimo, line 15: What does this `type` represent?
At this point it is unclear to me whether we want to have `Expression` be isomorphic to an s-expresssion or if we want to go further than that.

upf/grpc/upf.proto
```diff
@@ -0,0 +1,63 @@
...
+    Expression x = 1;
+    Expression v = 2;
+}
+
+message Payload {
+    string type = 1;
+    string value = 2;
+}
+
+message Action {
```
arbimo, line 30: The `Action` name is probably to generic given that it is limited to classical/instantaneous actions.
Here me might want to have a unified view of an action or split it into different subclasses (InstantaneousAction, DurativeAction).

arbimo (commented):
Just a few comments after a quick look a the protobuf specification.

PR #51, opened 2021-10-01
Add temporal planning

alvalentini:


.github/workflows/main.yml
```diff
@@ -31,11 +31,11 @@ jobs:
         with:
           repository: aiplan4eu/tamer-upf
           path: tamer-upf
-          ref: "1887e041bcb5f6541fa0710481d9c3581458a86e"
+          ref: "7966608ea53d1b5c57f70cbbca8f89b579548a4f"
```
alvalentini, line 34:
```suggestion
          ref: "97d1394ceebc00c8e4effcc10f987bc33ace51ce"
```

upf/problem.py
```diff
@@ -107,89 +135,102 @@ def add_fluent(self, fluent: upf.Fluent, *,
...
+            for c in action.preconditions():
+                self._update_problem_kind_condition(c)
+            for e in action.effects():
+                self._update_problem_kind_effect(e)
+        elif isinstance(action, upf.DurativeAction):
+            for i in action.durative_conditions().keys():
+                if i.lower().bound() != 0:
+                    self._kind.set_time('ICE') # type: ignore
+                if i.upper().bound() != 0:
+                    self._kind.set_time('ICE') # type: ignore
```
alvalentini, line 175:
```suggestion
            for i, l in action.durative_conditions().items():
                if i.lower().bound() != 0 or i.upper().bound() != 0:
                    self._kind.set_time('ICE') # type: ignore
                for c in l:
                    self._update_problem_kind_condition(c)
```

alvalentini:
We should rename `ConstantTiming` with `AbsoluteTiming`

upf/io/pddl_writer.py
```diff
@@ -157,6 +162,8 @@ def _write_domain(self, out: IO[str]):
                 out.write(' :existential-preconditions')
             if self.problem.kind().has_universal_preconditions(): # type: ignore
                 out.write(' :universal-preconditions')
+            if self.problem.kind().has_continuous_time(): # type: ignore
```
alvalentini, line 165:
```suggestion
            if (self.problem.kind().has_continuous_time() or # type: ignore
                self.problem.kind().has_discrete_time()): # type: ignore
```

upf/io/pddl_writer.py
```diff
@@ -124,13 +127,15 @@ def walk_equals(self, expression, args):
...
 class PDDLWriter:
-    """This class can be used to write a Problem in PDDL."""
+    '''This class can be used to write a Problem in PDDL.'''
 
     def __init__(self, problem: 'upf.Problem', needs_requirements: bool = True):
         self.problem = problem
         self.needs_requirements = needs_requirements
 
     def _write_domain(self, out: IO[str]):
+        if self.problem.kind().has_ice() or self.problem.kind().has_maintain_goals(): # type: ignore
```
alvalentini, line 137: Here we need to add also `timed_goals` and `timed_effects`

upf/io/pddl_writer.py
```diff
@@ -124,13 +127,15 @@ def walk_equals(self, expression, args):
...
+            raise UPFProblemDefinitionError('PDDL2.1 does not support ICE or maintained goals.')
```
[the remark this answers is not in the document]
alvalentini: I agree. It should be fixed also according to the comment I left

PR #55, opened 2021-10-11
Temporal adapted quantifiers-remover

Framba-Luca:


upf/quantifiers_remover.py
```diff
@@ -123,25 +138,51 @@ def _create_problem_copy_without_actions_and_goals(self):
             new_problem.set_initial_value(fl, v)
         return new_problem
 
+    def _durative_action_without_quantifiers(self, action) -> DurativeAction:
+        new_action = DurativeAction(action.name(), OrderedDict((ap.name(), ap.type()) for ap in action.parameters()), self._env)
+        new_action.set_duration_constraint(action.duration())
+
+        for t, cl in action.conditions().items():
```
mikand, line 145: You are not using `t`, right?
```suggestion
        for cl in action.conditions().values():
```

upf/quantifiers_remover.py
```diff
@@ -123,25 +138,51 @@ def _create_problem_copy_without_actions_and_goals(self):
...
+            for c in cl:
+                nc = self._expression_quantifier_remover.remove_quantifiers(c, self._problem)
+                new_action.add_condition(t, nc)
+        for i, cl in action.durative_conditions().items():
```
mikand, line 149: Same

upf/quantifiers_remover.py
```diff
@@ -123,25 +138,51 @@ def _create_problem_copy_without_actions_and_goals(self):
...
+            for c in cl:
+                nc = self._expression_quantifier_remover.remove_quantifiers(c, self._problem)
+                new_action.add_durative_condition(i, nc)
+        for t, el in action.effects().items():
```
mikand, line 153: Same

upf/test/test_quantifier_remover.py
```diff
@@ -111,3 +111,19 @@ def test_robot_locations_visited(self):
...
             new_plan = qr.rewrite_back_plan(uq_plan)
             self.assertEqual(str(plan), str(new_plan))
+
+    def test_timed_connected_locations(self):
+        problem = self.problems['timed_connected_locations'].problem
+        plan = self.problems['timed_connected_locations'].plan
+        qr = QuantifiersRemover(problem)
+        uq_problem = qr.get_rewritten_problem()
+        # Temporal problem still missing a __repr__ function
+        # self.assertIn("Exists", str(problem))
```
mikand, line 121: Can't we add an `has_quantifiers` method to check this instead of relying on the repr?

mikand (commented):
LGTM except for the few comments I left

upf/quantifiers_remover.py
Framba-Luca: Actually yes, it's used in line 148. In general every time t (or i) are used when "putting back" the condition/effect in the new action

upf/test/test_quantifier_remover.py
Framba-Luca: Could also be a viable option to use the` problem_kind EXISTENTIAL_PRECONDITIONS` , which option do you like better?

upf/quantifiers_remover.py
mikand: Ah! I missed it. Ignore the comment

upf/test/test_quantifier_remover.py
mikand: Both are fine. You an implement a method `has quantifiers` by reading the problem kind

Framba-Luca: Perfect! I am happy we had the same idea.

Issue #61, opened 2021-10-22
Name of the fresh actions created by the transformers


Framba-Luca:
The transformers create a fresh action with a naming convention where `__` is reserved, but this method can be improved for better readability.
Mikand suggestions:

Ok, clear. This should work but maybe we can:

1)    Encapsulate the naming templates so that we can ask for fresh action/fluent names
2)    Follow C convention any symbol starting with `__` is reserved, internal underscores do not make a difference
3)    Insert in fluent/action names the name of the transformer (or an abbreviation): e.g. (a() -> __dnf_a_1()) so that it becomes a bit more readable

PR #65, opened 2021-11-10
Transformers-revision

Framba-Luca:
Closing #61


upf/transformers/conditional_effects_remover.py
```diff
@@ -95,14 +95,13 @@ def get_rewritten_problem(self) -> Problem:
...
             elif isinstance(action, DurativeAction):
                 timing_cond_effects: Dict[Timing, List[Effect]] = action.conditional_effects()
                 cond_effects_timing: List[Tuple[Effect, Timing]] = [(e, t) for t, el in timing_cond_effects.items() for e in el]
                 for p in self.powerset(range(len(cond_effects_timing))):
-                    nda = self._create_durative_action_copy(action, self._counter)
-                    self._durative_action_add_conditions(action, nda)
-                    self._durative_action_add_durative_conditions(action, nda)
+                    nda = action.clone()
+                    nda.name = self._get_fresh_action_name(action)
+                    nda._effects = {}
```
alvalentini, line 104:
```suggestion
                    nda.clear_effects()
```

upf/transformers/conditional_effects_remover.py
```diff
@@ -65,19 +64,20 @@ def get_rewritten_problem(self) -> Problem:
...
-        self._new_problem_add_goals()
+                    self._new_problem._add_effect_instance(t, e)
+        self._new_problem.clear_actions()
         for ua in self._problem.unconditional_actions():
             self._new_problem.add_action(ua)
+            self._new_to_old[ua] = ua
+            self._map_old_to_new_action(ua, ua)
         for action in self._problem.conditional_actions():
-            self._counter = 0
+            self._count = 0
```
alvalentini, line 74: This attribute should be hidden inside the base class

upf/transformers/conditional_effects_remover.py
```diff
@@ -65,19 +64,20 @@ def get_rewritten_problem(self) -> Problem:
                         nv = self._simplifier.simplify(em.Or(em.And(c, v), em.And(em.Not(c), f)))
                         self._new_problem.add_timed_effect(t, f, nv)
                 else:
-                    self._new_problem.add_timed_effect(t, e)
-        self._new_problem_add_timed_goals()
-        self._new_problem_add_maintain_goals()
-        self._new_problem_add_goals()
+                    self._new_problem._add_effect_instance(t, e)
```
alvalentini, line 67: Here we are passing the effect of the original problem, we should clone it.

upf/transformers/conditional_effects_remover.py
```diff
@@ -65,19 +64,20 @@ def get_rewritten_problem(self) -> Problem:
                         nv = self._simplifier.simplify(em.Or(em.And(c, v), em.And(em.Not(c), f)))
                         self._new_problem.add_timed_effect(t, f, nv)
```
alvalentini, line 65: It's wrong to use `f` here. We should use `e.fluent()`

upf/transformers/conditional_effects_remover.py
```diff
@@ -95,14 +95,13 @@ def get_rewritten_problem(self) -> Problem:
...
                 cond_effects_timing: List[Tuple[Effect, Timing]] = [(e, t) for t, el in timing_cond_effects.items() for e in el]
                 for p in self.powerset(range(len(cond_effects_timing))):
-                    nda = self._create_durative_action_copy(action, self._counter)
-                    self._durative_action_add_conditions(action, nda)
-                    self._durative_action_add_durative_conditions(action, nda)
+                    nda = action.clone()
+                    nda.name = self._get_fresh_action_name(action)
+                    nda._effects = {}
                     for t, e in action.unconditional_effects():
                         nda._add_effect_instance(t, e)
```
alvalentini, line 106: Same as above

upf/transformers/disjunctive_conditions_remover.py
```diff
@@ -35,34 +32,30 @@ class DisjunctiveConditionsRemover(Transformer):
...
+        self._new_problem.clear_actions()
         self._handle_actions()
 
         return self._new_problem
 
     def _handle_actions(self):
         dnf = Dnf(self._env)
         for a in self._problem.actions().values():
-            self.count = 0
+            self._count = 0
```
alvalentini, line 58: Same as above

upf/transformers/disjunctive_conditions_remover.py
```diff
@@ -73,65 +66,82 @@ def _handle_actions(self):
...
                 for co in c.args():
-                    self._add_condition(new_action, t, co)
+                    new_action.add_durative_condition(i, co)
             else:
-                self._add_condition(new_action, t, c)
-        self._durative_action_add_effects(original_action, new_action)
+                new_action.add_durative_condition(i, c)
+        for timing, effects in original_action.effects().items():
+            for effect in effects:
+                new_action._add_effect_instance(timing, effect)
```
alvalentini, line 115: effects are mutable now. Check them in all the transformers

upf/transformers/negative_conditions_remover.py
```diff
@@ -67,8 +72,10 @@ def get_rewritten_problem(self) -> Problem:
...
             return self._new_problem
         #NOTE that a different environment might be needed when multi-threading
-        self._create_problem_copy('no_negative_conditions')
-        self._new_problem_add_objects()
+
+        self._new_problem = Problem(f'{self._name}_{self._problem.name}', self._env)
+        for o in self._problem.all_objects():
+            self._new_problem.add_object(o)
         assert self._new_problem is not None
         self._modify_actions_and_goals()
```
alvalentini, line 80: In this method you are doing all the transformation

Framba-Luca: What do you mean?

alvalentini: All the transformation code is inside the `_modify_actions_and_goals` method.
You are not only modifying actions and goals.

Framba-Luca: Oh ok, so you mean I should update the name of the method to something more appropriate!

upf/transformers/conditional_effects_remover.py
alvalentini: The method `_get_fresh_action_name` should work without the need of calling other methods

upf/transformers/negative_conditions_remover.py
alvalentini: No, I think that if possible you should refactor the code in more methods.
Otherwise the code can simply stay in the `get_rewritten_problem` method.

upf/transformers/conditional_effects_remover.py
Framba-Luca: Yes, it works, but I like it better that the counter is reset for every different action. So in a problem with a very high number of actions that create a huge number of actions you don't end with super big `_count` for no reason.

It works in both ways, it's just a readability choiche

upf/transformers/negative_conditions_remover.py
Framba-Luca: Yes, makes sense

upf/transformers/conditional_effects_remover.py
alvalentini: You can always start with `count = 0` inside the `_get_fresh_action_name` and you will have the same behavior.

Framba-Luca: Actually no because the check for the fresh name is done on the original problem. We could simply shift it onto the new problem and then you will be right, and it will also be a better naming. (Because now, not every action created is added, so you could have an action name "move", splitted into "move_0", "move_1", "move_2" and "move_3", but only "move_0" and "move_3" actually added to the problem. So the readability also might become misleading.

While instead, if we check onto the new problem, if the user gives appropriate names, it will have an easier readability where every action created being in order without missing numbers.

upf/transformers/conditional_effects_remover.py
```diff
@@ -63,23 +62,23 @@ def get_rewritten_problem(self) -> Problem:
...
+                        self._new_problem.add_timed_effect(t, e.fluent(), nv)
                 else:
-                    self._new_problem.add_timed_effect(t, e)
-        self._new_problem_add_timed_goals()
-        self._new_problem_add_maintain_goals()
-        self._new_problem_add_goals()
+                    self._new_problem._add_effect_instance(t, e.clone())
+        self._new_problem.clear_actions()
         for ua in self._problem.unconditional_actions():
             self._new_problem.add_action(ua)
```
alvalentini, line 70: you need to clone the action here

alvalentini: @Framba-Luca double check in all the transformers that we clone all the mutable objects before adding them to the new problem.

upf/solvers/pddl_solver.py
```diff
@@ -47,6 +47,7 @@ def _plan_from_file(self, problem: 'upf.model.Problem', plan_filename: str) -> '
         actions = []
         with open(plan_filename) as plan:
             for line in plan.readlines():
+                print(line)
```
alvalentini, line 50: debug print to remove

upf/test/test_disjunctive_conditions_remover.py
```diff
@@ -180,7 +171,35 @@ def test_temproal_mockup(self):
...
         dnf_problem = dnfr.get_rewritten_problem()
-        new_act = dnfr.get_old_to_new_actions_mapping()[act]
+        new_act = dnfr.get_transformed_actions(act)
         self.assertEqual(len(dnf_problem.actions()), 81)
         self.assertEqual(len(new_act), 81)
         self.assertEqual(set(dnf_problem.actions().values()), set(new_act))
+
+
+    def test_temproal_mockup_2(self):
+
```
mikand, line 181:
```suggestion
```

mikand (approved):
LGTM I left few comments, then @alvalentini can merge this.

upf/test/test_problem.py
```diff
@@ -165,12 +165,12 @@ def test_robot(self):
...
         self.assertEqual(l2.type(), Location)
 
-        self.assertEqual(problem.name(), 'robot')
+        self.assertEqual(problem.name, 'robot')
         self.assertEqual(len(problem.fluents()), 2)
         self.assertEqual(problem.fluent('robot_at'), robot_at)
         self.assertEqual(problem.fluent('battery_charge'), battery_charge)
         self.assertEqual(len(problem.user_types()), 1)
-        self.assertEqual(problem.user_types()['Location'], Location)
+        self.assertEqual(problem.user_types()[0], Location)
```
alvalentini, line 173:
```suggestion
        self.assertEqual(problem.user_type('Location'), Location)
```

upf/test/test_problem.py
```diff
@@ -248,13 +248,13 @@ def test_robot_loader(self):
...
 
-        self.assertEqual(problem.name(), 'robot_loader')
+        self.assertEqual(problem.name, 'robot_loader')
         self.assertEqual(len(problem.fluents()), 3)
         self.assertEqual(problem.fluent('robot_at'), robot_at)
         self.assertEqual(problem.fluent('cargo_at'), cargo_at)
         self.assertEqual(problem.fluent('cargo_mounted'), cargo_mounted)
         self.assertEqual(len(problem.user_types()), 1)
-        self.assertEqual(problem.user_types()['Location'], Location)
+        self.assertEqual(problem.user_types()[0], Location)
```
alvalentini, line 257: Same as above

PR #66, opened 2021-11-22
Protobuf interface

alvalentini:


arbimo:
Hi, just to say, that I added a few comments in #48 related to the `.proto` specification that I believe are equally valid here.

mikand:
@arbimo I believe that @uwe-koeckemann is taking them into account right now, right @uwe-koeckemann ?

mikand:
Closed in favor of #78

PR #67, opened 2021-11-24
Setup file

Framba-Luca:


setup.py
```diff
@@ -0,0 +1,23 @@
...
+ UPF: A library that unifies planning frameworks
+ ============================================================
+    Insert long description here
+"""
+
+setup(name='upf',
+      version=upf.__version__,
+      description='Unified planning framework',
+      author='UPF Team',
+      author_email='info@upf.com',
```
alvalentini, line 18:
```suggestion
      author='AIPlan4EU Organization',
      author_email='aiplan4eu@fbk.eu',
```

@mikand do you have a better suggestion?

alvalentini:
We can remove `pysmt` from the requirements file

setup.py
alvalentini: @Framba-Luca for now proceed with this suggestion, also in the planner wrappers

PR #72, opened 2021-12-23
UPF to Tarski Converter

Framba-Luca:

PR #73, opened 2022-01-04
Extend the PDDL reader

alvalentini:


upf/io/pddl_reader.py
```diff
@@ -14,24 +14,321 @@
...
+                self._add_timed_effects(problem, act, e)
+        elif len(eff) == 3 and op == 'at' and eff[1] == 'start':
+            self._add_effect(problem, act, eff[2], timing=upf.model.StartTiming())
+        elif len(eff) == 3 and op == 'at' and eff[1] == 'end':
+            self._add_effect(problem, act, eff[2], timing=upf.model.EndTiming())
+        else:
+            raise upf.exceptions.UPFProblemDefinitionError(f'Not able to handle: {eff}')
+
+
+    def parse_problem(self, domain_filename: str, problem_filename: str) -> 'upf.model.Problem':
```
mikand, line 227: An interesting feature could be to parse just the domain and get a partial `upf.model.Problem`: the use-case would be to have a fixed domain written in PDDL that gets instantiated with dynamic data. 

Would it be possible to make the `problem_filename` optional here?

upf/io/pddl_reader.py
```diff
@@ -14,24 +14,321 @@
...
+        domain_res = pp_domain.parseFile(domain_filename)
+        problem_res = pp_problem.parseFile(problem_filename)
+
+        problem = upf.model.Problem(problem_res['name'], self._env,
+                                    initial_defaults={self._tm.BoolType(): self._em.FALSE()})
+
+        if 'predicates' in domain_res:
+            for p in domain_res['predicates']:
```
mikand, line 235:
```suggestion
        for p in domain_res.get('predicates', []):
```

upf/io/pddl_reader.py
```diff
@@ -14,24 +14,321 @@
...
+            return self._em.Real(Fraction(exp))
+        else:
+            raise upf.exceptions.UPFProblemDefinitionError(f'Not able to handle: {exp}')
+
+    def _add_effect(self, problem, act, exp, cond=True, timing=None):
+        op = exp[0]
+        if op == 'and':
+            exp = exp[1:]
+            for e in exp:
+                self._add_effect(problem, act, e, cond, timing)
```
mikand, line 156: It should be very easy to remove recursion here: it suffices to implement a super-basic conjunctive partitioning like: https://github.com/pysmt/pysmt/blob/master/pysmt/rewritings.py#L856

upf/io/pddl_reader.py
```diff
@@ -14,24 +14,321 @@
...
+            '-' : self._em.Minus,
+            '/' : self._em.Div,
+            '*' : self._em.Times
+        }
+
+    def _parse_exp(self, problem, act, var, exp):
+        if isinstance(exp, ParseResults) and len(exp) == 0:
+            return self._em.TRUE()
+        elif isinstance(exp, ParseResults) and exp[0] == '-' and len(exp) == 2:
+            return Times(-1, self._parse_exp(problem, act, var, exp[1]))
```
mikand, line 117: Maybe remove recursion here: it should be a matter of using a stack and a map for the results, no?

upf/io/pddl_reader.py
```diff
@@ -14,24 +14,321 @@
...
+                        for j in range(1, len(dur)):
+                            v = Fraction(dur[j][2])
+                            if dur[j][0] == '>=':
+                                if lower is None or v > lower:
+                                    lower = v
+                            elif dur[j][0] == '<=':
+                                if upper is None or v < upper:
+                                    upper = v
+                            else:
+                                raise upf.exceptions.UPFProblemDefinitionError(f'Not able to handle duration constraint of action {n}')
```
mikand, line 286: Maybe use a `SyntaxError` instead?

upf/test/test_pddl_io.py
```diff
@@ -234,6 +234,19 @@ def test_sailing_reader(self):
...
+        problem_filename = os.path.join(PDDL_DOMAINS_PATH, 'matchcellar', 'problem.pddl')
+        problem = reader.parse_problem(domain_filename, problem_filename)
+
+        self.assertTrue(problem is not None)
+        self.assertEqual(len(problem.fluents()), 4)
+        self.assertEqual(len(problem.actions()), 2)
+        self.assertEqual(len(problem.objects(problem.user_type('match'))), 3)
+        self.assertEqual(len(problem.objects(problem.user_type('fuse'))), 3)
+
     def test_matchcellar_writer(self):
```
mikand, line 250: Ad a few more tests: e.g. for each problems in UPF format, print them in PDDL and read them back asserting that the actions and initial states and goals are the same modulo reordering

mikand (changes requested):
@alvalentini Excellent work! I left some comments to be fixed, then we can merge!

upf/io/pddl_reader.py
```diff
@@ -14,24 +14,321 @@
...
+            '=' : self._em.Equals,
+            '+' : self._em.Plus,
+            '-' : self._em.Minus,
+            '/' : self._em.Div,
+            '*' : self._em.Times
+        }
+
+    def _parse_exp(self, problem, act, var, exp):
+        if isinstance(exp, ParseResults) and len(exp) == 0:
+            return self._em.TRUE()
```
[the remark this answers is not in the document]
alvalentini: This happen only when we have something like `:precondition ()`

mikand:
@alvalentini Coverage is low because now  the module `interop.tarski` becomes untested. Can you add a few tests to recover a bit of coverage? Or shall we wait for #72 to restore the coverage?

alvalentini:
> @alvalentini Coverage is low because now the module `interop.tarski` becomes untested. Can you add a few tests to recover a bit of coverage? Or shall we wait for #72 to restore the coverage?

@mikand I think we can wait #72 that adds several tests about the `interop.tarski` module

PR #79, opened 2022-02-01
Hierarchical Typing

Framba-Luca:


mikand (changes requested):
@Framba-Luca excellent work! I left some comments. 

Add a test where a type `object` is defined as root and also a test where it it in the middle of the hierarchy. We have then to decide if to keep it or not.

unified_planning/model/problem.py
```diff
@@ -362,12 +395,21 @@ def has_object(self, name: str) -> bool:
 
     def objects(self, typename: 'unified_planning.model.types.Type') -> List['unified_planning.model.object.Object']:
         '''Returns the objects of the given user types.'''
-        res = []
+        res:List['unified_planning.model.object.Object'] = []
```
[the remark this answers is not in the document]
Framba-Luca: If we make this as an Iterator, shall we make every other method, like "fluents()", "actions()" etc. as an Iterator? @mikand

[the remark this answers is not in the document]
mikand: Well, no because when you can simply return a list you already have stored in memory there is no point in forcing an iteration. Iterators make sense when you have to create a new list/set and return it

unified_planning/io/pddl_writer.py
```diff
@@ -146,31 +146,47 @@ def _write_domain(self, out: IO[str]):
...
+                problem_kind.has_discrete_time()): # type: ignore
                 out.write(' :durative-actions')
-            if self.problem.kind().has_duration_inequalities(): # type: ignore
+            if problem_kind.has_duration_inequalities(): # type: ignore
                 out.write(' :duration-inequalities')
             out.write(')\n')
 
-        out.write(f' (:types {" ".join([t.name() for t in self.problem.user_types()])})\n' if len(self.problem.user_types()) > 0 else '') # type: ignore
+        if problem_kind.has_hierarchical_typing(): # type: ignore
+            self.object_freshname = 'object'
```
mikand, line 174: Why storing the object fresh name in self? Creating fields in methods other than `__init__` is quite ugly

unified_planning/io/pddl_writer.py
```diff
@@ -146,31 +146,47 @@ def _write_domain(self, out: IO[str]):
...
+            while self.problem.has_type(self.object_freshname):
+                self.object_freshname = self.object_freshname + '_'
+            user_types_hierarchy = self.problem.user_types_hierarchy()
+            out.write(f' (:types\n')
+            stack: List['unified_planning.model.Type'] = user_types_hierarchy[None] if None in user_types_hierarchy else []
+            out.write(f'    {" ".join(t.name() if t.name() != "object" else self.object_freshname for t in stack)} - object\n')  # type: ignore
```
mikand, line 180: why disabling the type-checker?

mikand: probably stack should contain UserType's not Types?

unified_planning/io/pddl_writer.py
```diff
@@ -146,31 +146,47 @@ def _write_domain(self, out: IO[str]):
...
+            while stack:
+                current_type = stack.pop()
+                direct_sons: List['unified_planning.model.Type'] = user_types_hierarchy[current_type]
+                if direct_sons:
+                    stack.extend(direct_sons)
+                    out.write(f'    {" ".join([t.name() if t.name() != "object" else self.object_freshname for t in direct_sons])} - {current_type.name() if current_type.name() != "object" else self.object_freshname}\n')  # type: ignore
+            out.write(' )\n')
+        else:
```
mikand, line 188: Is this else needed? if there is no type hierarchy the code above should work already. Moreover, you are not handling the case of no-hierarchy with a type called `object`

unified_planning/io/pddl_writer.py
```diff
@@ -180,7 +196,7 @@ def _write_domain(self, out: IO[str]):
                 i = 0
                 for p in f.signature():
                     if p.is_user_type():
-                        params.append(f' ?p{str(i)} - {p.name()}') # type: ignore
+                        params.append(f' ?p{str(i)} - {p.name() if p.name() != "object" else self.object_freshname}') # type: ignore
```
mikand, line 199: you are using this piece of code everywhere, maybe incapsulate it in a utility function?

unified_planning/io/pddl_writer.py
```diff
@@ -312,7 +328,9 @@ def _write_problem(self, out: IO[str]):
         if len(self.problem.user_types()) > 0:
             out.write(' (:objects ')
             for t in self.problem.user_types():
-                out.write(f'\n   {" ".join([o.name() for o in self.problem.objects(t)])} - {t.name()}') # type: ignore
+                objects: List['unified_planning.model.Object'] = list(self.problem.objects(t))
+                if len(objects) > 0:
+                    out.write(f'\n   {" ".join([o.name() for o in objects])} - {t.name() if t.name() != "object" else self.object_freshname}') # type: ignore
```
mikand, line 333: Again, why disable the type-checker?

unified_planning/model/types.py
```diff
@@ -53,17 +53,24 @@ def is_bool_type(self) -> bool:
...
+    def __init__(self, name: str, father: Optional[Type] = None):
         Type.__init__(self)
         self._name = name
+        if father is not None and (not father.is_user_type()):
+            raise UPTypeError('father field of a UserType must be a UserType.')
+        self._father = father
 
     def __repr__(self) -> str:
-        return self.name()
+        return self._name if self._father is None else f'{self._name} - {self._father.name()}' # type: ignore
```
mikand, line 64: Why disabling the type-checker?

unified_planning/model/types.py
```diff
@@ -151,20 +158,34 @@ def RealType(self, lower_bound: Fraction = None, upper_bound: Fraction = None) -
...
+            ut = _UserType(name, father)
+            self._user_types[(name, father)] = ut
             return ut
+    
+    def user_type_ancestors(self, user_type: Type) -> Iterator[Type]:
+        '''Returns all the ancestors of the given UserType, including itself.'''
+        if not user_type.is_user_type():
+            raise UPTypeError('The function user_type_ancestors can be called only on UserTypes.')
+        yield user_type
+        father: Optional[Type] = user_type.father() # type: ignore
```
mikand, line 177: SDTC = stop disabling the type checker!

unified_planning/test/test_pddl_io.py
```diff
@@ -287,6 +287,9 @@ def test_examples_io(self):
                 reader = PDDLReader()
                 parsed_problem = reader.parse_problem(domain_filename, problem_filename)
 
+                if problem.has_type('object'):
+                    continue #due to the renaming of the type object, the following tests will fail
```
mikand, line 291: well, adapt them so that they pass...

mikand (changes requested):
@Framba-Luca Please review *all* the instances where you disabled the type-checking: the `type: ignore` annotation should be used sparingly otherwise having type annotations is pointless.

When needed (and only when needed) you can also use `cast()`: https://adamj.eu/tech/2021/07/06/python-type-hints-how-to-use-typing-cast

mikand (approved):
LGTM

alvalentini (approved):
I saw you remove some `type: ignore` using `cast`.
We can do the same in many other cases where we use `type: ignore`, but maybe we can do that in a specific PR.

Issue #81, opened 2022-02-04
Floating-point numbers converted to fractions by the pddl writer


LBonassi95:
**Bug description**
When using ``` pddl_writer ``` to write a problem in PDDL, floating-point numbers are converted to fractions. For instance, 1.7 becomes 17/10 (see below). In PDDL this syntax is not supported.

**To reproduce**
```python
import unified_planning
from unified_planning.shortcuts import *
from unified_planning.io.pddl_reader import PDDLReader
from unified_planning.io.pddl_writer import PDDLWriter

reader = PDDLReader()
pddl_problem = reader.parse_problem('./domain.pddl', './instance_2_1000_1229.pddl')
w = PDDLWriter(pddl_problem)
print(w.get_problem())
```

The goal field in the output problem is:
``` (:goal (and (and (<= 1 (x farm0)) (<= 1 (x farm1)) (<= 1400 (+ (* 17/10 (x farm1)) (x farm0)))))) ```

PDDL files: [pddl_files.zip](https://github.com/aiplan4eu/unified-planning/files/8002867/pddl_files.zip)

**Expected behavior**
The goal field in the output problem is:
``` (:goal (and (and (<= 1 (x farm0)) (<= 1 (x farm1)) (<= 1400 (+ (* 1.7 (x farm1)) (x farm0)))))) ```

**Environment**
 - OS: Ubuntu 20.04.3 LTS
 - Python version: 3.8.12

PR #82, opened 2022-02-14
Fixed print or decimal numbers in PDDL

mikand:
Fixed issue #81 by adding a proper print and warning for rational numbers


alvalentini (approved):
It looks good to me!

PR #84, opened 2022-02-17
Add quality metrics

alvalentini:
See discussion #69


unified_planning/io/pddl_reader.py
```diff
@@ -357,22 +369,66 @@ def parse_problem(self, domain_filename: str,
...
+                            if totalcost in fve.get(c):
+                                has_actions_cost = False
+                                break
+                if has_actions_cost:
+                    for _, el in dur_act.effects().items():
+                        for e in el:
+                            if totalcost in fve.get(e.fluent()) or totalcost in fve.get(e.value()) \
+                               or totalcost in fve.get(e.condition()):
+                                has_actions_cost = False
+                                break
```
mikand, line 404: If you wrap this code in a function, you can use `return` to avoid the repeated `if has_actions_cost:`

unified_planning/io/pddl_reader.py
```diff
@@ -357,22 +369,66 @@ def parse_problem(self, domain_filename: str,
...
                     act.add_precondition(self._parse_exp(problem, act, {}, a['pre'][0]))
                 if 'eff' in a:
                     self._add_effect(problem, act, a['eff'][0])
+
+                if has_actions_cost:
+                    for c in act.preconditions():
+                        if totalcost in fve.get(c):
+                            has_actions_cost = False
+                            break
+                if has_actions_cost:
```
mikand, line 419: Same as above

unified_planning/io/pddl_reader.py
```diff
@@ -405,4 +461,44 @@ def parse_problem(self, domain_filename: str,
 
             problem.add_goal(self._parse_exp(problem, None, {}, problem_res['goal'][0]))
 
+            if has_actions_cost:
+                if not problem.initial_value(totalcost).constant_value() == 0:
+                    has_actions_cost = False
+            if has_actions_cost:
```
mikand, line 467: Same

unified_planning/model/action.py
```diff
@@ -90,6 +91,13 @@ def name(self, new_name: str):
         """Sets the parameter name."""
         self._name = new_name
 
+    def cost(self) -> Optional['up.model.fnode.FNode']:
```
mikand, line 94: This is OK. Another possible design would be to have a map inside the plan quality metric from actions to costs. Maybe we can discuss this in next WP3 meeting.

unified_planning/model/problem.py
```diff
@@ -568,14 +582,20 @@ def kind(self) -> 'unified_planning.model.problem_kind.ProblemKind':
                 self._update_problem_kind_condition(goal)
         for goal in self._goals:
             self._update_problem_kind_condition(goal)
+        for metric in self._metrics:
+            if isinstance(metric, up.model.metrics.MinimizeExpressionOnFinalState) or \
+               isinstance(metric, up.model.metrics.MaximizeExpressionOnFinalState):
+                self._kind.set_quality_metrics('FINAL_VALUE') # type: ignore
+            elif isinstance(metric, up.model.metrics.MinimizeActionCosts):
+                self._kind.set_quality_metrics('ACTIONS_COST') # type: ignore
```
mikand, line 590: Add `else: assert(False, 'Unknown quality metric')`

unified_planning/solvers/factory.py
```diff
@@ -80,7 +84,8 @@ def _get_solver(self, solver_kind: str, name: Optional[str] = None,
     def OneshotPlanner(self, *, name: Optional[str] = None,
                        names: Optional[List[str]] = None,
                        params: Union[Dict[str, str], List[Dict[str, str]]] = None,
-                       problem_kind: ProblemKind = ProblemKind()) -> Optional['unified_planning.solvers.solver.Solver']:
+                       problem_kind: ProblemKind = ProblemKind(),
+                       optimality_guarantee: Optional[int] = None) -> Optional['up.solvers.solver.Solver']:
```
mikand, line 88: I think it would be useful to also support the strings `"OPTIMAL"` and `"SATISFICING"` in addition to the INT values

mikand (approved):
Excellent work @alvalentini! I left some minor commets. The only major thing to discuss is whether it is better to have costs within actions or within the quality metric.

Issue #76, opened 2022-01-31
Improving on the interface after planning to interpret additional planner output


hstairs:
The Unified Planning Framework should be able to have some feature to get additional information after the planner has done its task. This may be related to both the case where the planner has not been able to find a solution and the case where the planner did find a solution. For the former, additional information explaining the reasons of the failure are desirable, like memory crash, instance proved unsolvable so on and so forth. For the latter, additional information could include statistics of the planner, or the cost of the found solution.
@mikand that perhaps we can have a more structured data type for the result of the planner.

arbimo:
Just to mention that I made a related proposal for the gRPC interface in PR #78.
The `Stopped` message essentially encode the data returned by the planner, independently of any plan provided previously.

Needless to say, this should match almost one-to-one with what is expressible with the python API so it might a good place to discuss it based on a concrete format ;)

hstairs:
Hi, I saw your proposal, thanks! It seems it contains all the necessary data fields. What are the next steps then to get the UP to actually exposes these features so that they can be used also by planners that do not adopt a gRPC interface? @mikand ?

madmage:
I guess we should first talk about the timeline of the features we intend to include in the UP. What features come first? What is really necessary? A prioritized list of features should be published somewhere, I think. It could be also important for us to contribute to the discussions and maybe to the implementation.
The protobuf definition, being something used to expose features in the UP, should follow this development. I don't see it good to provide the full protobuf API first (as stubs) and then start implementing (maybe) the features in the UP.
Moreover, I see there is already a page with a Taxonomy https://github.com/aiplan4eu/unified-planning/wiki/Planning-Taxonomy that we can use to split the topics and start discussing/prioritizing the features.

madmage:
Just to summarize, in this issue we are actually asking the UP to provide in the final output of the planner, the following additional fields, as requested by the OP of this issue and by the gRPC proposal (https://github.com/aiplan4eu/unified-planning/pull/78/files)"

- metrics (as in the OP and in the gRPC API proposal), for example (this is what we currently have, just to have concrete examples) time spent planning or number of states visited
- reason of failure (in case of failure), some of them are listed in the gRPC API proposal in the Stopped message (do we need two separate messages Report and Stopped? isn't it enough to put everything in the same planner output response?)
  - no plan exists
  - search space exhausted
  - timeout
  - out of memory
  - internal error
  - etc.

is it correct?

[...]

PR #94, opened 2022-03-09
Result object

Framba-Luca:
Implemented the FinalReport and IntermediateReport class with all the needed adjustments and new parameter to the solve method to send the IntermediateReports.

This solves #76


unified_planning/plan.py
```diff
@@ -102,3 +114,119 @@ def __eq__(self, oth: object) -> bool:
     def actions(self) -> List[Tuple[Fraction, ActionInstance, Optional[Fraction]]]:
         '''Returns the sequence of action instances.'''
         return self._actions
+
+class IntermediateReport: #NOTE should we create a class "Report", that then is extended?
```
alvalentini, line 118: I think it is a good idea to refactor the common fields in a base class

unified_planning/solvers/parallel.py
```diff
@@ -90,9 +91,9 @@ def solve(self, problem: 'unified_planning.model.Problem') -> 'unified_planning.
                 else:
                     raise
             actions.append(ActionInstance(new_a, tuple(params)))
-        return SequentialPlan(actions)
+        return up.plan.FinalReport(final_report.status(), SequentialPlan(actions), final_report.planner_name(), final_report.metrics(), final_report.log_messages()) #NOTE: this can be only a SequentialPlan?
```
alvalentini, line 94: you are right! Here we should handle all possible plans

unified_planning/plan.py
```diff
@@ -14,12 +14,24 @@
...
+ALL_STATUS = list(range(0, 8))
+
+(
+SATISFIED, OPTIMAL, UNSATISFIED, SEARCH_SPACE_EXHAUSTED,
+TIMEOUT, MEMOUT, INTERNAL_ERROR, UNSUPPORTED_PROBLEM
+) = ALL_STATUS
+
+POSITIVE_OUTCOMES = frozenset([SATISFIED, OPTIMAL])
+
+NEGATIVE_OUTCOMES = frozenset([UNSATISFIED, SEARCH_SPACE_EXHAUSTED, UNSUPPORTED_PROBLEM])
```
alvalentini, line 33: Should we include also TIMEOUT, MEMOUT, INTERNAL_ERROR?

alvalentini (changes requested):

alvalentini:
The `IntermediateReport` and the `FinalReport` proposed by @arbimo in the proto file share the same contents except for the `Status` field. @Framba-Luca followed the same structured in the UP implementation of this PR.

I'm wondering if it would be better to have a `Answer` message/class that has the same fields as the `FinalReport` with an additional possible value `INTERMEDIATE` for the `Status`.

```
message Answer {
    enum Status {
        // ==== Planner stopped normally ======

        // Valid plan found and search stopped immediately
        // The `best_plan` field must be set.
        SAT = 0;
        // Plan found with optimality guarantee
        // The `best_plan` field must be set and contain an optimal solution.
        OPT = 1;
        // No plan exists
        UNSAT = 2;
        // The planner was not able to find a solution but does not give any guarantee that none exist
        // (i.e. the planner might not be complete)
        SEARCH_SPACE_EXHAUSTED = 3;

        // ====== Planner exited before making any conclusion ====
        // Search stopped before concluding OPT or UNSAT
        // If a plan was found, it might be reported in the `best_plan` field

        // The planner ran out of time
        TIMEOUT = 13;
        // The planner ran out of memory
        MEMOUT = 14;
        // The planner faced an internal error.
        INTERNAL_ERROR = 15;
        // The problem submitted is not supported by the planner.
        UNSUPPORTED_PROBLEM = 16;

        // ====== Intermediate answer ====
        INTERMEDIATE = 17;
    }
    Status status = 1;

    // Optional. Best plan found so far, if any.
    Plan best_plan = 2;

    // A set of planner specific values that can be reported, for instance
    // - "grounding-time": "10ms"
    // - "expanded-states": "1290"
    map<string, string> metrics = 3;

    // Optional logs about the planner's activity.
    // Note that it should not be expected that logging messages are visible to the end user.
    // If used in conjunction with INTERNAL_ERROR or UNSUPPORTED_PROBLEM, it would be expected to have at least one log message at the ERROR level.
    repeated LogMessage logs = 4;
}
```

@arbimo @mikand what do you think?

arbimo:
@alvalentini Yes, I think it makes sense to merge them into a single one. While they were initially clearly distinct, they essentially converged over time. 
It might make sense to keep the two classes in the python API however (just like for actions).

In your proposal it might make sense to rename `best_plan` to just `plan` which is less awkward to use. With a comment specifying that this is a plan that should not be dominated by a previous plan and that it should be left empty if it is an intermediate message and no plan has been found since the previous reported one.

unified_planning/plan.py
Framba-Luca: I don't think so because NEGATIVE_OUTCOMES assumes that a plan is not found. TIMEOUT, MEMOUT and INTERNAL_ERROR, at least from my interpretation of their meaning, could return a plan, still knowing that something went wrong.

alvalentini: I don't agree. They are reasons why a plan is not found.
TIMEOUT for example means that a plan is not found within the time limit

arbimo: In the case of planning with explicit costs, it could be the case that a plan was found but the planner exited (e.g. timeout) before finding the optimal one. In this case, the "best plan found" remains very important.

alvalentini: Do you think this is the case also for MEMOUT, INTERNAL_ERROR and SEARCH_SPACE_EXHAUSTED?

alvalentini: From what I understood, SATISFIED and OPTIMAL requires the `plan` field set while the UNSATISFIED and UNSUPPORTED_PROBLEM never have the `plan` field set. In the other cases a plan might be set, right?

unified_planning/plan.py
```diff
@@ -102,3 +114,101 @@ def __eq__(self, oth: object) -> bool:
...
+        '''Returns the status as a str.'''
+        return __STATUS_STR__[self._status]
+
+
+def status_to_str(id: int) -> str:
+    '''Returns a string representation of the given status.'''
+    return __STATUS_STR__[id]
+
+__STATUS_STR__ = {
+    SATISFIED: 'STATISFIED',
```
alvalentini, line 206: Typo:
```suggestion
    SATISFIED: 'SATISFIED',
```

Check to use the same name here and in the proto file.

Framba-Luca: Should we use SAT or SATISFIED? I personally prefer the long option (more clear), but it's not really something major.

unified_planning/plan.py
arbimo: Correct except for SEARCh_SPACE_EXHAUSTED which that the planner stopped because he is not capable of generating more potential solutions (but there might other that are possible, i.e., the planner is incomplete). 

In a nutshell :
-  if a plan is provided with the SEARCH_SPACE_EXHAUSTED flag, it is a weaker version of the OPTIMAL flag (without an actual guarantee on the optimality).
- if no plan is provided with the SEARCH_SPACE_EXHAUSTED flag, it is a weaker version of the UNSATISFIABLE version : the planner reports that he is not capable of finding a solution but a complete planner might find one.

I was unsure whether to include this flag. The reasoning was that (1) we want to trust the planner when it says that something is optimal/unsat because this is not something we can routinely verify, (2) if we don't include those weaker cases, then incomplete planners will have to (incorrectly) report optimal/unsat.

This is mostly related to planner incompletness so it might handle another way (as distinct boolean flag, directly in the planner description, ...)

alvalentini: I agree on what you said about the case "no plan is provided with the SEARCH_SPACE_EXHAUSTED flag".
For the case "a plan is provided with the SEARCH_SPACE_EXHAUSTED flag", I think we can express the same message with the SATISFIED flag: the planner reports that the plan is the best one found so far (with no guarantee on the optimality) and it is not able to find any better plans.

alvalentini (changes requested):
Except for the minor comment, I think the PR is ready to merge.

alvalentini (approved):

mikand (approved):
@Framba-Luca good work! I left some minor comments, but this looks OK.
@roeger @hstairs @LBonassi95: let us know if it is OK to merge this or if you have any feedback.

unified_planning/solvers/pddl_solver.py
```diff
@@ -69,16 +75,20 @@ def solve(self, problem: 'unified_planning.model.Problem') -> Optional['unified_
...
+            logs: List['up.solvers.results.LogMessage'] = []
+            logs.append(up.solvers.results.LogMessage(up.solvers.results.INFO, res.stdout.decode()))
+            logs.append(up.solvers.results.LogMessage(up.solvers.results.ERROR, res.stderr.decode()))
+            if os.path.isfile(plan_filename):
                 plan = self._plan_from_file(problem, plan_filename)
+        status: int = self._result_status(problem, plan)
+        return PlanGenerationResult(status, plan, log_messages=logs)
 
-        return plan
+    def _result_status(self, problem: 'up.model.Problem', plan: Optional['up.plan.Plan']) -> int:
```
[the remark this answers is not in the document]
roeger: No problem from my side.

roeger (commented):
The changes look good. I am only unsure about the terminology "satisfied/unsatisfied". What is satisfied or unsatisfied here? From my understanding "solved" and "unsolvable" would be the idiomatic terms for what it should express according to the comments. In particular, I don't like satisfied so much, because my brain always wants to connect it to satisficing planning (which obviously is not called satisfying planning).

mikand:
@arbimo @roeger To be clear, here we have three "successful outcomes":
- SATISFIED/SAT: A valid plan has been found, the plan is not necessarily optimal 
- UNSOLVABLE/UNSAT: We proved that no valid plan exists
- OPTIMAL: An optimal plan has been found (proving that no better plan can exist)

Now, for the naming I like SATISFIED/UNSOLVABLE/OPTIMAL better than SAT/UNSAT/OPTIMAL, but this is a matter of taste. 

Do you agree on the three cases above? Any feeedback/suggestion?

arbimo:
> Now, for the naming I like SATISFIED/UNSOLVABLE/OPTIMAL better than SAT/UNSAT/OPTIMAL, but this is a matter of taste.
@mikand Yes, I think it is a perfectly fine combination :+1:   (my personal preference would have been satisfied/unsatisfiable/optimal but as you said it is a matter of taste and "unsolvable" might more common in the planning community).

roeger:
I also don't like the asymmetry between SOLVED (or SATISFIED) and OPTIMAL because the latter implies the former. If they were not so long, I would argue for SOLVED_WITHOUT_OPTIMALITY_GUARANTEE / DETECTED_UNSOLVABLE / SOLVED_WITH_OPTIMALITY_GUARANTEE.

hstairs:
we could have SOLVED_OPTIMALLY/SOLVED/UNSOLVABLE. Or we can leave that to be SOLVED/UNSOLVED and have an OPTIMALITY_GUARANTEE field to be true or false. This last option is what I like the most because it is more factored; not sure though where that will fit in the Answer message defined above, probably outside "status"?

unified_planning/solvers/parallel.py
```diff
@@ -56,25 +58,74 @@ def _run_parallel(self, fname, *args):
...
+                                        up.solvers.results.TIMEOUT,
+                                        up.solvers.results.UNSOLVABLE_INCOMPLETELY,
+                                        up.solvers.results.MEMOUT,
+                                        up.solvers.results.INTERNAL_ERROR]
+            for ro in result_order:
+                for r in results:
+                    if r.status == ro:
+                        return r
+            # if no results are given by the planner, we create a default one
+            return up.solvers.PlanGenerationResult(up.solvers.results.TIMEOUT if timeout else up.solvers.results.UNSOLVABLE_INCOMPLETELY,
```
alvalentini, line 98: Here I would keep only UNSOLVABLE_INCOMPLETELY

Framba-Luca: Why? If a planner went in TIMEOUT it would make sense to let the user know that with more time a plan might be found, while UNSOLVABLE_INCOMPLETELY means that, even with more time, those planners will not find a plan.

unified_planning/solvers/parallel.py
```diff
@@ -56,25 +58,74 @@ def _run_parallel(self, fname, *args):
...
+                        timeout = True
         for p in processes:
             p.terminate()
-        return res
+        if definitive_result_found: # A planner found a definitive result
+            return res
+        else:
+            result_order: List[int] = [ up.solvers.results.SOLVED_SATISFICING,  # List containing the results in the order we prefer them
+                                        up.solvers.results.TIMEOUT,
+                                        up.solvers.results.UNSOLVABLE_INCOMPLETELY,
```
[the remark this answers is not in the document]
Framba-Luca: Why not?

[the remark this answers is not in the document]
alvalentini: Because when the status is UNSOLVABLE_INCOMPLETELY, no plans are present in the result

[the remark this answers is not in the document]
Framba-Luca: True.

unified_planning/solvers/parallel.py
```diff
@@ -87,16 +89,16 @@ def _run_parallel(self, fname, *args):
         else:
             result_order: List[int] = [ up.solvers.results.SOLVED_SATISFICING,  # List containing the results in the order we prefer them
                                         up.solvers.results.TIMEOUT,
-                                        up.solvers.results.UNSOLVABLE_INCOMPLETELY,
                                         up.solvers.results.MEMOUT,
                                         up.solvers.results.INTERNAL_ERROR]
+            logs = up.solvers.LogMessage(up.solvers.results.INFO, ' '.join(planners_final_status))
```
alvalentini, line 94: Here you have at least to connect the status with its planner name.

unified_planning/solvers/parallel.py
alvalentini: This is not solved. You need to remove the timeout.

alvalentini (changes requested):

unified_planning/solvers/parallel.py
```diff
@@ -56,25 +58,73 @@ def _run_parallel(self, fname, *args):
...
+                break
             (idx, res) = signaling_queue.get(block=True)
+            processes_alive -= 1
             if isinstance(res, BaseException):
                 raise res
             else:
-                break
+                assert isinstance(res, up.solvers.PlanGenerationResult)
+                # If the planner is sure about the result (optimality of the result or impossibility of the problem) exit the loop
+                if res.status == up.solvers.results.SOLVED_OPTIMALLY or res.status == up.solvers.results.UNSOLVABLE_PROVEN:
```
mikand, line 75: I would terminate also when a SATISFICING plan is returned if the problem (args[0]) has no optimality metrics.

mikand (approved):
I left two minor comments, after those, @alvalentini  please go ahead and merge

PR #99, opened 2022-03-11
Simulated effects

alvalentini:


unified_planning/model/state.py
```diff
@@ -0,0 +1,21 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import unified_planning as up
+
+
+class State:
```
mikand, line 19: Maybe 2 lines of doc :smile: ?

unified_planning/transformers/negative_conditions_remover.py
```diff
@@ -147,6 +147,8 @@ def get_rewritten_problem(self) -> Problem:
 
         for action in self._problem.actions:
             if isinstance(action, InstantaneousAction):
+                if action.simulated_effects is not None:
+                    raise up.exceptions.UPUsageError('NegativeConditionsRemover does not work with simulated effects')
```
mikand, line 151: can't we trap this with the `ProblemKind` at the beginning of the function?

mikand (changes requested):
@alvalentini Outstanding work!
Two things
1. I do not like the plural in the `SimulatedEffects`, I understand that a single instance affects multiple fluents, but the plural is just weird. I would suggest saying that a single SimulatedEffect and affect more than one fluent and use the singular.
2. You added guards only to the negative condition remover. What about the other transformers? do they work or shall we guard them?

unified_planning/model/state.py
alvalentini: :+1:

unified_planning/transformers/negative_conditions_remover.py
alvalentini: Right!

alvalentini:
@mikand  The other transfomers should not have an impact on the simulated effects and then they should work fine

mikand (approved):

PR #101, opened 2022-03-17
Added a test to the grounder so that it is called via factory

Framba-Luca:


mikand (changes requested):
Please rename the embedded grounder, having it called `grounder` is quite confusing

mikand (approved):
LGTM

PR #112, opened 2022-03-29
96 protobuf reader writer

uwe-koeckemann:
As suggested by @arbimo I made a PR out of the on-going protobuf conversion work.

I added a few comments to the protobuf definitions as well (as TODOs) in cases I found something did not align or possibly missing.


unified_planning/grpc/proto_writer.py
```diff
@@ -0,0 +1,341 @@
...
+        )
+
+    @handles(unified_planning.model.FNode)
+    def _convert_fnode(self, exp):
+        payload = exp._content.payload
+        args = exp._content.args
+        atom = None
+        p_type = ""
+        kind = unified_planning_pb2.ExpressionKind.Value("UNKNOWN")
+        arg_list = []
```
arbimo, line 83: arg_list might be a confusing name as the expression is either an atom or a list containing the "function" (first element of the list) and its parameters (rest of the list)

unified_planning/grpc/proto_writer.py
```diff
@@ -0,0 +1,341 @@
...
+            atom = unified_planning_pb2.Atom(symbol=payload.name())
+        else:
+            kind = unified_planning_pb2.ExpressionKind.Value("FUNCTION_SYMBOL")
+            atom = unified_planning_pb2.Atom(symbol=map_operator(exp.node_type()))
+
+        for arg in args:
+            arg_list.append(self.convert(arg))
+
+        return unified_planning_pb2.Expression(
+            atom=atom, list=arg_list, kind=kind, type=p_type
```
arbimo, line 121: Here it seems that the `atom` and `arg_list` are not exclusive (same as comment on `arg_list` name).

arbimo (commented):
@franklinselva related to our discussion on converting expressions.

requirements.txt
```diff
@@ -1 +1,5 @@
 pyparsing
+git+https://github.com/aig-upf/tarski.git@ebfda1c13ac908904d5b74587971cc7149e73d85#egg=tarski[arithmetic]
```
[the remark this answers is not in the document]
arbimo: Thanks, I was about to go dive in this issue and I am really happy that you pointed out before I had to =)

arbimo:
Hi everyone and in particular @mikand and @alvalentini ,
I believe that this PR is now ready for review.

In essence it provides a complete bi-directional translation between a UP `Problem` and the protobuf format.
There were some minimal changes to the protobuf definition in the process, mostly to better align it with the UP 
 - full feature list and associated documentation
 - Fractions for numbers
 -  a distinct `Engine` field in the planner's report to allow giving the name of the planning engine. Note that: (1) it is called Engine as we probably want to reuse for e.g. a validator (2) it is a complete message because it will certainly be extended in the future to allow more information in the future (in particular, options and version number which would be immensely useful when someone wants to fill in a bug report for a planner)

The changes to the UP side are kept extremely minimal (thanks to the prior alignment done by @alvalentini). 

The test strategy is for the most part to translate **all** tests problems and plans into protobuf and back, asserting that the result is identical to the original problem. Thus we have the guarantee that nothing is lost during the translation. Currently, all test problems pass, it should be an objective to keep it this way (it will obviously be caught by CI if it wasn't the case).

I would of course like to see this merged as soon as the code is considered good enough (I can't speak for @franklinselva and @uwe-koeckemann  that laid the foundations, but I am very far from a python expert) but consider the protobuf implementation as experimental (i.e. subject to change without a version bump). We will certainly flush out the last bugs and design problem when plugging a planner (which we are close to but it would greatly ease our work once this is merged).

.mypy.ini
```diff
@@ -0,0 +1,6 @@
+[mypy]
+python_version = 3.8
+
+[mypy-unified_planning.grpc.generated.*]
+ignore_errors = True
+exclude = generated
```
alvalentini, line 6: I don't think the python version is needed and the `exclude` option may only be set in the global section (`[mypy]`), but I think we can remove it and just keep the `ignore_errors`.

```suggestion
[mypy]

[mypy-unified_planning.grpc.generated.*]
ignore_errors = True
```

unified_planning/grpc/proto_reader.py
```diff
@@ -0,0 +1,544 @@
...
+    DurativeAction,
+    Effect,
+    InstantaneousAction,
+    Parameter,
+    Problem,
+    Variable,
+)
+from unified_planning.model.effect import EffectKind
+from unified_planning.model.operators import OperatorKind
+from unified_planning.shortcuts import BoolType, IntType, RealType, UserType
```
alvalentini, line 33: The `shortcuts` should be used only by an end user but not inside the `unified-planning`

unified_planning/grpc/proto_reader.py
```diff
@@ -0,0 +1,544 @@
...
+    elif "integer[" in s:
+        lb = int(s.split("[")[1].split(",")[0])
+        ub = int(s.split(",")[1].split("]")[0])
+        return problem.env.type_manager.IntType(lb, ub)
+    elif s == "real":
+        return problem.env.type_manager.RealType()
+    elif "real[" in s:
+        lb = float(s.split("[")[1].split(",")[0])
+        ub = float(s.split(",")[1].split("]")[0])
+        return problem.env.type_manager.RealType(lb, ub)
```
alvalentini, line 50: `RealType` bounds should be `Fraction`, not `float`

unified_planning/grpc/proto_reader.py
```diff
@@ -0,0 +1,544 @@
...
+            args = []
+            payload = None
+
+            fluent = msg.list.pop(0)
+            if fluent.kind == unified_planning_pb2.ExpressionKind.Value(
+                "FLUENT_SYMBOL"
+            ):
+                payload = self.convert(fluent.atom, problem)
+            else:
+                args.append(self.convert(fluent, problem, param_map))
```
alvalentini, line 142: This `else` branch is never reached, right?

unified_planning/grpc/proto_reader.py
```diff
@@ -0,0 +1,544 @@
...
+
+            args.extend([self.convert(m, problem, param_map) for m in msg.list])
+            return problem.env.expression_manager.create_node(
+                node_type=OperatorKind.FLUENT_EXP,
+                args=tuple(args),
+                payload=payload,
+            )
```
alvalentini, line 149: Why did you use the low lovel `create_node` method?

```suggestion
            return problem.env.expression_manager.FluentExp(payload, tuple(args))
```

unified_planning/grpc/proto_reader.py
```diff
@@ -0,0 +1,544 @@
...
+            args = []
+            payload = None
+
+            symbol = msg.list.pop(0)
+            if symbol.kind == unified_planning_pb2.ExpressionKind.Value(
+                "FUNCTION_SYMBOL"
+            ):
+                node_type = op_to_node_type(symbol.atom.symbol)
+            else:
+                args.append(self.convert(symbol, problem, param_map))
```
alvalentini, line 163: Same as above

unified_planning/grpc/proto_reader.py
```diff
@@ -0,0 +1,544 @@
...
+            PROBLEM.add_fluent(
+                self.convert(f, problem),
+                default_initial_value=self.convert(f.default_value, problem, [])
+                if f.HasField("default_value")
+                else None,
+            )
+        for f in msg.actions:
+            PROBLEM.add_action(self.convert(f, problem))
+        for eff in msg.timed_effects:
+            PROBLEM.add_timed_effect(self.convert(eff, problem))
```
alvalentini, line 254: The `add_timed_effect` has a different firm.

```
def add_timed_effect(self, timing: 'up.model.timing.Timing', fluent: Union['up.model.fnode.FNode', 'up.model.fluent.Fluent'],
                     value: 'up.model.expression.Expression', condition: 'up.model.expression.BoolExpression' = True)
```

unified_planning/grpc/proto_reader.py
```diff
@@ -0,0 +1,544 @@
...
+
+        for assign in msg.initial_state:
+            PROBLEM.set_initial_value(
+                fluent=self.convert(assign.fluent, problem, []),
+                value=self.convert(assign.value, problem, []),
+            )
+
+        for goal in msg.goals:
+            timing = self.convert(goal.timing)
```
alvalentini, line 263: Here you are discarding the temporal information

unified_planning/grpc/proto_reader.py
```diff
@@ -0,0 +1,544 @@
...
+
+    @handles(unified_planning_pb2.ObjectDeclaration)  # type: ignore
+    def _convert_object(self, msg, problem):
+        obj = unified_planning.model.Object(
+            msg.name, convert_type_str(msg.type, problem)
+        )
+        return obj
+
+    @handles(unified_planning_pb2.Expression)  # type: ignore
+    def _convert_expression(self, msg, problem, param_map):
```
alvalentini, line 117: Is the `param_map` needed?

unified_planning/grpc/proto_reader.py
```diff
@@ -0,0 +1,544 @@
...
+            goal = self.convert(goal.goal, problem, [])
+            PROBLEM.add_goal(goal)
+
+        for metric in msg.metrics:
+            PROBLEM.add_quality_metric(self.convert(metric, problem, None))
+
+        return PROBLEM
+
+    @handles(unified_planning_pb2.Metric)  # type: ignore
+    def _convert_metric(self, msg, problem, param_map):
```
alvalentini, line 273: Is the `param_map` needed?

unified_planning/grpc/proto_reader.py
```diff
@@ -0,0 +1,544 @@
...
+        elif result.status == unified_planning_pb2.PlanGenerationResult.Status.Value(
+            "UNSUPPORTED_PROBLEM"
+        ):
+            status = (
+                unified_planning.solvers.results.PlanGenerationResultStatus.UNSUPPORTED_PROBLEM
+            )
+        else:
+            raise UPException(f"Unknown Planner Status: {result.status}")
+
+        # FIXME: Metrics and logs are not supported yet
```
alvalentini, line 514: This is something that you want to handle in a different PR?

unified_planning/grpc/proto_writer.py
```diff
@@ -0,0 +1,534 @@
...
+            if fluent in problem.fluents_defaults
+            else None,
+        )
+
+    @handles(unified_planning.model.Object)
+    def _convert_object(self, obj):
+        return unified_planning_pb2.ObjectDeclaration(name=obj.name, type=str(obj.type))
+
+    @handles(unified_planning.model.FNode)
+    def _convert_fnode(self, exp):
```
alvalentini, line 91: Here we could also use a `walker` to convert the FNode

unified_planning/grpc/proto_writer.py
```diff
@@ -0,0 +1,534 @@
...
+                unified_planning_pb2.Assignment(
+                    fluent=self.convert(x), value=self.convert(v)
+                )
+                for (x, v) in problem.initial_values.items()
+            ],
+            timed_effects=[self.convert(e) for e in problem.timed_effects],
+            goals=goals,
+            features=[map_feature(feature) for feature in problem.kind.features],
+            metrics=[self.convert(m) for m in problem.quality_metrics],
+        )
```
alvalentini, line 359: Here we discard the `_initial_defaults` attribute

unified_planning/grpc/proto_writer.py
```diff
@@ -0,0 +1,534 @@
...
+
+    @handles(unified_planning.model.metrics.MinimizeActionCosts)
+    def _convert_minimize_action_costs(self, metric):
+        action_costs = {}
+        for action, cost in metric.costs.items():
+            # TODO: Add default cost
```
alvalentini, line 365: This TODO has to be fixed

unified_planning/grpc/proto_writer.py
```diff
@@ -0,0 +1,534 @@
...
+            instance = self._convert_action_instance(
+                a[1], start_time=start_time, end_time=end_time
+            )
+            action_instances.append(instance)
+
+        return unified_planning_pb2.Plan(actions=action_instances)
+
+    @handles(unified_planning.solvers.PlanGenerationResult)
+    def _convert_plan_generation_result(self, result):
+        # FIXME: Logs and metrics are not supported yet
```
alvalentini, line 453: Same as above

alvalentini (changes requested):
I left some comments.
I think it would be nice also to add type annotations to the new methods.

unified_planning/grpc/proto_writer.py
arbimo: No, it is included in the fluents, which is why the fluent conversion requires the problem as an additional argument.

alvalentini: Not really. We have two different kind of defaults: one is the fluents default (`_fluents_defaults` attribute) and you are including it; the other one (`_initial_defaults` attribute) associate a default to the different `Type`, and can be set as a parameter in the `Problem` constructor (e.g. `problem = Problem('name', {BoolType(): False})`).

alvalentini: I merged the PR #138

unified_planning/grpc/proto_writer.py
alvalentini: @mikand what do you think here, about using a walker for the FNode conversion?
Anyhow the current implementation works.

unified_planning/grpc/proto_writer.py
arbimo: Thanks, I rebased the PR so the fix is now in this branch.

unified_planning/grpc/proto_writer.py
arbimo: I added the default cost in the protobuf as well as the conversion. This is a quick fix, awaiting for the result of #134

arbimo:
One comment regarding the type annotations: I does not seem to play well at all with the generated bindings and adding type annotation would typically make mypy trigger many false positive which in turn requires deactivating type checking for the affected methods. 
So in the end type annotations would mostly clutter the source code with no benefits :/

alvalentini:
@arbimo I set the grpc requirements as `extra requirements`.

If this is OK for you, I will proceed immediately with the merge.

PR #119, opened 2022-04-01
Python writer debug

Framba-Luca:


unified_planning/io/python_writer.py
```diff
@@ -268,11 +273,30 @@ def write_problem_code_to_file(self, filename: str):
             self._write_problem_code(f)
 
 
-def _print_python_type(type: 'up.model.types.Type') -> str:
+def _get_mangled_name(original_name: str, names_mapping: Dict[str, str]) -> str:
+    '''Important note: This method updates the nammes_mapping '''
+    new_name: Optional[str] = names_mapping.get(original_name, None)
+    if new_name is None: # The name is not in the map, so it must be added
+        if not original_name.isidentifier(): # Make the name a python identifier
+            new_name = re.sub('[^0-9a-zA-Z_]', '_', f'{original_name}')
```
mikand, line 281: Unfortunately, this is not enough. Try to add a test with an action or fluent called `4theglory` or even just `5`

unified_planning/io/python_writer.py
```diff
@@ -137,8 +142,10 @@ def __init__(self, problem: 'up.model.Problem'):
 
     def _write_problem_code(self, out: IO[str]):
 
-        converter = ConverterToPythonString(self.problem.env)
+        names_mapping: Dict[str, str] = {}
```
mikand, line 145: In order to avoid weird renamings, we could initialize this map with the valid names:
```
for all names n in self.problem:
  if is_valid_python_name(n):
    names_mapping[n] = n
```

unified_planning/io/python_writer.py
```diff
@@ -268,11 +273,30 @@ def write_problem_code_to_file(self, filename: str):
...
+    '''Important note: This method updates the nammes_mapping '''
+    new_name: Optional[str] = names_mapping.get(original_name, None)
+    if new_name is None: # The name is not in the map, so it must be added
+        if not original_name.isidentifier(): # Make the name a python identifier
+            new_name = re.sub('[^0-9a-zA-Z_]', '_', f'{original_name}')
+        else:
+            new_name = original_name
+        test_name = new_name # Init values
+        count = 0
+        while test_name in names_mapping.values(): # Loop until we find a fresh name #NOTE is it a problem to always call names_mapping.values()? Is it better to define something like "invalid_names = [names_mapping.values()]" and use invalid_names instead?
```
mikand, line 286: This is OK, you can remove the note

mikand (changes requested):
@Framba-Luca good work, but I think there might be some corner cases to handle.

unified_planning/io/python_writer.py
Framba-Luca: Actually this should not be a problem, because we always "'type_name'_" before the actual name, so it would result in "fluent_4theglory". But to make sure I will add a specific test; better safe than sorry

mikand: Ah, you are right. Just to be sure, add an assertion that original_name always starts with a `[a-zA-Z_]` (it is the case now, but failing to meet this constraint would beak this function and cause assertion error in line 292)

mikand (approved):

PR #120, opened 2022-04-01
Add ENHSP in windows CI

alvalentini:


mikand:
This is the issue: https://stackoverflow.com/questions/22251809/python-select-select-on-windows
We need to find a way to avoid the select or create an alternative implementation foe Windows.

mikand:
Something like the following should be platform independent (and Python 3.7+)
```python
#!/usr/bin/env python3
import asyncio
from os import terminal_size
import sys
from asyncio.subprocess import PIPE, STDOUT
import time

ENHSP_CMD = "java -jar enhsp.jar -o domain.pddl -f instance_2.pddl -planner opt-hrmax".split(" ")


async def run_command(cmd, timeout=None):
    start = time.time()
    process = await asyncio.create_subprocess_exec(*cmd, stdout=PIPE, stderr=PIPE)

    res = "OK"
    data = [[], []] #stdout, stderr
    while True:
        lines = ["", ""]
        for idx, stream in enumerate([process.stdout, process.stderr]):
            try:
                lines[idx] = await asyncio.wait_for(stream.readline(), 1.0)
            except asyncio.TimeoutError:
                pass
            
        if not lines[0] and not lines[1]: # EOF
            break
        else:
            for idx in range(2):
                data[idx].append(lines[idx].decode())

        if time.time() - start >= timeout:
            process.kill() # Timeout or some criterion is not satisfied
            res = "TimeoutError"
            break

    await process.wait() # Wait for the child process to exit
    return res, data


if sys.platform == "win32":
    loop = asyncio.ProactorEventLoop() # For subprocess' pipes on Windows
    asyncio.set_event_loop(loop)
else:
    loop = asyncio.get_event_loop()

res, data = loop.run_until_complete(run_command(ENHSP_CMD, timeout=10))
loop.close()

if res == "TimeoutError":
    print("TIMEOUT!!")

print("STDOUT:")
print("".join(data[0]))

print("STDERR:")
print("".join(data[1]))
```

mikand:
@Framba-Luca I meant a different fix: I would replace `\r\n` **inside** the PDDLPlanner, not in the tests :)

unified_planning/solvers/pddl_solver.py
```diff
@@ -82,44 +89,22 @@ def solve(self, problem: 'up.model.Problem',
...
-                            proc_out.append(out_str)
-                            last_red_out = len(out_in_bytes)
-                        else:
-                            proc_err.append(out_str)
-                            last_red_err = len(out_in_bytes)
-                proc.wait()
+
+            if sys.platform == "win32":
+                loop = asyncio.ProactorEventLoop() # For subprocess' pipes on Windows
+                asyncio.set_event_loop(loop)
```
mikand, line 95: Is this line needed?

PR #125, opened 2022-04-13
[RFC] HTN problem specification

arbimo:
Now that the protobuf implementation is almost finished (#112 thanks to @uwe-koeckemann and @franklinselva), it is time to think of extending the initial proposal with hierarchical planning.
This PR extends the protobuf format to support hierarchical planning problems. It is intended as a first design step before the actual implementation that should add the corresponding python API and converters.

The design is orthogonal and purely additive with respect to the previous one for generative planning. It essentially consists of additional fields in the problem to:
 - declare the abstract tasks
 - declare the methods
 - declare an initial task network

There are in addition dedicated message for the HTN datastructures (`AbstractTaskDeclaration`, `Method` and `Task`).
A subtle change is in expressions and timings that should allow referring to the start/end of a subtask. This allows to state ordering constraints and referring to intermediate timepoints in conditions. (Note that this mechanism could also be used to state ordering constraints among actions in a plan but is beyond the scope of this proposal). @mikand I would be interested in having your input on this as it clearly overlaps with the temporal planning features.

Design options:
- group the `conditions` and `constraints` in methods. This deviates from what is done in HDDL but is closer to how constraints (.i.e. conditions on "static" variables) are expressed in actions).
- group the three declarations in the `Problem` in a single object to better scope them and limit the number of fields in `Problem`
- have a a list of initial task networks instead of single one. This again departs from the HDDL but is more flexible if we want to model oversubscription planning (i.e. associating a cost to each task network).

Future work:

- Implementation in python (requires #112 to have landed)
- Plan representation (not strictly needed and can be left as future work)
- Fine-grained features and associated documentation (for instance, total-order in task networks)


uwe-koeckemann:
Sorry this took me some time. I think it looks very good!

Some comments on how parts of this proposal could carry over to other things:
- The Task id field would carry over nicely to partially-ordered plans which could also be represented by (id, action instance) pairs and constraints.
- Optional ids on conditions and effects of (temporal) actions could allow to impose constraints on condition- and effect-intervals for timeline-based planners.

unified_planning/io/python_writer.py
```diff
@@ -276,6 +290,48 @@ def _write_problem_code(self, out: IO[str]):
                 raise NotImplementedError
             out.write(')\n')
 
+        if self.problem.kind.has_hierarchical():  # type: ignore
+            assert isinstance(self.problem, up.model.htn.hierarchical_problem.HierarchicalProblem)  # type: ignore
```
alvalentini, line 294: We removed almost all `#type: ignore` and we are trying to avoid them.

For the first one, we disabled the type checking for the `ProblemKind` class.

unified_planning/model/__init__.py
```diff
@@ -32,6 +32,7 @@
 from unified_planning.model.types import Type, TypeManager
 from unified_planning.model.variable import Variable, FreeVarsOracle
 import unified_planning.model.metrics
+#import unified_planning.model.htn
```
alvalentini, line 35: Leftover?

unified_planning/model/timing.py
```diff
@@ -39,11 +39,12 @@ def __init__(self, kind: TimepointKind, container=None):
         ----------
         kind: TimepointKind
           Kind of the timepoint.
-        container: Optional[Subtask]
-          Container in which the timepoint is defined.
+        container: Optional[str]
+          Identifier of the container in which the timepoint is defined.
           If not set, then a start/end timepoint refers to the enclosing action or method.
           The container must have an `identifier` property that is a string uniquely identifying it.
```
alvalentini, line 45:
```suggestion
```

alvalentini (changes requested):
It looks good to me! I left only a few very minor comments.

mikand (approved):
@arbimo This looks very good. My only question is about the `container` attribute of the `TimePoint` class: we are planning of using that class also for STNs and we need to refer to start and ends of `ActionInsatnce`s. The change as it is now requires us to add an indentifier to the `ActionInstance` or keep a (private to the STN) map from names to `ActionInstances`. This is possible, but I was wandering if you have a better idea.

unified_planning/model/timing.py
arbimo: Done

alvalentini:
@arbimo I rebased again the branch because in the meantime I merged another branch and I fixed the `# type: ignore`.

If you don't see any issues, I can merge this PR now.

unified_planning/io/python_writer.py
arbimo: That's exactly what I tried but that was before rebasing. So it was probably be fixed in the commit added with the rebase.

PR #130, opened 2022-04-13
Add a CI step that checks whether the protobuf definition is valid.

arbimo:
It also fixes several problems in the protobuf definition that were introduced when metrics were added.

@alvalentini For metrics, I inlined the `MetricBody` in the main message, mostly due to the fact the `map` structures cannot appear in `oneof` structures.
While I was at it, I added a TODO related to the documentation of action costs, in particular to state whether it is allowed to refer to action parameters in such detached cost expressions.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -185,7 +185,7 @@ jobs:
 
   deploy-pypi:
     runs-on: ubuntu-latest
-    needs: [run-tests-ubuntu, run-tests-windows, run-tests-macos] # We only deploy if the tests were successful
+    needs: [run-tests-ubuntu, run-tests-windows, run-tests-macos, check-protobuf] # We only deploy if the tests were successful
     if: github.ref == 'refs/heads/master' # We only deploy on master commits
 
     steps:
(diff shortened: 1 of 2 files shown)


alvalentini:
@arbimo thanks for the fix! I didn't know that limitation.

About the TODO, I think that we should definitely allow the use of action parameters, as well as state-dependent costs (that in my opinion should refer to the start of the action). @mikand @arbimo what do you think?

alvalentini:
I see the documentation of the action costs in the proto file related to a more general activity in the UP #133.
I think they can be addressed together in a dedicated PR.

Another TODO related to the proto file (taken by a @arbimo's comment in #78) is:

> extend the Feature enum with all existing features and make the field annotation map to the right feature name.

I think also this point can be addressed in a dedicated PR.

@arbimo @mikand If you agree, I think we can merge this PR.

arbimo:
I agree that this should be addressed elsewhere, I just felt like I had to mention the issue while touching the problematic lines.
I'll open a separate issue for action cost expression.

Regarding the alignment of features in protobuf/upf it should be done as a part of #112 

So for me, this can be merged as it is :+1:

alvalentini (approved):

PR #131, opened 2022-04-13
Fix asyncio on google colab

mikand:


unified_planning/solvers/pddl_solver.py
```diff
@@ -79,21 +97,24 @@ def solve(self, problem: 'up.model.Problem',
...
         with tempfile.TemporaryDirectory() as tempdir:
-            domanin_filename = os.path.join(tempdir, 'domain.pddl')
+            domain_filename = os.path.join(tempdir, 'domain.pddl')
             problem_filename = os.path.join(tempdir, 'problem.pddl')
             plan_filename = os.path.join(tempdir, 'plan.txt')
-            w.write_domain(domanin_filename)
+            w.write_domain(domain_filename)
             w.write_problem(problem_filename)
-            cmd = self._get_cmd(domanin_filename, problem_filename, plan_filename)
+            cmd = self._get_cmd(domain_filename, problem_filename, plan_filename)
```
alvalentini, line 106: I would add this code here and remove it from the ` run_command_unix_select` function
```suggestion
            if output_stream is None:
                proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
                try:
                    out_err_bytes = proc.communicate(timeout=timeout)
                    proc_out, proc_err = [out_err_bytes[0].decode()], [out_err_bytes[1].decode()]
                except subprocess.TimeoutExpired:
                    timeout_occurred = True
                retval = proc.returncode
            else:
```

unified_planning/solvers/pddl_solver.py
```diff
@@ -142,8 +167,73 @@ async def run_command(cmd, timeout: Optional[float] = None, output_stream: Optio
...
+    proc_err_buff: List[str] = []
+
+    process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    timeout_occurred: bool = False
+    if output_stream is None:
+        try:
+            out_err_bytes = process.communicate(timeout=timeout)
+            proc_out, proc_err = [out_err_bytes[0].decode()], [out_err_bytes[1].decode()]
+        except subprocess.TimeoutExpired:
+            timeout_occurred = True
```
alvalentini, line 197: This is the code I think can be moved above

unified_planning/solvers/pddl_solver.py
mikand: Done

PR #132, opened 2022-04-14
Implemented an ANML writer and some tests

Framba-Luca:


unified_planning/io/anml_writer.py
```diff
@@ -0,0 +1,329 @@
...
+    def walk_real_constant(self, expression, args):#NOTE is this right? Same ANML Implementation
+        assert len(args) == 0
+        frac = expression.constant_value()
+
+        with localcontext() as ctx:
+            ctx.prec = self.DECIMAL_PRECISION
+            dec = frac.numerator / Decimal(frac.denominator, ctx)
+
+            if Fraction(dec) != frac:
+                warn("The ANML printer cannot exactly represent the real constant '%s'" % frac)
```
alvalentini, line 112: real constants in the UP are Fraction

unified_planning/io/anml_writer.py
```diff
@@ -0,0 +1,329 @@
...
+        out = StringIO()
+        self._write_problem(out)
+        return out.getvalue()
+
+    def write_problem(self, filename: str):
+        '''Dumps to file the ANML problem.'''
+        with open(filename, 'w') as f:
+            self._write_problem(f)
+
+def _convert_effect(effect: 'up.model.Effect', converter: ConverterToANMLString, timing: 'up.model.Timing' = None) -> str:
```
alvalentini, line 257: I would move these functions as class methods

unified_planning/io/anml_writer.py
```diff
@@ -0,0 +1,329 @@
...
+    results: List[str] = []
+    if effect.is_conditional():
+        results.append(f'when [ all ] {converter.convert(effect.condition)}\n{{')
```
alvalentini, line 260: I think the correct way to convert conditional effects is:
```suggestion
    if effect.is_conditional():
        results.append(f'when [ {_convert_anml_timing(timing)} ] {converter.convert(effect.condition)}\n{{')
```

unified_planning/model/problem.py
```diff
@@ -385,6 +385,35 @@ def user_types_hierarchy(self) -> Dict[Optional['up.model.types.Type'], List['up
...
+    def all_types(self) -> Iterator['up.model.types.Type']:
+        '''Returns an Iterator of all the types present in the problem, considering UserTypes, IntTypes
+            and RealTypes.
+
+        IMPORTANT NOTE: this property does some computation, so it should be called as
+            minimum time as possible.'''
+        #NOTE this could also return the list "returned_types"
+        # Also a set could be used, but we lose order certainty. Maybe we can use the "OrderedDict" trick?
+        # But then it need to be made as a list before returning.
+        returned_types: List['up.model.types.Type'] = []
```
alvalentini, line 398: I think a set is the best solution here

alvalentini (changes requested):
I added some minor comments

unified_planning/io/anml_writer.py
mikand: In ANML you can write `frac.numerator/frac.denominator` and it should work I think

unified_planning/io/anml_writer.py
```diff
@@ -0,0 +1,329 @@
...
+from unified_planning.exceptions import UPTypeError, UPProblemDefinitionError
+from typing import IO, Dict, List, Optional, cast
+from io import StringIO
+from functools import reduce
+
+
+class ConverterToANMLString(walkers.DagWalker):
+    '''Expression converter to an ANML string.'''
+
+    DECIMAL_PRECISION = 10 # Number of decimal places to print real constants
```
mikand, line 38: In ANML this should not be needed

unified_planning/io/anml_writer.py
```diff
@@ -0,0 +1,329 @@
...
+        walkers.DagWalker.__init__(self)
+        self.simplifier = walkers.Simplifier(env)
+
+    def convert(self, expression):
+        '''Converts the given expression to a ANML string.'''
+        return self.walk(self.simplifier.simplify(expression)) #NOTE maybe the converter could remove the first and last char, if they are '(' and ')'
+
+    def walk_exists(self, expression, args):
+        assert len(args) == 1
+        vars_string_list = [f'{str(cast(_UserType, v.type).name)} {v.name}' for v in expression.variables()]
```
mikand, line 50:
```suggestion
        vars_string_gen = (f'{str(cast(_UserType, v.type).name)} {v.name}' for v in expression.variables())
```

unified_planning/io/anml_writer.py
```diff
@@ -0,0 +1,329 @@
...
+        return f'(exists({", ".join(vars_string_list)}) {{ {args[0]} }})' #NOTE i assumed vars are divided by ', '. #TODO check it
+
+    def walk_forall(self, expression, args):
+        assert len(args) == 1
+        vars_string_list = [f'{str(cast(_UserType, v.type).name)} {v.name}' for v in expression.variables()]
```
mikand, line 55: Same

unified_planning/io/anml_writer.py
```diff
@@ -0,0 +1,329 @@
...
+        # Init types_mapping.
+        types_mapping[self.problem.env.type_manager.IntType()] = 'integer'
+        types_mapping[self.problem.env.type_manager.RealType()] = 'rational'
+        for t in self.problem.user_types:
+            ut = cast(_UserType, t)
+            if _is_valid_anml_name(ut.name): # No renaming needed
+                types_mapping[t] = ut.name
+
+        for t in self.problem.all_types:
+            #NOTE If ANML does not accept a re-defintion of 'integer' or 'rational' or maybe also 'boolean', they must be skipped
```
mikand, line 165: Not sure to understand the note: `integer`, `float`, `rational`, `boolean` are reserved keywords in ANML...

unified_planning/io/anml_writer.py
```diff
@@ -0,0 +1,329 @@
...
+            elif t.is_user_type(): # and cast(_UserType, t).father is not None:
+                # For construction in the Problem, the father of a UserType is always added before the UserType itself.
+                father = cast(_UserType, t).father
+                assert father is not None
+                assert types_mapping[father] is not None
+                out.write(f' < {types_mapping[father]};\n')
+            elif (t.is_int_type() or t.is_real_type()) and cast(_IntType, t).lower_bound is None and cast(_IntType, t).upper_bound is None: #If first NOTE is True, this if is impossible
+                out.write(';\n')
+            elif t.is_int_type() and cast(_IntType, t).upper_bound is None: # and t.lower is not None
+                out.write(f' < integer [{str(cast(_IntType, t).lower_bound)}, infinity);\n')
```
mikand, line 179: What is this syntax? There is no need for sub-typing here just "integer[123, infinity]" represents the type.

unified_planning/io/anml_writer.py
```diff
@@ -0,0 +1,329 @@
...
+            elif t.is_real_type() and cast(_RealType, t).lower_bound is None: # and t.upper is not None
+                out.write(f' < rational (-infinity, {str(cast(_RealType, t).upper_bound)}];\n')
+            elif t.is_real_type(): # and t.upper is not None and t.upper is not None
+                out.write(f' < rational [{str(cast(_RealType, t).lower_bound)}, {str(cast(_RealType, t).upper_bound)}];\n')
+            else:
+                raise NotImplementedError
+
+        for f in self.problem.fluents:
+            out.write(f'fluent {_get_mangled_type_name(f.type, types_mapping)} {f.name};\n')
+            #NOTE It could be nice to get the static fluents of the problem and define them as constants, if we do, also initial_values should change
```
mikand, line 195: +1 for this

unified_planning/io/anml_writer.py
```diff
@@ -0,0 +1,329 @@
...
+        return time
+    else: #timing.delay < 0
+        return f'{time} - {str(timing.delay * (-1))}'
+
+def _convert_anml_interval(interval: 'up.model.TimeInterval') -> str:
+    left_bracket = '(' if interval.is_left_open() else '['
+    right_bracket = ')' if interval.is_left_open() else ']'
+    return f'{left_bracket} {_convert_anml_timing(interval.lower)}, {_convert_anml_timing(interval.upper)} {right_bracket}'
+
+def _is_valid_anml_name(name: str) -> bool:
```
mikand, line 293: ANML has some keywords that are not valid names (e.g. "action") those should be handled. Moreover, I do not see why standard types such as "boolean" should enter here: I was expecting only user-defined types, and names of entities such as fluents should enter here and `_get_anml_valid_name`

unified_planning/test/test_anml_writer.py
```diff
@@ -0,0 +1,230 @@
...
+class TestPythonWriter(TestCase):
+    def setUp(self):
+        TestCase.setUp(self)
+        self.problems = get_example_problems()
+
+    def test_basic(self):
+        problem = self.problems['basic'].problem
+        aw = ANMLWriter(problem)
+        anml_problem = aw.get_problem()
+        self.assertIn('type boolean;\n', anml_problem)
```
mikand, line 30: NO! `boolean` is a built-in type, it must not be declared!

unified_planning/test/test_anml_writer.py
```diff
@@ -0,0 +1,230 @@
...
+            '   [ start ] (not (is_at(robot) == l_to));',
+            '   [ start ] is_at(robot) := l_to;',
+            '}',
+            'instance Location l1, l2;',
+            'instance Robot r1, r2;',
+            '[ start ] is_at(r1) := l2;',
+            '[ start ] is_at(r2) := l1;',
+            '[ end ] (is_at(r1) == l1);',
+            '[ end ] (is_at(r2) == l2);\n']
+        self.assertEqual(anml_problem, '\n'.join(expected_result))
```
mikand, line 204:
```suggestion
        expected_result = """
            type Location;
            type Robot;
            fluent Location is_at;
            action move(Robot robot, Location l_from, Location l_to) {
               [ start ] (is_at(robot) == l_from);
               [ start ] (not (is_at(robot) == l_to));
               [ start ] is_at(robot) := l_to;
            }
            instance Location l1, l2;
            instance Robot r1, r2;
            [ start ] is_at(r1) := l2;
            [ start ] is_at(r2) := l1;
            [ end ] (is_at(r1) == l1);
            [ end ] (is_at(r2) == l2);"""
        self.assertEqual(anml_problem, expected_result)
```
(I left some extra spaces, but you get the idea)

unified_planning/test/test_anml_writer.py
```diff
@@ -0,0 +1,230 @@
...
+
+    def test_robot(self):
+        problem = self.problems['robot'].problem
+        aw = ANMLWriter(problem)
+        anml_problem = aw.get_problem()
+        expected_result = [
+            'type Location;',
+            'type boolean;',
+            'type rational_0 < rational [0, 100];',
```
mikand, line 213: NO! rationals are bulilt-in!

unified_planning/test/test_anml_writer.py
```diff
@@ -0,0 +1,230 @@
...
+            'fluent boolean robot_at;',
+            'fluent rational_0 battery_charge;',
```
mikand, line 215:
```suggestion
            'fluent rational[0, 100] battery_charge;',
```

mikand (changes requested):
@Framba-Luca please cross-check the printed files with TAMER parser as the generated syntax is not totally valid (you can do `./tamer solve problem.anml` to check that it parses)

unified_planning/io/anml_writer.py
```diff
@@ -0,0 +1,330 @@
...
+        else: #timing.delay < 0
+            return f'{time} - {str(timing.delay * (-1))}'
+
+    def _convert_anml_interval(self, interval: 'up.model.TimeInterval') -> str:
+        left_bracket = '(' if interval.is_left_open() else '['
+        right_bracket = ')' if interval.is_left_open() else ']'
+        return f'{left_bracket} {self._convert_anml_timing(interval.lower)}, {self._convert_anml_timing(interval.upper)} {right_bracket}'
+
+def _is_valid_anml_name(name: str) -> bool:
+    regex = re.compile('^[a-zA-Z]+.*')
```
mikand, line 289: Why you allow `.*` at the end? We can be strict here:

```suggestion
    regex = re.compile(r'^[a-zA-Z_][a-zA-Z0-9_]*')
```

Also note that when writing regular expressions it is better to use raw strings (`r"..."`) to avoid problems with escaping

unified_planning/io/anml_writer.py
```diff
@@ -0,0 +1,330 @@
...
+    if re.match(regex, name) is None or name in ANML_KEYWORDS: # If the name does not start with an alphabetic char or is a keyword
+        return False
+    return name.isidentifier() #NOTE Here I am creating a dependency between python identifiers and ANML names. For now they are (almost) the same, but in future versions?
```
mikand, line 292: If you implement the strict suggestion above, you can safely return `True` as the regexp I suggest works for both languages

unified_planning/io/anml_writer.py
```diff
@@ -0,0 +1,330 @@
...
+            else:
+                raise NotImplementedError
+
+        for t in self.problem.user_types: # Define objects
+            objects: List['unified_planning.model.Object'] = list(self.problem.objects(t))
+            obj_names = [_get_anml_name(o, names_mapping) for o in self.problem.objects(t)]
+            if len(objects) > 0:
+                out.write(f'instance {_get_anml_name(t, names_mapping)} {", ".join(obj_names)};\n')
+
+        for fe, v in self.problem.initial_values.items():
```
mikand, line 218: Just a question, are we sure to print all the initial values even when there are defaults?

mikand (changes requested):
@Framba-Luca Good work. I left some comments to fix.
@arbimo Can you also have a look? I suppose there will be some differences with your planners, maybe we can try to list them and see how to handle them

unified_planning/io/anml_writer.py
Framba-Luca: Ok perfect. The only problem that i see here is the first "_", because an "_" is not a valid first char in ANML

unified_planning/io/anml_writer.py
Framba-Luca: Isn't this the only option? Because using the * to define a default behaviour should be possible only in Tamer dialect. Right?

unified_planning/io/anml_writer.py
```diff
@@ -0,0 +1,334 @@
...
+        for o in self.problem.all_objects:
+            if _is_valid_anml_name(o.name): # No renaming needed
+                names_mapping[o] = o.name
+
+        for t in self.problem.user_types:
+            anml_type_name = _get_anml_name(t, names_mapping)
+            out.write(f'type {anml_type_name}')
+            if cast(_UserType, t).father is None:
+                out.write(';\n')
+            else: # and cast(_UserType, t).father is not None:
```
mikand, line 174:
```suggestion
            else:
```

unified_planning/io/anml_writer.py
mikand: Yes. My question was whether the `initial_values` completed the initial states if defaults are used, but I checked and it does

unified_planning/test/test_anml_writer.py
```diff
@@ -0,0 +1,252 @@
...
+}
+instance Location l1, l2;
+[ start ] robot_at(l1) := true;
+[ start ] robot_at(l2) := false;
+[ start ] battery_charge := 100;
+[ end ] robot_at(l2);
+'''
+        self.assertEqual(anml_problem, expected_result)
+
+    def test_ad_hoc_1(self):
```
mikand, line 228: @alvalentini Can you show @Framba-Luca how to use the pytamer API to parse and solve an ANML problem?

I think that a meaningful test here would be something like:
```
if tamer is installed:
for all example problems p:
  print p in ANML in a temp file file f
  ask tamer to parse and solve f
```

mikand: Another similar idea: we can create an ANML planner and use Tamer's command-line

mikand (approved):
@Framba-Luca This looks good! I left some ideas on how to test this.

I will keep this open for the moment, as it would be useful to get @arbimo feedback.

arbimo:
I looks mostly good to me with just two comments.

I see that increase/decrease effects are translated as `:+=` and `:-=` will I believe the ANML standard uses the more verbose form `:increase` and `:decrease`. I am not sure where the operators come from but it might be best to skip with the original ones. Note that this should not prevent the reader to accept this form, applying the moto "Be strict (standard compliant) in what you produce but flexible in what you accept".

A more problematic point is regarding the timing of effects. 
- In ANML, `[end] sv := v` means that the state variable `sv` takes the value `v` at time `end`
- In PDDL, `(at end (= sv v))` means that the state variable `sv` takes the value `v` **immediately after** time `end` (with a "transition period" starting at `end`)

We have not formally discussed this but I always assumed that the UPF follow the PDDL semantics (even though I have a strong preference for the ANML ones). Note that the epsilon separation in PDDL would prevent an exact encoding of many scheduling problems (e.g. even as simple as the jobshop).

I don't see this handled in the PR, which essentially means that we give different semantics depending on the target language.


Regarding testing, I agree with the proposal to go through a planner to test the problems. In particular, having a plan validation step might allow catching some diverging semantics.

alvalentini (approved):

mikand:
@arbimo Thanks for the feedback, we changed to `:increase` and `:decrease` (even though in ANML2 Dave is working on, he changed to `:+=` and `:-=`, this printer will be focused on ANML1). Concerning the time semantics, for ANML in continuous time the two semantics do match, in discrete time you are right in saying that they are different (I think we should somehow tag the semantics in the file eventually).
I am going to merge this, then we can see if the printer is not compatible enough with our planners, we will add some "dialect" parameter at the constructor.

PR #136, opened 2022-04-20
Validation grounding results objects

Framba-Luca:


unified_planning/solvers/results.py
```diff
@@ -65,9 +71,34 @@ class PlanGenerationResult:
...
+        # Check that status and error_info are consistent with eachother
+        if self.status == ValidationResultStatus.VALID and self.error_info != '':
+            raise UPUsageError(f'The Validation Result Status is {str(self.status)} but the error_info string is set.')
+
+
+@dataclass
+class GroundingResult:
+    '''Class that represents the result of a Solver.ground call.'''
+    problem: Optional[Problem]
+    rewrite_back_plan: Optional[Callable[[Plan], Plan]] #NOTE maybe "lift_plan" is a better name
```
mikand, line 97: +1 for `lift_plan`

unified_planning/solvers/results.py
```diff
@@ -65,9 +71,34 @@ class PlanGenerationResult:
...
+            raise UPUsageError(f'The Result status is {str(self.status)} but the plan is {str(self.plan)}.\nWith this status the plan must be None.')
         self.metrics = {}
         self.log_messages = [] #NOTE Here, is this init right? Since it is done after the __init__ the value might be deleted
         return self
+
+@dataclass
+class ValidationResult:
+    '''Class that represents the result of a validate call.'''
+    status: ValidationResultStatus
+    error_info: str = ''
```
mikand, line 85: Maybe we can optionally ask a validator to report the sequence of states? Something like `trace : List[State]`?
@alvalentini what do you think? We could re-use the state object we defined for simulated effects?

mikand (changes requested):
@Framba-Luca looks good!

I left some comments, but the major missing point is to update the protobuf file in the `grpc` package!

unified_planning/solvers/results.py
alvalentini: We cannot reuse the same state object because the one we defined for simulated effects is an interface, but we can define a class that implement it.

However, I think that a `trace : List[State]` makes sense only for sequential plan validation, right?

unified_planning/grpc/unified_planning.proto
```diff
@@ -482,6 +482,31 @@ message PlanGenerationResult {
...
+// Message sent by the validator.
+message ValidationResult {
+    enum ValidationResultStatus {
+        // The Plan is valid for the Problem; error_info can't be set.
+        VALID = 0;
+        // The Plan is not valid for the Problem; error_info must be set.
+        INVALID = 1;
+    }
+    ValidationResultStatus status = 1;
+    // Optional. Informations about why the Plan is not valid fo the Problem
```
mikand, line 494:
```suggestion
    // Optional. Information about why the Plan is not valid for the Problem
```

unified_planning/solvers/results.py
mikand: You are right, I'll open an issue and we postpone this.

unified_planning/test/test_transformer.py
```diff
@@ -36,5 +36,5 @@ def test_transformer(self):
             t.get_transformed_actions(a)
         with self.assertRaises(NotImplementedError):
             t.get_rewritten_problem()
-        with self.assertRaises(NotImplementedError):
+        with self.assertRaises(AttributeError):
```
mikand, line 39: I would say that this is not a meaningful test :smile:

unified_planning/transformers/transformer.py
```diff
@@ -46,27 +46,18 @@ def get_rewritten_problem(self) -> Problem:
...
-                old_actions.append(ActionInstance(self.get_original_action(ai.action), ai.actual_parameters))
-            return SequentialPlan(old_actions)
-        elif isinstance(plan, TimeTriggeredPlan):
-            s_new_actions_d = plan.actions
-            s_old_actions_d = []
-            for s, ai, d in s_new_actions_d:
-                s_old_actions_d.append((s, ActionInstance(self.get_original_action(ai.action), ai.actual_parameters), d))
-            return TimeTriggeredPlan(s_old_actions_d)
-        raise NotImplementedError
+        This method uses self._replace_action_instance; which MUST be rewritten if the specific Transformer extension changes
```
mikand, line 58: Is it a MUST or a MAY? You only override it in the grounder...

mikand (approved):
Very good job @Framba-Luca .
I assume you also aligned the pyperplan grounder, right? 
I left some minor comments to be fixed, then @alvalentini you can go ahead and merge without another round of code review.

mikand:
@arbimo This PR has a slight overlap with #112 . Do you prefer that we wait for #112 for merging this?

arbimo:
> @arbimo This PR has a slight overlap with #112 . Do you prefer that we wait for #112 for merging this?

That would be great, mostly because then the responsibility of writing the protobuf converters and tests would rely on someone else than me =)
Jokes appart, if #112 can be reviewed promptly and does not necessite many iterations then I think it might make sense to wait.

As food for thought, #112 extends the Result message with an `Engine` message that I think would also make sense in the new results here (in fact, I named it engine precisely with this in mind).

unified_planning/grpc/unified_planning.proto
```diff
@@ -531,6 +531,31 @@ message Engine {
...
+message ValidationResult {
+    enum ValidationResultStatus {
+        // The Plan is valid for the Problem; error_info can't be set.
+        VALID = 0;
+        // The Plan is not valid for the Problem; error_info must be set.
+        INVALID = 1;
+    }
+    ValidationResultStatus status = 1;
+    // Optional. Information about why the Plan is not valid for the Problem
+    string error_info = 2;
```
arbimo, line 544: Maybe this could reuse the log message of plan results. I think the 4 verbosity level there are equally applicable to validation output (and we definitly want to allow having detailed feedback from the validator).

arbimo (commented):
Just a quick comment on the `.proto`. 

Regarding the generation of the bindings, we currently check in CI there that the binding present are the ones that would be generated from `protoc 3.6.1` (the version available on ubuntu 20.04). This is mostly an arbitrary choice that can be changed but should ensure that the protobuf bindings do not change depending on which developer generated them.

unified_planning/solvers/parallel.py
```diff
@@ -72,33 +71,18 @@ def _run_parallel(self, fname, *args):
             if isinstance(res, BaseException):
                 raise res
             else:
-                assert isinstance(res, up.solvers.PlanGenerationResult)
+                assert isinstance(res, up.solvers.result.Result)
                 # If the planner is sure about the result (optimality of the result or impossibility of the problem or the problem does not need optimality) exit the loop
-                if res.status == PlanGenerationResultStatus.SOLVED_OPTIMALLY or res.status == PlanGenerationResultStatus.UNSOLVABLE_PROVEN or (res.status == PlanGenerationResultStatus.SOLVED_SATISFICING and not optimality_required):
+                if res.is_definitive_result(args[0]):
```
alvalentini, line 76:
```suggestion
                if res.is_definitive_result(*args):
```

alvalentini (changes requested):

alvalentini (approved):

mikand (changes requested):
@Framba-Luca  Looks good, but please add some tests as the coverage decreases quite a lot with this PR

unified_planning/grpc/proto_reader.py
```diff
@@ -524,13 +524,13 @@ def _convert_plan_generation_result(self, result: proto.PlanGenerationResult, pr
...
-        # TODO: Extend the protobuf convertors to handle metrics and logs in results
+        # TODO: Extend the protobuf convertors to handle metrics and log_messages in results
         return unified_planning.solvers.PlanGenerationResult(
             status=status,
             plan=self.convert(result.plan, problem),
-            planner_name=result.planner.name,
+            engine_name=result.engine.name,
             # metrics=result.metrics,
-            # log_messages=[self.convert(log) for log in result.logs],
+            # log_messages=[self.convert(log) for log in result.log_messages],
```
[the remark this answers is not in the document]
arbimo: @franklinselva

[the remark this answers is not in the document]
franklinselva: The updated `PlanGenerationResult` is now lined up with the proto convertors.

PR #137, opened 2022-04-20
Added a __repr__ method to ProblemKind class

Framba-Luca:


unified_planning/model/problem_kind.py
```diff
@@ -52,6 +52,9 @@ class ProblemKind(metaclass=ProblemKindMeta):
     def __init__(self, features: Set[str] = set()):
         self._features: Set[str] = set(features)
 
+    def __repr__(self) -> str:
+        return f'ProblemKind features = [{", ".join(self._features)}]'
```
alvalentini, line 56: I think it is better:
```suggestion
        return f'[{", ".join(self._features)}]'
```

mikand:
@Framba-Luca 3 things:
1. Also add a __str__ method
2. In python (generally) the __repr__ method returns a representation that is valid python code, s here we could print someting like "ProblemKind([<features>])"?
3. I would like to have a better readable representation in the __str__ method, even multi-line if possible

PR #138, opened 2022-04-28
Autopromote initial defaults to fluent defaults on fluent creation.

arbimo:
The problem class currently contains an initial default parameter that can be set in the constructor.

This is essentially useful to support some classes of problems (especially in PDDL), however it has several shortcomings:

 - it can only be set at problem creation, which means it can only be used for primitive types.
 - it is redundent with the fluent defaults. To know whether a given fluent has a default value, one would to check into two places.

This PR essentially, keeps the initial defaults construct as it is useful in a few contexts but modifies its usage.
However it changes its usage: whenever a new fluent is created with no explicit default, we check in the initial default whether this type has a global default and if so we associate it with the fluent.

This does not change any behavior but ensures that the default values are stored in a single place.

### DIFF
--- a/unified_planning/model/problem.py
+++ b/unified_planning/model/problem.py
@@ -42,7 +42,7 @@ class Problem:
         self._timed_goals: Dict['up.model.timing.TimeInterval', List['up.model.fnode.FNode']] = {}
         self._goals: List['up.model.fnode.FNode'] = list()
         self._metrics: List['up.model.metrics.PlanQualityMetric'] = []
-        # The field initial default optionally associates a non-user type to a default value. When a new fluent is
+        # The field initial default optionally associates a type to a default value. When a new fluent is
         # created with no explicit default, it will be associated with the initial-default of his type, if any.
         self._initial_defaults: Dict['up.model.types.Type', 'up.model.fnode.FNode'] = {}
         for k, v in initial_defaults.items():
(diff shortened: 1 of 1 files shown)


unified_planning/model/problem.py
```diff
@@ -42,6 +42,8 @@ def __init__(self, name: str = None, env: 'up.environment.Environment' = None, *
         self._timed_goals: Dict['up.model.timing.TimeInterval', List['up.model.fnode.FNode']] = {}
         self._goals: List['up.model.fnode.FNode'] = list()
         self._metrics: List['up.model.metrics.PlanQualityMetric'] = []
+        # The field initial default optionally associates a non-user type to a default value. When a new fluent is
```
alvalentini, line 45: Why only `non-user` type?

alvalentini:
Apart the very minor comment, I think this PR is good and solves the issue in #112.

@mikand what do you think?

mikand:
LGTM! @alvalentini merge at will! :smile:

unified_planning/model/problem.py
alvalentini: @arbimo I updated your comment removing the `non-user`, since it is possible to define the default also for a `UserType`.

arbimo: That's right, thank you.

PR #139, opened 2022-04-29
CI colab testing

Framba-Luca:


test_colab.sh
```diff
@@ -0,0 +1,32 @@
...
+            trim_rows="true"
+        elif [[ "${line}" == *"${eoi}"* ]]
+        then
+            trim_rows="false"
+        elif [ "${trim_rows}" = "false" ]
+        then
+            echo "${line}" >> ${file_tested_name}
+        fi
+    done < "${python_file}"
+    ipython3 ${file_tested_name}
```
mikand, line 25: It would be nice to only show the output of a failing execution instead of the whole output

test_colab.sh
```diff
@@ -0,0 +1,32 @@
...
+    if [ ! $? == 0 ]; then
+        rm ./unified_planning/notebook/*.py
```
mikand, line 27: No need to clean after yourself in CI, but OK

test_colab.sh
```diff
@@ -0,0 +1,32 @@
...
+        exit 1
```
mikand, line 28: You can Avoid the entire if by using `set -e` (https://www.delftstack.com/howto/linux/bash-set-e/#:~:text=If%20the%20command%20returns%20a,exits%20when%20a%20command%20fails.)

mikand (commented):
@Framba-Luca Good work! I left some small improvements suggestions

test_colab.sh
```diff
@@ -0,0 +1,36 @@
...
+        then
+            echo "${line}" >> ${file_tested_name}
+        fi
+    done < "${python_file}"
+    echo "Testing ${file_tested_name}"
+    output="$( { ipython3 ${file_tested_name}; } 2>&1 )"
+    if [ ! $? == 0 ]; then
+        rm ./unified_planning/notebook/*.py
+        echo "Testing ${file_tested_name} FAILED"
+        echo ${output}
```
mikand, line 30: You need quotes to keep the newlines in printing (see https://stackoverflow.com/questions/22101778/how-to-preserve-line-breaks-when-storing-command-output-to-a-variable)
```suggestion
        echo "${output}"
```

test_colab.sh
```diff
@@ -0,0 +1,36 @@
...
+        elif [[ "${line}" == *"${eoi}"* ]]
+        then
+            trim_rows="false"
+        elif [ "${trim_rows}" = "false" ]
+        then
+            echo "${line}" >> ${file_tested_name}
+        fi
+    done < "${python_file}"
+    echo "Testing ${file_tested_name}"
+    output="$( { ipython3 ${file_tested_name}; } 2>&1 )"
```
mikand, line 26: Backticks are a shorter syntax, but this change is optional (it just looks nicer to me)

```suggestion
    output=`ipython3 ${file_tested_name} 2>&1`
```

test_colab.sh
```diff
@@ -0,0 +1,36 @@
...
+            trim_rows="true"
+        elif [[ "${line}" == *"${eoi}"* ]]
+        then
+            trim_rows="false"
+        elif [ "${trim_rows}" = "false" ]
+        then
+            echo "${line}" >> ${file_tested_name}
+        fi
+    done < "${python_file}"
+    echo "Testing ${file_tested_name}"
```
mikand, line 25:
```suggestion
    echo -n "  -> Testing ${file_tested_name}..."
```

test_colab.sh
```diff
@@ -0,0 +1,36 @@
...
+    output="$( { ipython3 ${file_tested_name}; } 2>&1 )"
+    if [ ! $? == 0 ]; then
+        rm ./unified_planning/notebook/*.py
+        echo "Testing ${file_tested_name} FAILED"
```
mikand, line 29:
```suggestion
        echo " FAILED for the following reason:"
```

test_colab.sh
```diff
@@ -0,0 +1,36 @@
...
+        echo ${output}
+        exit 1
+    fi
+    echo "Testing ${file_tested_name} DONE"
```
mikand, line 33:
```suggestion
    echo " DONE!"
```

test_colab.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+boi="# begin of installation"
+eoi="# end of installation"
+for colab_file in ./unified_planning/notebook/*.ipynb ; do
+    file_name="${colab_file%.ipynb}"
+    jupyter nbconvert --to python ${colab_file}
+    python_file="${file_name}.py"
```
mikand, line 8: Just to be sure that we convert to the right file :)
```suggestion
    python_file="${file_name}.py"
    jupyter nbconvert --to python ${colab_file} --output ${python_file}
```

mikand (approved):
@Framba-Luca Looks very good! I did some cosmetic improvements to the output of the script. 
Please move the notebook directory to top level and call it `notebooks`, then we can merge and then we can add a Readme with the colab/binder links

PR #140, opened 2022-04-30
Python API for hierarchical problems

arbimo:
This is a first attempt at the Python API for hierarchical problems.

- The `model.htn` module contains all implementations that are required for defining hierarchical problems
- It introduces a `HierarchicalProblem` that is a subclass of `Problem` (anything that is valid in a problem is also valid in a hierarchical problem)
- It extends `Timepoint` to allow refering to the start/end of a subtask
- It extends `FNode` to also support containing `Timing` and expressing temporal constraints between the start/end timepoints of subtasks

An example problem is provided in the (runnable) file `test/problems/hierarchical.py`.

There are currently a few shortcomings (features have not been updated, there are no automated tests, the method API is incomplete, docstrings are missing or incorrect, there is some duplicated logic) but it should give an accurate idea of where this is going.
Given the size of this hierarchical extension and its impact on the `FNode` API, it would be great to have someone take a high level look at this to check if this a desirable path to follow (@mikand, @alvalentini ?).

Once #125 is merged, it will be extended to support the serialization to protobuf.


arbimo:
This is now complete. Compared to the initial comment, the code has been cleaned up and now support parsing basic HDDL problems.
The support for hierarchical problems does not go as far as it should be in my opinion but I think this is a good baseline.

My suggestion for continuing the work on HTN support would be follow this sequence:
 - review, finalize and merge this PR
 - update #125 with the conversion from/into protobuf (this will facilitate testing). This will increase our testing capabilites.
 - support hierarchical problems with the Aries planner (finally ^^)
 - extend support for hierarchical problem, relying on the previous two steps to facilitate testing.

@mikand @alvalentini what to you think?

unified_planning/model/htn/method.py
```diff
@@ -0,0 +1,235 @@
...
+from collections import OrderedDict
+from typing import List, Union, Optional
+
+import unified_planning as up
+from unified_planning.environment import get_env, Environment
+from unified_planning.exceptions import UPUnboundedVariablesError, UPValueError
+from unified_planning.model import Timing, Parameter
+from unified_planning.model.action import Action
+from unified_planning.model.htn.task import Task, Subtask
+from unified_planning.model.timing import Timepoint
```
alvalentini, line 29: Inside a package it is better to do the imports explicitly:
```suggestion
from unified_planning.model.parameter import Parameter
from unified_planning.model.action import Action
from unified_planning.model.htn.task import Task, Subtask
from unified_planning.model.timing import Timing, Timepoint
```

This comment also applies in the files below.

unified_planning/model/htn/task_network.py
```diff
@@ -0,0 +1,116 @@
...
+        param = Parameter(name, typename)
+        self._variables[name] = param
+        return param
+
+    @property
+    def subtasks(self) -> List['Subtask']:
+        """Returns the list of the subtasks."""
+        return self._subtasks
+
+    def add_subtask(self, task: Union[Subtask, Action, Task], *args, ident: Optional[str] = None) -> Subtask:
```
alvalentini, line 81: Missing `args` type annotation

unified_planning/model/timing.py
```diff
@@ -28,28 +28,49 @@ class TimepointKind(Enum):
 
 
 class Timepoint:
-    def __init__(self, kind: TimepointKind):
+    def __init__(self, kind: TimepointKind, container=None):
```
alvalentini, line 31: Missining `container` type annotation.

unified_planning/model/timing.py
```diff
@@ -28,28 +28,49 @@ class TimepointKind(Enum):
...
+        """Creates a new timepoint.
+
+        Parameters
+        ----------
+        kind: TimepointKind
+          Kind of the timepoint.
+        container: Optional[Subtask, ActionInstance]
+          Container in which the timepoint is defined.
+          If not set, then a start/end timepoint refers to the enclosing action or method.
+          The container must have an `identifier` property that is a string uniquely identifying it.
```
alvalentini, line 41: `ActionInstance` does not have an `identifier` property

unified_planning/model/timing.py
```diff
@@ -28,28 +28,49 @@ class TimepointKind(Enum):
...
     def __eq__(self, oth: object) -> bool:
         if isinstance(oth, Timepoint):
-            return self._kind == oth._kind
+            return self._kind == oth._kind and self._container == oth._container
         else:
             return False
 
     def __hash__(self) -> int:
-        return hash(self._kind)
+        return hash(self._kind) + hash(self._container)
```
alvalentini, line 63: `ActionInstance` and `Subtask` do not have an `__hash__` method.

alvalentini (changes requested):
@mikand @arbimo This PR looks really good to me.
I only added a few, very minor, comments.

unified_planning/model/timing.py
arbimo: Yes, my understanding is that we cannot provide the type annotation here because it would create a circular dependency: `Subtask` uses `Timepoint` for its `start`/`end` properties so referencing the `Subtask` class here would create create a cycle in the imports.
So I relied on duck-typing and method documentation.

unified_planning/model/timing.py
arbimo: Right, I removed the reference to `ActionInstance`. 
Essentially, this is designed to eventually support having an `ActionInstance` there which I believe will be useful for STN/POP plans.
However, this is not required by this PR or hierarchical planning in general so I removed it for now.

unified_planning/model/timing.py
alvalentini: To avoid the circular dependency, you can specify the type annotation as a string.

```suggestion
    def __init__(self, kind: TimepointKind, container: Optional['Subtask'] = None):
```

unified_planning/model/timing.py
arbimo: I added hash/eq for Subtask

unified_planning/model/timing.py
arbimo: I already tried that without success :/ Could you give a try ? I might be missing some subtleties on how mypy works.

unified_planning/model/htn/method.py
```diff
@@ -0,0 +1,239 @@
...
+    def achieved_task(self) -> ParameterizedTask:
+        """Returns the task that this method achieves."""
+        assert self._task is not None
+        return self._task
+
+    def set_task(self, task: Union[Task, ParameterizedTask], *arguments: Parameter):
+        """Defines the task that is method achieves."""
+        assert self._task is None, f"Method {self.name} was already assigned a task"
+        if isinstance(task, ParameterizedTask):
+            assert all(p.name in self._parameters for p in task.parameters)
```
mikand, line 142: We can check that both the type and name match like so:
```suggestion
            assert all(p in self._parameters for p in task.parameters)
```

unified_planning/model/htn/method.py
```diff
@@ -0,0 +1,239 @@
...
+    @property
+    def achieved_task(self) -> ParameterizedTask:
+        """Returns the task that this method achieves."""
+        assert self._task is not None
+        return self._task
+
+    def set_task(self, task: Union[Task, ParameterizedTask], *arguments: Parameter):
+        """Defines the task that is method achieves."""
+        assert self._task is None, f"Method {self.name} was already assigned a task"
+        if isinstance(task, ParameterizedTask):
```
mikand, line 141: here we assume that `len(arguments) == 0`, right?

unified_planning/model/htn/method.py
```diff
@@ -0,0 +1,239 @@
...
+        """Returns the action name."""
+        return self._name
+
+    @property
+    def achieved_task(self) -> ParameterizedTask:
+        """Returns the task that this method achieves."""
+        assert self._task is not None
+        return self._task
+
+    def set_task(self, task: Union[Task, ParameterizedTask], *arguments: Parameter):
```
mikand, line 138: In this methods (and other places) asserts are ued to report errors that can arise by API misuse. I would use a `ValueError` for these cases (that is not disabled by `python -O`)

unified_planning/walkers/type_checker.py
```diff
@@ -233,7 +238,7 @@ def walk_div(self, expression, args):
     @walkers.handles(OperatorKind.LE, OperatorKind.LT)
     def walk_math_relation(self, expression, args):
         for x in args:
-            if x is None or not (x.is_int_type() or x.is_real_type()):
+            if x is None or not (x.is_int_type() or x.is_real_type() or x.is_time_type()):
```
mikand, line 241: Not sure if needed, but this only allows for `<` and `<=` relations. The equality between a time and a real constant is forbidden (see line 256/261 of this file) (that would be equivalent to the conjunction of two `<=` constraints BTW)

mikand (changes requested):
@arbimo This looks great! I left some high-level comments on the `Method` implementation. FOr the rest, I think the only issue (that we can discuss in the next WP3 meeting) is how to adapt/deal with transformers because you inherited from `Problem`, we might need to adapt the transformaer code to either reject HTN problems or support them.

unified_planning/model/timing.py
alvalentini: You are right, it's harder than I thought. You can ignore this comment for the moment.

unified_planning/walkers/type_checker.py
arbimo: Yes, currently the API does not allow constructing so I did not implement them as they would not have been tested. It is definitly something that should indeed be supported in the future :+1:

unified_planning/model/htn/method.py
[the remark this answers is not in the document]
arbimo: I added a TODO referencing issue #153 that would describe the missing feature.

unified_planning/model/htn/hierarchical_problem.py
```diff
@@ -0,0 +1,107 @@
...
+            return False
+        if not isinstance(oth, HierarchicalProblem):
+            return False
+        return (self._initial_task_network == oth._initial_task_network and
+                self._methods == oth._methods and
+                self._abstract_tasks == oth._abstract_tasks)
+
+    def __hash__(self):
+        res = super().__hash__()
+        res += sum(map(hash, self._abstract_tasks.values()))
```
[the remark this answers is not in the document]
arbimo: While I love folding, I find the initial expression much clearer and less error prone.
This is a pattern that is used over and over in hash code implementations so there is a great value in having the simple pattern with no subtleties. In the reduce, expression is easy to miss a typo in one of the `+` or `res` element.

I cannot say which would be the most efficient: one creates an intermediate list and the other creates a lambda instance that must capture its environment...

Is there any advantage that I did not see?

unified_planning/model/htn/method.py
arbimo: Done

unified_planning/model/htn/method.py
arbimo: Done

unified_planning/model/htn/method.py
arbimo: Hummm, I had very similar discussion with @Framba-Luca a few days ago where I essentially declared my love of asserts : https://github.com/aiplan4eu/unified-planning/pull/141#issuecomment-1124926482

In particular here, the checks have some associated computational cost (e.g. membership of a list, type checking) so one might want to disable them if they are sure that they use the API correctly. Those are essentially programming error for which we should not try to recover but we just want to warn the user so that it does not shoot himself in the foot.

I do use specialized exception in the PDDL reader where we should be able to recover from an invalid input (and in fact I converted an assert to a `SyntaxError` after double checking.

unified_planning/model/timing.py
arbimo: Ok, I removed the annotation and tests are now passing again.

unified_planning/model/htn/method.py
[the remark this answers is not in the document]
mikand: The TODO is OK for now then. (Having the environment in FNode is indeed a good idea, we "copied" expressions from pySMT, and there there is no env inside because of efficiency reasons, but in planning expressions are generally way samller than in SMT )

unified_planning/model/htn/hierarchical_problem.py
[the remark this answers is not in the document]
mikand: A matter of taste I guess :smile: 
Let's keep your version

unified_planning/walkers/type_checker.py
mikand: OK! Can you add a `#TODO` or an issue to keep track of this?

unified_planning/model/htn/method.py
mikand: We disagree on this one :smiling_imp: 

In my opinion asserts should only be used for code invariants, something that your code guaranteed to hold. Validation of user inputs should raise a proper exception, especially in this library where we aim to make it as user-friendly as possible. (More or less I agree with this blog: https://stefan.sofa-rockers.org/2018/04/16/assertions-and-exceptions/)

Not a big deal for this round, though, we can merge and fix this later on.

mikand (approved):

mikand:
:champagne:

unified_planning/model/htn/method.py
arbimo: Ok, I guess I will play by the local rules then =)
I'll try to revisit those asserts incrementally next time I am on this part of the code.

arbimo:
:partying_face: 
@alvalentini and @mikand Thanks a lot to the two of you for the time reviewing and improving this and other pull requests !
Over all the code base is of amazing quality and it is a pleasure to hack on it !

PR #141, opened 2022-04-30
Allow add_XXX methods to act as constructors

arbimo:
This is a quick and dirty experiment for an API improvement to limit the boilerplate when creating elements that must be added to the problem and reused in other parts of the problem definition (notably objects, fluents, ...).

Manipulating such elements is always a two step process: creating the element and then adding it to the `Problem`.

```python
# Previous syntax (still supported)
pb = Problem()
Item = UserType('Item')
heavy = Fluent('heavy', Item)
o = Object('o', Item)
pb.add_object(o)
pb.add_fluent(heavy)
pb.set_initial_value(heavy(o), True)
```

This is PR is a proposal for `Problem.add_XXX` methods to support to kinds of input:
 - the already constructed element (e.g. an `Object` instance)
 - the parameters to the element's constructor, used to initialize it (e.g. the `name` and `typename` of the `Object` constructor)

Regardless of its input, it returns the element to enable its future usage.
In addition to the previous one, this allows to support the more straightforward syntax:

```python
Item = UserType('Item')
pb = Problem()
o = pb.add_object('o', Item)
heavy = pb.add_fluent('heavy', Item)
pb.set_initial_value(heavy(o), True)
```

At this point, this PR is not battle tested but is only meant to check if there is interest in this API change. This is a pattern that I use in #140 for building tasks and subtask an later referencing them.


mikand:
I like this idea very much! It simplifies the creation of problems without sacrificing retro-compatibility and the flexibility of having reusable Fluents and Objects.

@alvalentini @Framba-Luca what do you think? Can you see any criticality?

unified_planning/model/problem.py
```diff
@@ -226,10 +226,15 @@ def has_fluent(self, name: str) -> bool:
                 return True
         return False
 
-    def add_fluent(self, fluent: 'up.model.fluent.Fluent', *,
-                   default_initial_value: Union['up.model.fnode.FNode', 'up.model.object.Object', bool,
-                                                int, float, Fraction] = None):
+    def add_fluent(self, fluent_or_name: Union['up.model.fluent.Fluent', str], *args,
+                   default_initial_value: Union['up.model.fnode.FNode', 'up.model.object.Object', bool, int, float, Fraction] = None,
+                   typename: 'up.model.types.Type' = None) -> 'up.model.fluent.Fluent':
```
alvalentini, line 231:
```suggestion
    def add_fluent(self, fluent_or_name: Union['up.model.fluent.Fluent', str], 
                   typename: 'up.model.types.Type' = None, *,
                   default_initial_value: Union['up.model.fnode.FNode', 'up.model.object.Object',
                   bool, int, float, Fraction] = None
                   **kwargs: 'up.model.types.Type') -> 'up.model.fluent.Fluent':
```

unified_planning/model/problem.py
```diff
@@ -226,10 +226,15 @@ def has_fluent(self, name: str) -> bool:
...
         '''Adds the given fluent.'''
+        if isinstance(fluent_or_name, up.model.fluent.Fluent):
+            assert len(args) == 0
```
alvalentini, line 234:
```suggestion
            assert len(kwargs) == 0 and typename is None
```

unified_planning/model/problem.py
```diff
@@ -226,10 +226,15 @@ def has_fluent(self, name: str) -> bool:
...
+            fluent = fluent_or_name
+        else:
+            fluent = up.model.fluent.Fluent(fluent_or_name, typename, *args)
```
alvalentini, line 237:
```suggestion
            fluent = up.model.fluent.Fluent(fluent_or_name, typename, env=self.env, **kwargs)
```

alvalentini (changes requested):
@arbimo @mikand I like the proposal, I think it can be useful!

I added a few comment in the code.

unified_planning/model/problem.py
```diff
@@ -226,10 +226,17 @@ def has_fluent(self, name: str) -> bool:
                 return True
         return False
 
-    def add_fluent(self, fluent: 'up.model.fluent.Fluent', *,
-                   default_initial_value: Union['up.model.fnode.FNode', 'up.model.object.Object', bool,
-                                                int, float, Fraction] = None):
+    def add_fluent(self, fluent_or_name: Union['up.model.fluent.Fluent', str], 
+                   typename: 'up.model.types.Type' = None, *,
+                   default_initial_value: Union['up.model.fnode.FNode', 'up.model.object.Object',
+                   bool, int, float, Fraction] = None
```
alvalentini, line 232: @arbimo I forgot the comma in my last suggestion:
```suggestion
                   bool, int, float, Fraction] = None,
```

arbimo:
@alvalentini Thanks for the review, I added the suggestions (incorporating your suggestions).

The main downside of this proposal is that the function' signature does not immediately convey what it does. Hence I documented the two methods. The docstrings contain examples that will be automatically checked on CI (I enabled doctest in pytest).

unified_planning/model/problem.py
```diff
@@ -226,10 +226,38 @@ def has_fluent(self, name: str) -> bool:
...
+        >>> location = UserType("Location")
+        >>> at_loc = Fluent("at_loc", BoolType(), l=location)  # creates a new fluent
+        >>> problem.add_fluent(at_loc)  # adds it to the problem
+        bool at_loc[l=Location]
+        >>> problem.add_fluent("connected", BoolType(), l1=location, l2=location)  # creates a new fluent and add it to the problem.
+        bool connected[l1=Location, l2=Location]
+        >>>
+        """
+        if isinstance(fluent_or_name, up.model.fluent.Fluent):
+            assert len(kwargs) == 0 and typename is None
```
Framba-Luca, line 257: Here I would raise a meaningful Exception, so that the code is more user-fail-friendly

unified_planning/model/problem.py
```diff
@@ -385,22 +414,42 @@ def user_types_hierarchy(self) -> Dict[Optional['up.model.types.Type'], List['up
...
+        >>> from unified_planning.shortcuts import *
+        >>> problem = Problem()
+        >>> cup = UserType("Cup")
+        >>> o1 = Object("o1", cup)  # creates a new object o1
+        >>> problem.add_object(o1)  # adds it to the problem
+        o1
+        >>> o2 = problem.add_object("o2", cup)  # alternative syntax to create a new object and add it to the problem.
+        """
+        if isinstance(obj_or_name, up.model.object.Object):
+            assert typename is None
```
Framba-Luca, line 437: Same

unified_planning/model/problem.py
```diff
@@ -385,22 +414,42 @@ def user_types_hierarchy(self) -> Dict[Optional['up.model.types.Type'], List['up
...
+            obj = obj_or_name
+        else:
+            assert typename is not None, "Missing type of the object"
```
Framba-Luca, line 440: Same

Framba-Luca (commented):
I really like the idea and think it is very useful and allows faster coding and less prone to errors due to the user forgetting to add the object/fluent to the problem!

For coherence do you plan to do the same for the actions? The use would be much less and the implementation more problematic, due to the difference between InstantaneousAction and DurativeAction, but it's a point worth thinking about.

I also left 1 very minor comment in the code.

Framba-Luca (commented):
Perfect, makes sense for the Actions. 

Ye, our discrepancy about "assert or exception" originated from the fact that I prefer using exceptions also for user errors and input validation.

alvalentini (approved):

PR #142, opened 2022-05-02
Update FNode to protobuf converter with a walker

alvalentini:
I implemented a walker to convert FNode to protobuf.

@arbimo if for you this is OK, you can merge this PR.

After that we can proceed with #112


arbimo:
LGTM. Thanks for this improvement :+1:

PR #144, opened 2022-05-03
Fix proto converter

alvalentini:
@Framba-Luca noted that the `Problem` conversion method from proto to up has a `Problem` as a parameter, and this does not make any sense.

I created this PR to fix the problem.

@arbimo does it make sense to you?

### DIFF
--- a/unified_planning/grpc/proto_reader.py
+++ b/unified_planning/grpc/proto_reader.py
@@ -13,12 +13,13 @@
 # limitations under the License.
 #
 # type: ignore[attr-defined]
-from typing import Tuple, Union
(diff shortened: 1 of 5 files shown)


arbimo (approved):
Yes, I agree that it does not make sense to have the problem as parameter. and this LGTM.
@franklinselva feel to comment if there is something that we do not see in the justification for this.

By the way, there might be many little things to fix/improve on the current implementation (that has seen three different authors), feel free to ping me if there is anything on which you have a doubt

franklinselva:
Yes, I agree that the `problem` as an argument serves no use in reading back the `Problem` message.

PR #146, opened 2022-05-11
Implemented a solver's credict mechanism

Framba-Luca:


unified_planning/solvers/factory.py
```diff
@@ -97,20 +99,29 @@ def _get_solver(self, solver_kind: str, name: Optional[str] = None,
             solvers = []
             for name, param in zip(names, params):
                 SolverClass = self._get_solver_class(solver_kind, name)
+                if SolverClass is None:
+                    raise
```
mikand, line 103: This is not valid python3 code, you need an exception to raise!

unified_planning/solvers/factory.py
```diff
@@ -97,20 +99,29 @@ def _get_solver(self, solver_kind: str, name: Optional[str] = None,
...
-            return up.solvers.parallel.Parallel(solvers)
+            p_solver = up.solvers.parallel.Parallel(solvers)
+            return p_solver
         else:
             if params is None:
                 params = {}
             assert isinstance(params, Dict)
             SolverClass = self._get_solver_class(solver_kind, name, problem_kind, optimality_guarantee)
+            if SolverClass is None:
+                raise
```
mikand, line 114: Same as above

unified_planning/solvers/grounder.py
```diff
@@ -24,6 +25,13 @@
 from unified_planning.model import Problem, ProblemKind
 
 
+credits = solvers.solver.Credits('UP naive grounder',
+                                 'FBK-team',
+                                 'aiplan4eu@fbk.eu',
+                                 'https://github.com/aiplan4eu/unified-planning/solvers/grounder',
+                                 'This grounder, does basic simplification and removes impossible actions and actions with different assignment to the same variable.',
+                                 'This grounder, does basic simplification and removes impossible actions and actions with different assignment to the same variable.'
+                                )
```
mikand, line 34:
```suggestion
credits = solvers.solver.Credits('UP naive grounder',
                                 'AIPlan4EU-team',
                                 'aiplan4eu@fbk.eu',
                                 'https://github.com/aiplan4eu/unified-planning/solvers/grounder',
                                 'This grounder, does basic simplification and removes impossible actions and actions with different assignment to the same variable.',
                                 'This grounder, does basic simplification and removes impossible actions and actions with different assignment to the same variable. It implements the naive grounding algorithm for classical, numeric and temporal planning'
                                )
```

unified_planning/solvers/grounder.py
```diff
@@ -44,35 +52,44 @@ def name(self):
...
+    def supports(problem_kind):
+        return problem_kind <= Grounder.supported_kind()
 
     @staticmethod
     def is_grounder():
         return True
 
+    @staticmethod
+    def credits(stream: Optional[IO[str]] = sys.stdout, full_credits: bool = False):
+        if stream is not None:
```
mikand, line 90: Why not moving this `if` into the caller and have this function take a non-optional `IO[str]`?

unified_planning/solvers/factory.py
```diff
@@ -97,20 +99,29 @@ def _get_solver(self, solver_kind: str, name: Optional[str] = None,
             solvers = []
             for name, param in zip(names, params):
                 SolverClass = self._get_solver_class(solver_kind, name)
+                if SolverClass is None:
+                    raise
+                SolverClass.credits(credits_stream)
```
mikand, line 104:
```suggestion
                if credits_stream is not None:
                    SolverClass.credits(credits_stream)
```

unified_planning/solvers/solver.py
```diff
@@ -27,6 +30,28 @@ class OptimalityGuarantee(Enum):
...
+        if not full_credits:
+            stream.write(self.short_description)
+            stream.write('\n\n')
+        else:
+            stream.write(f'The author can be reached here: {self.contact}\n')
+            stream.write(f'This is the engine website: {self.website}\n')
+            stream.write(self.long_description)
+            stream.write('\n\n')
+
+credits = Credits('Base solver class', 'FBK-team', 'aiplan4eu@fbk.eu', 'https://github.com/aiplan4eu/unified-planning', 'Solver base class', 'Solver base class')
```
mikand, line 53: Why creating this instance here? I think it would be far better to have a staticmethod in the `Solver` class that returns a `Credits` instance (that is then printed from outside)

unified_planning/solvers/solver.py
```diff
@@ -93,7 +122,15 @@ def ground(self, problem: 'up.model.Problem') -> Tuple[Problem, Callable[[Plan],
...
-        raise NotImplementedError
+        raise
+
+    @staticmethod
+    def credits(stream: Optional[IO[str]] = sys.stdout, full_credits: bool = False):
+        '''
+        This method takes an IO[str] and writes the credits of the planner onto the stream.
+        '''
+        if stream is not None:
+            credits.write_credits(stream, full_credits)
```
mikand, line 133:
```suggestion
    @staticproperty
    def credits() -> Optional[Credits]:
        '''
        This method returns the credits for this solver, that will be printed when the solver is used.
        If this function returns None, it means no credits to print.
        '''
        return None
```

PS: you need to define static properties as follows:
```
class staticproperty(staticmethod):
    def __get__(self, *args):
        return self.__func__()
```

unified_planning/solvers/solver.py
```diff
@@ -27,6 +30,28 @@ class OptimalityGuarantee(Enum):
...
+    long_description: str
+
+    def write_credits(self, stream: IO[str], full_credits: bool = False):
+        stream.write(f'You are using {self.name}, developed by {self.author}.\n')
+        if not full_credits:
+            stream.write(self.short_description)
+            stream.write('\n\n')
+        else:
+            stream.write(f'The author can be reached here: {self.contact}\n')
+            stream.write(f'This is the engine website: {self.website}\n')
```
mikand, line 49:
```suggestion
            stream.write(f'Website: {self.website}\n')
```

unified_planning/solvers/solver.py
```diff
@@ -27,6 +30,28 @@ class OptimalityGuarantee(Enum):
...
+    short_description: str
+    long_description: str
+
+    def write_credits(self, stream: IO[str], full_credits: bool = False):
+        stream.write(f'You are using {self.name}, developed by {self.author}.\n')
+        if not full_credits:
+            stream.write(self.short_description)
+            stream.write('\n\n')
+        else:
+            stream.write(f'The author can be reached here: {self.contact}\n')
```
mikand, line 48:
```suggestion
            stream.write(f'Contacts of the authors: {self.contact}\n')
```

unified_planning/solvers/solver.py
```diff
@@ -27,6 +30,28 @@ class OptimalityGuarantee(Enum):
     SATISFICING = auto()
     SOLVED_OPTIMALLY = auto()
 
+@dataclass
+class Credits:
+    name: str
+    author: str
+    contact: str
+    website: str
```
mikand, line 38: Add a license field

mikand (changes requested):
@Framba-Luca good work. I left some comments to fix. In particular, I think that the printing shall be invoked outside the solvers that shall only (optionally) provide an instance of the `Credits` class.

mikand:
@Framba-Luca You also need to rebase and please check that you do not undo modifications done in master

unified_planning/test/test_credits.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+import sys
+from io import StringIO
+import unified_planning
+from unified_planning.shortcuts import *
+from unified_planning.test import TestCase, skipIfSolverNotAvailable
+
+
+class TestCredits(TestCase):
```
alvalentini, line 23: I would simplify these tests avoiding checking exactly a specific output, but just some general keywords, such as `developers`, `contacts`..

alvalentini (changes requested):

unified_planning/environment.py
```diff
@@ -30,11 +31,22 @@ def __init__(self):
...
         import unified_planning.walkers
         self._type_manager = unified_planning.model.TypeManager()
-        self._factory = unified_planning.solvers.Factory()
+        self._factory = unified_planning.solvers.Factory(self)
         self._tc = unified_planning.walkers.TypeChecker(self)
         self._expression_manager = unified_planning.model.ExpressionManager(self)
         self._free_vars_oracle = unified_planning.model.FreeVarsOracle()
+        self._credits_stream: Optional[IO[str]] = sys.stdout
 
+    def __getstate__(self):
```
mikand, line 40: write a couple of lines of comment to explain why this is needed, otherwise tomorrow we'll forget!

unified_planning/solvers/__init__.py
```diff
@@ -27,6 +27,6 @@
             'Parallel',
             'PDDLSolver',
             'SequentialPlanValidator',
-            'Solver', 'OptimalityGuarantee',
+            'Solver', 'OptimalityGuarantee', 'Credits',
```
mikand, line 30: At this point, we could move `Credits` in `environment.py` or in a module at top-level, no?

unified_planning/solvers/factory.py
```diff
@@ -97,20 +101,35 @@ def _get_solver(self, solver_kind: str, name: Optional[str] = None,
...
             assert isinstance(params, Dict)
             SolverClass = self._get_solver_class(solver_kind, name, problem_kind, optimality_guarantee)
+            credits = SolverClass.get_credits(**params)
+            if self.env.credits_stream is not None and credits is not None:
+                self.env.credits_stream.write('You are using ')
+                credits.write_credits(self.env.credits_stream)
             return SolverClass(**params)
 
+    @property
+    def env(self) -> 'Environment':
```
mikand, line 123: "Explicit is better than implicit"

```suggestion
    def environment(self) -> 'Environment':
```

unified_planning/solvers/grounder.py
```diff
@@ -48,36 +57,44 @@ def name(self):
...
+    @staticmethod
+    def supports(problem_kind):
+        return problem_kind <= Grounder.supported_kind()
 
     @staticmethod
     def is_grounder():
         return True
 
+    @staticmethod
+    def get_credits(**kwargs) -> Optional[solvers.solver.Credits]:
```
mikand, line 95: To reduce the noise, let's return `None` here and in other `UP-internal` things

unified_planning/solvers/plan_validator.py
```diff
@@ -238,28 +247,36 @@ def name(self):
...
+    def supports(problem_kind):
+        return problem_kind <= SequentialPlanValidator.supported_kind()
 
     @staticmethod
     def is_plan_validator():
         return True
 
+    @staticmethod
+    def get_credits(**kwargs) -> Optional[solvers.solver.Credits]:
+        return credits
```
mikand, line 279:
```suggestion
        return None
```

unified_planning/solvers/solver.py
```diff
@@ -27,6 +28,28 @@ class OptimalityGuarantee(Enum):
     SATISFICING = auto()
     SOLVED_OPTIMALLY = auto()
 
+@dataclass
+class Credits:
```
mikand, line 32: I would move this class at top level or in `environment.py`

unified_planning/solvers/solver.py
```diff
@@ -96,6 +123,14 @@ def ground(self, problem: 'up.model.AbstractProblem') -> 'up.solvers.results.Gro
         fits the solver implementation better than retrieving a Callable.'''
         raise NotImplementedError
 
+    @staticmethod
+    def get_credits(**kwargs) -> Optional[Credits]:
+        '''
+        This method returns the credits for this solver, that will be printed when the solver is used.
+        If this function returns None, it means no credits to print.
```
mikand, line 130: Document what are the `kwargs`!

unified_planning/test/pddl/enhsp.py
```diff
@@ -14,18 +14,26 @@
...
 
 FILE_PATH = os.path.dirname(os.path.abspath(__file__))
 
+credits = up.solvers.Credits('ENHSP',
+                             'Enrico Scala',
+                             'enricos83@gmail.com',
+                             'https://sites.google.com/view/enhsp/home',
+                             'GNU General Public license, version 3 or later',
+                             'Expressive Numeric Heuristic Search Planner',
+                             'ENHSP is a forward heuristic search planner, but it is expressive in that it can handle:\n - Classical Planning\n - Numeric Planning with linear and non-linear (!!) expressions\n - Planning with discretised autonomous processes and events\n - Global constraints, which are the analogous of always constraints of PDDL.'
```
mikand, line 31: @hstairs Not sure if you agree, but since this is ONLY used for unit-testing we could simply avoid the credits here and have a more "silent" testing with less clutter

unified_planning/test/test_credits.py
mikand: :+1:

unified_planning/transformers/transformer.py
```diff
@@ -16,6 +16,8 @@
 
 from unified_planning.model import AbstractProblem
 from unified_planning.plan import Plan
+from unified_planning.solvers import Credits
```
mikand, line 19: This is why I think moving credits is a good idea ;-)

mikand (changes requested):
@Framba-Luca Good work. I left some minor comments and I'll slightly change the text to be printed myself.

mikand (approved):
LGTM

alvalentini (approved):

PR #148, opened 2022-05-12
Added the missing python notebook that shows the use of transformers

Framba-Luca:


mikand:
@Framba-Luca The idea of the notebook is nice but there are a couple of isssues:
- The example is way too complicated for a demo: you need to reduce the code and complexity and add more explanation and insight in the textual part. (Imagine being someone trying this out: there is a lot of unexplained code that is non-obvious)
- Use some high-level python functions to simplify the modeling (e.g. you can create a se of python tuples/classes to model the geographical information and then you can create the objects of the problem for these python objects, similarly to the ICAPS demo with the networkx graph)
- Use some operation modes to show that a planner can/cannot solve a problem (also for your debug)

alvalentini:
@Framba-Luca I have a few comment on this notebook:
- you have to update the README in the notebooks folder;
- you have to install pyperplan, since you use it;
- I would skip the explanation on how the compilers are implemented, since we want to show how to use this operation mode independently from the actual implementation;
- super minor: in a few places you can indent the code a bit better.

alvalentini (approved):

PR #152, opened 2022-05-19
Partial order plan

Framba-Luca:


unified_planning/solvers/parallel.py
```diff
@@ -129,16 +129,16 @@ def solve(self, problem: 'up.model.Problem',
...
                 objects[obj.name] = obj
         em = problem.env.expression_manager
         actions: List[ActionInstance] = []
-        if isinstance(plan, up.plan.SequentialPlan):
+        if isinstance(plan, up.plans.SequentialPlan):
             actions = plan.actions
-        elif isinstance(plan, up.plan.TimeTriggeredPlan):
+        elif isinstance(plan, up.plans.TimeTriggeredPlan):
             actions = [a for _, a, _ in plan.actions]
         else:
```
alvalentini, line 143: Here you should handle the new plan type.

unified_planning/solvers/parallel.py
```diff
@@ -159,10 +159,10 @@ def convert_plan(self, plan: 'up.plan.Plan', problem: 'up.model.Problem')-> 'up.
...
             new_actions.append(ActionInstance(new_a, tuple(params)))
-        if isinstance(plan, up.plan.SequentialPlan):
-            return up.plan.SequentialPlan(new_actions)
-        elif isinstance(plan, up.plan.TimeTriggeredPlan):
-            return up.plan.TimeTriggeredPlan([(t, a, d) for (t, _, d), a in zip(plan.actions, new_actions)])
+        if isinstance(plan, up.plans.SequentialPlan):
+            return up.plans.SequentialPlan(new_actions)
+        elif isinstance(plan, up.plans.TimeTriggeredPlan):
+            return up.plans.TimeTriggeredPlan([(t, a, d) for (t, _, d), a in zip(plan.actions, new_actions)])
         else:
```
alvalentini, line 166: Same here

notebooks/Planning_engine_demo.ipynb
```diff
@@ -1,320 +1 @@
...
...pp' from 'notebook' (/home/luca/.local/lib/python3.8/site-packages/notebook/__init__.py)\n","\n","Failed to run jupyter as observable with args notebook --no-browser --notebook-dir=\"/home/luca/aiplan4eu/upf/notebooks\" --config=/tmp/67f562a5-f711-4f6c-a815-4318cd138d63/jupyter_notebook_config.py --NotebookApp.iopub_data_rate_limit=10000000000.0. \n","View Jupyter <a href='command:jupyter.viewOutput'>log</a> for further details."]}],"source":["with env.factory.OneshotPlanner(name='yoloplanner') as p:\n","    result = p.solve(problem)\n","    if result.status == PlanGenerationResultStatus.SOLVED_SATISFICING:\n","        print(f'{p.name} found a valid plan!')\n","        print(f'The plan is: {result.plan}')\n","        print('\\n'.join(str(x) for x in result.plan.actions))\n","    else:\n","        print('No plan found!')"]}],"metadata":{"colab":{"collapsed_sections":[],"name":"Planning Engine Demo.ipynb","provenance":[]},"interpreter":{"hash":"31f2aee4e71d21fbe5cf8b01ff0e069b9275f58929596ceb00d14d90e3e16cd6"},"kernelspec":{"display_name":"Python 3.8.10 64-bit","language":"python","name":"python3"},"language_info":{"name":"python","version":"3.8.10"}},"nbformat":4,"nbformat_minor":0}
```
alvalentini, line 1: Try to keep the same formatting, in order to understand the real changes.

Execution output should not be part of the committed notebook.

alvalentini (changes requested):

unified_planning/plans/partial_order_plan.py
```diff
@@ -0,0 +1,70 @@
...
+    '''Represents a partial order plan. Actrions are represent as an adjagency list graph.'''
+    def __init__(self, actions: Dict['plans.plan.ActionInstance', List['plans.plan.ActionInstance']]):
+        self._graph = nx.convert.from_dict_of_lists(actions, create_using=nx.DiGraph)
+
+    def __repr__(self) -> str:
+        return str(self._graph )
+
+    def __eq__(self, oth: object) -> bool:
+        if isinstance(oth, PartialOrderPlan):
+            adj_dict = self.actions
```
mikand, line 34: Modulo the comment below this looks OK. But a much simple alternative would be to use `is_isomorphic` [1] (if you specify a `node_match`, in our case the complexity should be O(V^2))

[1] https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.isomorphism.is_isomorphic.html

unified_planning/plans/partial_order_plan.py
```diff
@@ -0,0 +1,70 @@
...
+                # iterate from edges starting from this node and check that they are (semantically) the same
+                for ai_successor, oth_ai_successor in zip(ai_edges, oth_ai_edges):
+                    if ai_successor.action != oth_ai_successor.action or ai_successor.actual_parameters != oth_ai_successor.actual_parameters:
+                        return False
+            return True
+        else:
+            return False
+
+    @property
+    def actions(self) -> Dict['plans.plan.ActionInstance', List['plans.plan.ActionInstance']]:
```
mikand, line 53: This is a strange name (i was expecting a set of actions). `get_adjacency_list(self)`?
In either case, provide some documentation

unified_planning/plans/partial_order_plan.py
```diff
@@ -0,0 +1,70 @@
...
+        '''Returns the graph of action instances as an adjagency list.'''
+        return nx.convert.to_dict_of_lists(self._graph)
+
+    def replace_action_instances(self, replace_function: Callable[['plans.plan.ActionInstance'], 'plans.plan.ActionInstance']) -> 'plans.plan.Plan':
+        old_adj_list = nx.convert.to_dict_of_lists(self._graph)
```
mikand, line 58: You do not really need to construct this data structure, you can directly iterate over the `nx.DigGraph` object! An even better solution could be to provide a "private construction mechanism" that simply copies `self._graph` since you are in the same class... (see https://networkx.org/documentation/stable/reference/classes/generated/networkx.Graph.copy.html)

unified_planning/plans/partial_order_plan.py
```diff
@@ -0,0 +1,70 @@
...
+        new_adj_list: Dict['plans.plan.ActionInstance', List['plans.plan.ActionInstance']] = {}
+        # Populate the new adjacency list with the replaced action instances
+        for node, successor_list in old_adj_list:
+            new_adj_list[replace_function(node)] = [replace_function(successor) for successor in successor_list]
+        return PartialOrderPlan(new_adj_list)
+
+    def to_sequential_plan(self) -> SequentialPlan:
+        return SequentialPlan(list(nx.topological_sort(self._graph)))
+
+    def all_sequential_plans(self) -> Iterator[SequentialPlan]:
```
mikand, line 68: document!

unified_planning/plans/plan.py
```diff
@@ -0,0 +1,90 @@
...
+from unified_planning.walkers import Substituter, Simplifier
+from typing import Callable, Dict, Optional, Set, Tuple, List
+from fractions import Fraction
+
+
+'''This module defines the different plan classes.'''
+
+
+class ActionInstance:
+    '''Represents an action instance with the actual parameters.'''
```
mikand, line 29:
```suggestion
    '''Represents an action instance with the actual parameters.
    NOTE: two action instances of the same action with the same parameters are considered different as it is possible to have the same action twice in a plan.
    '''
```

unified_planning/plans/plan.py
```diff
@@ -0,0 +1,90 @@
...
+    def action(self) -> 'Action':
+        '''Returns the action.'''
+        return self._action
+
+    @property
+    def actual_parameters(self) -> Tuple['FNode', ...]:
+        '''Returns the actual parameters.'''
+        return self._params
+
+def ground_action_instance(action_instance: ActionInstance, substituter: 'Substituter', simplifier: 'Simplifier') -> 'Action':
```
mikand, line 58: document!

unified_planning/plans/plan.py
```diff
@@ -0,0 +1,90 @@
...
+    old_action = action_instance.action
+    if isinstance(old_action, InstantaneousAction):
+        assert len(old_action.parameters) == len(action_instance.actual_parameters)
+        if len(old_action.parameters) == 0:
+            return old_action.clone()
+        new_action = InstantaneousAction(_name=old_action.name, _env=old_action._env)
+        assignments: Dict[Expression, Expression] = {param : value for param, value in zip(old_action.parameters, action_instance.actual_parameters)}
```
mikand, line 65:
```suggestion
        assignments: Dict[Expression, Expression] = dict(zip(old_action.parameters, action_instance.actual_parameters))
```

unified_planning/plans/sequential_plan.py
```diff
@@ -0,0 +1,104 @@
...
+
+    @property
+    def actions(self) -> List['plans.plan.ActionInstance']:
+        '''Returns the sequence of action instances.'''
+        return self._actions
+
+    def replace_action_instances(self, replace_function: Callable[['plans.plan.ActionInstance'], 'plans.plan.ActionInstance']) -> 'up.plans.plan.Plan':
+        return SequentialPlan([replace_function(ai) for ai in self._actions])
+
+    def to_partial_order_plan(self, _env: Optional['up.Environment'] = None) -> 'up.plans.partial_order_plan.PartialOrderPlan':
```
mikand, line 50: document!

unified_planning/plans/sequential_plan.py
```diff
@@ -0,0 +1,104 @@
...
+        env = up.environment.get_env(_env)
```
mikand, line 51: I think that any kind of plan shall contain an environment. Can we put it inside the `Plan` object?

@alvalentini what do you think?

unified_planning/plans/sequential_plan.py
```diff
@@ -0,0 +1,104 @@
...
+                last_modifier[eff.fluent] = action_instance
+                dependent_action_instance_list = all_required.get(eff.fluent, None)
+                if dependent_action_instance_list is not None:
+                    for dependent_action_instance in dependent_action_instance_list:
+                        if dependent_action_instance != action_instance:
+                            graph.add_edge(dependent_action_instance, action_instance)
+
+        # remove redundant edges and return the up.plans.partial_order_plan.PartialOrderPlan structure.
+        assert nx.is_directed_acyclic_graph(graph)
+        return up.plans.partial_order_plan.PartialOrderPlan(nx.convert.to_dict_of_lists(nx.transitive_reduction(graph)))
```
mikand, line 104: Make yourself a favor and create a constructor that accepts a `nx.DiGraph` for internal use

unified_planning/plans/sequential_plan.py
```diff
@@ -0,0 +1,104 @@
...
+                assert eff.fluent.is_fluent_exp()
+                last_modifier[eff.fluent] = action_instance
+                dependent_action_instance_list = all_required.get(eff.fluent, None)
+                if dependent_action_instance_list is not None:
+                    for dependent_action_instance in dependent_action_instance_list:
+                        if dependent_action_instance != action_instance:
+                            graph.add_edge(dependent_action_instance, action_instance)
+
+        # remove redundant edges and return the up.plans.partial_order_plan.PartialOrderPlan structure.
+        assert nx.is_directed_acyclic_graph(graph)
```
mikand, line 103: This does not remove the redundant edges, you need to use https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.dag.transitive_reduction.html

unified_planning/plans/sequential_plan.py
```diff
@@ -0,0 +1,104 @@
...
+            graph.add_node(action_instance)
+
+            # required_fluents contains all the fluents that this action_instance "reads"
+            required_fluents: Set[FNode] = set()
+            # add free vars of preconditions
+            for prec in grounded_action.preconditions:
+                required_fluents = required_fluents.union(fve.get(prec))
+            # add in the required_fluents all the free fluents this action instance deals with
+            for eff in grounded_action.effects:
+                required_fluents = required_fluents.union(fve.get(eff.condition))
```
mikand, line 74: Same as above, union creates new sets each time!

unified_planning/plans/sequential_plan.py
```diff
@@ -0,0 +1,104 @@
...
+                required_fluents = required_fluents.union(fve.get(eff.fluent))
+                required_fluents = required_fluents.union(fve.get(eff.value))
+
+            # for every required fluent, add this action instance to the list of action instances that requires this fluent
+            # and order the current action instance after the last modifier of the fluent
+            for required_fluent in required_fluents:
+                action_instance_list = all_required.get(required_fluent, None)
```
mikand, line 81: You can be smart and do 
```suggestion
                action_instance_list = all_required.setdefault(required_fluent, [])
```
so that you do not need the switch below

unified_planning/test/test_negative_conditions_remover.py
```diff
@@ -99,12 +99,16 @@ def test_matchcellar(self):
         mend_f2 = ActionInstance(mend_fuse, (ObjectExp(f2), ))
         mend_f3 = ActionInstance(mend_fuse, (ObjectExp(f3), ))
         npa = [a for s, a, d in new_plan.actions]
-        self.assertIn(light_m1, npa)
-        self.assertIn(light_m2, npa)
-        self.assertIn(light_m3, npa)
-        self.assertIn(mend_f1, npa)
-        self.assertIn(mend_f2, npa)
-        self.assertIn(mend_f3, npa)
```
mikand, line 107: Maybe (just maybe) we can allow for the `in` operator to work with plans? Something like `action_instance in plan` where `plan` is any kind of plan (in Python this is done by overriding the `__contains__` method)

unified_planning/model/problem.py
```diff
@@ -14,6 +14,8 @@
 #
 '''This module defines the problem class.'''
 
+from functools import partial
+from typing_extensions import Self
```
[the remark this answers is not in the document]
Framba-Luca: Yes, it is an automatic import of visual studio code. Didn't notice it went through.

unified_planning/plans/plan.py
Framba-Luca: Actually, I was thinking about moving the logic of an Action grounding into the Action class itself. This might be also used by the Grounder, what do you think about this?

unified_planning/plans/sequential_plan.py
alvalentini: Yes, I think it make sense. Now the environment is part of the actions, but I agree that we should put it also inside the `Plan`. Do you think that then we should add a check that the environments are the same?

unified_planning/test/test_negative_conditions_remover.py
alvalentini: It can be a nice feature, even if not used much. Anyway, I do not see problems in adding it.

unified_planning/plans/partial_order_plan.py
```diff
@@ -0,0 +1,70 @@
...
+            # check number of nodes
+            if len(adj_dict.keys()) != len(oth_adj_dict.keys()):
+                return False
+            # iterate through both dicts
+            for (ai, ai_edges), (oth_ai, oth_ai_edges) in zip(adj_dict.items(), oth_adj_dict.items()):
+                # check that the nodes are (semantically) the same and that the edges starting from the current node are the same in number
+                if ai.action != oth_ai.action or ai.actual_parameters != oth_ai.actual_parameters or len(ai_edges) != len(oth_ai_edges):
+                    return False
+                # iterate from edges starting from this node and check that they are (semantically) the same
+                for ai_successor, oth_ai_successor in zip(ai_edges, oth_ai_edges):
```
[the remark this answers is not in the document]
Framba-Luca: Yes, the ordering is supposed to be equal in this implementation, which is not exactly what we want. The problem with the edges_equal is that 2 ActionInstance are never equal (unless they are exactly the same object), right?

unified_planning/plans/sequential_plan.py
```diff
@@ -0,0 +1,147 @@
...
+                lifted_required_fluents |= fve.get(eff.condition)
+                lifted_required_fluents |= fve.get(eff.fluent)
+                lifted_required_fluents |= fve.get(eff.value)
+
+            assignments: Dict[Expression, Expression] = dict(zip(inst_action.parameters, action_instance.actual_parameters))
+            for lifted_fluent in lifted_required_fluents:
+                assert lifted_fluent.is_fluent_exp()
+                for arg in lifted_fluent.args: # check that we don't have "nested" fluents
+                    if len(fve.get(arg)) != 0:
+                        raise UPUsageError(f'The partial deordering of a Sequential Plan does not allow the use of fluents inside the parameter of fluents!\nThe fluent: {lifted_fluent} does violates this contraint.')
```
alvalentini, line 121: Here we do not handle fluents that are inside an expression with quantifiers

alvalentini (changes requested):

alvalentini (changes requested):

PR #164, opened 2022-06-15
Conflicting effects check

Framba-Luca:


unified_planning/model/action.py
```diff
@@ -485,11 +486,9 @@ def add_decrease_effect(self, timing: 'up.model.timing.Timing', fluent: Union['u
                                          condition_exp, kind = up.model.effect.EffectKind.DECREASE))
 
     def _add_effect_instance(self, timing: 'up.model.timing.Timing', effect: 'up.model.effect.Effect'):
-        if timing in self._effects:
-            if effect not in self._effects[timing]:
-                self._effects[timing].append(effect)
-        else:
-            self._effects[timing] = [effect]
+        concurrent_effects = self._effects.setdefault(timing, [])
```
mikand, line 489: Why do you need to store a list of effects? A **set** of fluent expressions associated to a timing should be enough (and way more efficient!)

unified_planning/model/action.py
```diff
@@ -226,7 +226,7 @@ def add_decrease_effect(self, fluent: Union['up.model.fnode.FNode', 'up.model.fl
         self._add_effect_instance(up.model.effect.Effect(fluent_exp, value_exp, condition_exp, kind = up.model.effect.EffectKind.DECREASE))
 
     def _add_effect_instance(self, effect: 'up.model.effect.Effect'):
-        if effect not in self._effects:
+        if _check_conflicting_or_existing_effects(effect, self._simulated_effect, self._effects):
```
mikand, line 229: here you reuse `self._effects`, saving memory but the cost is that each effect addition becomes a linear-complexity operation....

mikand, line 229: Careful here, it is possible (even though very strage) to have multiple increase/decrease effects

mikand (changes requested):
@Framba-Luca Generally this is OK, but we need to better discuss the semantics of the concurrent effects

unified_planning/engines/sequential_simulator.py
```diff
@@ -0,0 +1,211 @@
...
+        '''
+        if not self.is_applicable(event, state):
+            return None
+        else:
+            return self.apply_unsafe(event, state)
+
+    def _apply_unsafe(self, event: 'Event', state: 'up.model.RWState') -> 'up.model.RWState':
+        '''
+        Returns a new RWState, which is a copy of the given state but the applicable effects of the event are applied; therefore
+        some fluent values are updated.
```
alvalentini, line 117: You are modifying the same state, not a copy of it.
I think you can implement this method using only `ROState`.

unified_planning/model/state.py
```diff
@@ -13,12 +13,108 @@
...
+        '''
+        Returns a different UPRWState in which every value in new_values.keys() is evaluated as his mapping
+        in new the new_values dict and every other value is evaluated as in self.
+        :param new_values: The dictionary that contains the values that need to be updated in the new State.
+        :return: The new State created.
+        '''
+        raise NotImplementedError
+
+
+class UPRWState(RWState):
```
alvalentini, line 47:
```suggestion
class UPRWState(RWState, ROState):
```
Right?

alvalentini (changes requested):

unified_planning/engines/sequential_simulator.py
mikand: but the `set_values` method is only defined for RWStates (It does return a new state "copied" from `state`)

unified_planning/model/state.py
mikand: Why? `RWState` inherits from `ROState`...

alvalentini: You are right, I missed it

unified_planning/engines/sequential_simulator.py
alvalentini: Where is the state copied?

In any case I think that the input state could be a ROState, since we want only to read from it.

For the returned state, with the current implementation we need it, but in principle we could set the updated values in the constructor of a new ROState.

mikand: Have a look at the `set_values` method (the name is indeed misleading)

alvalentini: Ok, now I got it!

I think both `RWState` and `set_values` are misleading names

unified_planning/engines/mixins/simulator.py
```diff
@@ -0,0 +1,131 @@
...
+                   action: 'up.model.Action',
+                   parameters: Union[Tuple['up.model.Expression', ...], List['up.model.Expression']]) -> List['Event']:
+        '''
+        Returns a list containing all the events derived from the given action, grounded with the given parameters.
+        :param action: the action containing the information to create the event.
+        :param parameters: the parameters needed to ground the action
+        :return: the List of Events derived from this action with these parameters.
+        '''
+        if action not in cast(up.model.Problem, self._problem).actions:
+            raise UPUsageError('The action given as parameter does not belong to the problem given to the SimulatorMixin.')
```
alvalentini, line 119: This should be moved again in the sequential simulator since we cannot assume that self._problem is a `Problem`

mikand (approved):
We reviewed this with @alvalentini. Please do the renaming and the simple moving of checks and then we can merge

alvalentini (changes requested):

alvalentini (approved):

PR #173, opened 2022-06-20
Added lpg engine

serivan:
Added lpg to engines

### DIFF
--- a/unified_planning/engines/factory.py
+++ b/unified_planning/engines/factory.py
@@ -32,6 +32,7 @@ DEFAULT_ENGINES = {
     'pyperplan' : ('up_pyperplan.engine', 'EngineImpl'),
     'enhsp' : ('up_enhsp.enhsp_planner', 'ENHSPSatEngine'),
     'enhsp-opt' : ('up_enhsp.enhsp_planner', 'ENHSPOptEngine'),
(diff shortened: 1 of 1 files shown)


stoppini:
hi @alvalentini, i have update the old pull request merging with the main.

unified_planning/engines/factory.py
```diff
@@ -45,6 +45,7 @@
     "enhsp": ("up_enhsp.enhsp_planner", "ENHSPSatEngine"),
     "enhsp-opt": ("up_enhsp.enhsp_planner", "ENHSPOptEngine"),
     "tamer": ("up_tamer.engine", "EngineImpl"),
+    "lpg": ("up_lpg.lpg_planner", "LPGEngine"),
```
alvalentini, line 48: `lpg` should be added also in the `DEFAULT_ENGINES_PREFERENCE_LIST`

alvalentini (changes requested):

unified_planning/engines/factory.py
alvalentini: @stoppini this comment is not solved!

alvalentini: Ok, now it is solved!

PR #174, opened 2022-06-21
Improve planner discovery and selection

### LINKED ISSUE #160: Improve planner discovery and selection from factory (config file for user)

alvalentini:
Closes #160

### DIFF
--- a/unified_planning/engines/factory.py
+++ b/unified_planning/engines/factory.py
@@ -88,55 +88,6 @@ def get_possible_config_locations() -> List[str]:
     return files
 
 
-def configure_factory(factory: 'Factory'):
-    """
-    Reads a configuration file and configures the factory.
-
-    The following is an example of configuration file:
-
-
-    [global]
-    engine_preference_list: fast-downward fast-downward-opt enhsp enhsp-opt tamer
-
-    [engine <engine-name>]
-    module_name: <module-name>
-    class_name: <class-name>
-
-
-    The configuration is read from the first `up.ini` or `.up.ini` file located in any of the
-    parent directories from which this code was called or, otherwise, from one
-    of the following files: `~/up.ini`, `~/.up.ini`, `~/.uprc`.
-    """
-    config = configparser.ConfigParser()
-    files = get_possible_config_locations()
-    print(files)
-    config.read(files)
-
-    new_engine_sections = [s for s in config.sections()
-                           if s.lower().startswith("engine ")]
-
-    for s in new_engine_sections:
-        name = s[len("engine "):]
-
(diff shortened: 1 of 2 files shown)


unified_planning/engines/factory.py
```diff
@@ -57,35 +69,126 @@ def format_table(header: List[str], rows: List[List[str]]) -> str:
...
 
+def get_possible_config_locations() -> List[str]:
+    """Returns all the possible location of the configuration file."""
+    home = os.path.expanduser('~')
+    files = []
+    stack = inspect.stack()
+    p = os.path.dirname(os.path.abspath(stack[-1].filename))
+    while os.path.join(p, 'up.ini') not in files:
+        files.append(os.path.join(p, 'up.ini'))
+        p = os.path.abspath(os.path.join(p, os.pardir))
```
mikand, line 80: OK, but there is a simpler way: https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.parents

unified_planning/engines/factory.py
```diff
@@ -57,35 +69,126 @@ def format_table(header: List[str], rows: List[List[str]]) -> str:
...
+                                        "of '%s' engine" % name)
+
+        factory.add_engine(name, module_name, class_name)
+
+
+    if "global" in config.sections():
+        pref_list = config.get("global", "engine_preference_list")
+
+        if pref_list is not None:
+            prefs = pref_list.split()
```
mikand, line 130: `split` can be dangerous if, for example I use two spaces in a row. my suggestion is to `trim` the resulting strings and skip empty strings

```suggestion
            prefs = [x.trim() for x in pref_list.split() if len(x.trim()) > 0] #or something like this
```

unified_planning/test/test_factory.py
```diff
@@ -0,0 +1,44 @@
...
+    def tearDown(self):
+        TestCase.tearDown(self)
+        if os.path.isfile(self.fname):
+            os.remove(self.fname)
+        if os.path.isfile(f'{self.fname}.temp'):
+            os.rename(f'{self.fname}.temp', self.fname)
+
+    def test_config_file(self):
+        self.assertTrue('pyperplan' in get_env().factory.preference_list)
+        with open(self.fname, 'w') as conf:
```
mikand, line 39: I think this test might fail if someone has a `.up.ini` in the directory where the `__main__` is, no?
Maybe you can make a simpler test by calling the `configure_factory` manually and allowing to specify a custom path?

mikand (approved):
@alvalentini Looks good to me apart for the test. Fix it and then go ahead and merge.

PR #176, opened 2022-06-21
Added name mangling to the PDDL writer

Framba-Luca:


unified_planning/io/pddl_writer.py
```diff
@@ -45,17 +56,17 @@ def convert(self, expression):
 
     def walk_exists(self, expression, args):
         assert len(args) == 1
-        vars_string_list = [f"?{v.name} - {str(v.type)}" for v in expression.variables()]
+        vars_string_list = [f"?{_get_pddl_name(v)} - {self.get_mangled_name(v.type)}" for v in expression.variables()]
```
mikand, line 59: why sometimes you use `_get_pddl_name` and sometimes `self.get_mangled_name`? I think this could cause some name clashes, no?

unified_planning/io/pddl_writer.py
```diff
@@ -192,23 +204,25 @@ def _write_domain(self, out: IO[str]):
                 out.write(' :action-costs')
             out.write(')\n')
 
-
         if problem_kind.has_hierarchical_typing(): # type: ignore
-            while self.problem.has_type(self.object_freshname):
-                self.object_freshname = self.object_freshname + '_'
+            object_freshname = 'object'
```
mikand, line 208: Why this code is not part of the mangling?

mikand (changes requested):
@Framba-Luca I left some comments to consider.

unified_planning/engines/pddl_planner.py
```diff
@@ -69,19 +69,19 @@ def _get_cmd(self, domain_filename: str, problem_filename: str, plan_filename: s
...
+        '''Takes a problem, a filename and a map of renamings and returns the plan parsed from the file.'''
         actions = []
         with open(plan_filename) as plan:
             for line in plan.readlines():
                 if re.match(r'^\s*(;.*)?$', line):
                     continue
                 res = re.match(r'^\s*\(\s*([\w?-]+)((\s+[\w?-]+)*)\s*\)\s*$', line)
                 if res:
-                    action = problem.action(res.group(1))
+                    action = problem.action(renamings[res.group(1)])
```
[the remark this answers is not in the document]
Framba-Luca: For construction of the `renamings` map this should never happen, so a `keyError` would be a "nice" way to find a possible mistake in the code. We could add a default to just `res.group(1)` (The first draft was like that), but then i decided that this shouldn't happen.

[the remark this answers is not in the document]
mikand: I was not suggesting to silence the error, but to raise a meaninful error instead of a `KeyError` :wink: 

("Errors should never pass silently" https://peps.python.org/pep-0020/)

mikand:
@Framba-Luca I forgot to mention: you also need to rebase :upside_down_face:

unified_planning/test/test_pddl_io.py
```diff
@@ -305,44 +305,26 @@ def test_examples_io(self):
                 w.write_problem(problem_filename)
 
                 reader = PDDLReader()
+                if problem.name != 'hierarchical_blocks_world_with_object':
```
mikand, line 308: What is this? I think it's better to look in the ProblemKind to skip the non-action-based tests, no?

unified_planning/test/test_pddl_io.py
```diff
@@ -305,44 +305,26 @@ def test_examples_io(self):
...
+                    continue
+                w.print_domain()
```
mikand, line 310: Leftover debug prints?

mikand (approved):
@Framba-Luca This looks good! There is a test to fix and you need to rebase before merging.

mikand:
@roeger We tried to fix the issues we could see in the PDDL printer. This should also improve the `PDDLEngine`(s). Can you try with fast-downward to see if any issue is remaining?

@Framba-Luca can you also try to make some example PDDL files where the master version fails with up-fast-downward and check that this branch fixes them? (no need to put them in the CI, but maybe attachthe files here)

unified_planning/test/test_pddl_io.py
mikand: @Framba-Luca you re-added them :laughing:

unified_planning/io/pddl_writer.py
```diff
@@ -16,28 +16,65 @@
...
+                 'universal-preconditions', 'quantified-preconditions', 'conditional-effects', 'fluents',
+                 'adl', 'durative-actions', 'derived-predicates', 'timed-initial-literals', 'preferences'
+                 }
+
+# The following map is used to mangle the invalid names by their class.
+INITIAL_LETTER: Dict[type, str] = {InstantaneousAction: 'a', DurativeAction: 'a', Fluent: 'f', Parameter: 'p', Problem: 'p', Object: 'o'}
+
+
+class ObjectsExtractor(walkers.DagWalker):
+    '''Returns the object in the expression.'''
```
mikand, line 53:
```suggestion
    '''Returns the object instances appearing in the expression.'''
```

unified_planning/io/pddl_writer.py
```diff
@@ -16,28 +16,65 @@
...
+        walkers.dag.DagWalker.__init__(self)
+
+    def get(self, expression: 'up.model.FNode') -> Set['up.model.Object']:
+        """Returns all the free vars of the given expression."""
+        return self.walk(expression)
+
+    @walkers.handles(up.model.OperatorKind)
+    def walk_all_types(self, expression: 'up.model.FNode', args: List[Set['up.model.Object']]) -> Set['up.model.Object']:
+        res = set(x for y in args for x in y)
+        if expression.is_object_exp():
```
mikand, line 65: Why this if instead of overriding the `walk_object_exp`?

unified_planning/io/pddl_writer.py
```diff
@@ -145,72 +182,101 @@ class PDDLWriter:
...
             out.write(' )\n')
         else:
-            out.write(f' (:types {" ".join([cast(UT, t).name for t in self.problem.user_types])})\n' if len(self.problem.user_types) > 0 else '')
+            out.write(f' (:types {" ".join([self._get_mangled_name(t) for t in self.problem.user_types])})\n' if len(self.problem.user_types) > 0 else '')
+
+        # Iterate the actions to retrieve domain objects
+        for a in self.problem.actions:
+            if isinstance(a, up.model.InstantaneousAction):
+                for p in a.preconditions:
+                    self.domain_objects = self.domain_objects | obe.get(p)
```
mikand, line 255: this way of adding to a set is very costly because you create a new set, you can simply add to the set in-place with `|=`
```suggestion
                    self.domain_objects |= obe.get(p)
```

unified_planning/io/pddl_writer.py
```diff
@@ -145,72 +182,101 @@ class PDDLWriter:
...
+                        self.domain_objects = self.domain_objects | obe.get(c)
+                for t, el in a.effects.items():
+                    for e in el:
+                        if e.is_conditional():
+                            self.domain_objects = self.domain_objects | obe.get(e.condition)
+                        self.domain_objects = self.domain_objects | obe.get(e.fluent)
+                        self.domain_objects = self.domain_objects | obe.get(e.value)
+        if len(self.domain_objects) > 0:
+            out.write(' (:constants')
+            for ut in self.problem.user_types:
```
mikand, line 275: You could have avoided this `for` loop is `self.domain_objects` were a `Dict[_UserType, Set[Object]]`...

mikand (changes requested):
@Framba-Luca generally OK! I left some comments to improve the new code for constants

unified_planning/io/pddl_writer.py
```diff
@@ -145,72 +190,100 @@ class PDDLWriter:
...
                     stack.extend(direct_sons)
-                    out.write(f'    {" ".join([self._type_name_or_object_freshname(t) for t in direct_sons])} - {self._type_name_or_object_freshname(current_type)}\n')
+                    out.write(f'    {" ".join([self._get_mangled_name(t) for t in direct_sons])} - {self._get_mangled_name(current_type)}\n')
             out.write(' )\n')
         else:
-            out.write(f' (:types {" ".join([cast(UT, t).name for t in self.problem.user_types])})\n' if len(self.problem.user_types) > 0 else '')
+            out.write(f' (:types {" ".join([self._get_mangled_name(t) for t in self.problem.user_types])})\n' if len(self.problem.user_types) > 0 else '')
+
+        # Iterate the actions to retrieve domain objects
+        for a in self.problem.actions:
```
mikand, line 260: Minor suggestion, this could be easily moved in a dedicated private function to clean the code and reduce cognitive complexity

unified_planning/model/timing.py
```diff
@@ -212,6 +218,20 @@ def __init__(self, lower: FNode, upper: FNode, is_left_open: bool = False, is_ri
...
+    def __eq__(self, oth: object) -> bool:
+        if isinstance(oth, DurationInterval):
+            return self._lower == oth._lower and self._upper == oth._upper and self._is_left_open == oth._is_left_open and self._is_right_open == oth._is_right_open
+        else:
+            return False
+
+    def __hash__(self) -> int:
+        res = hash(self._lower) + hash(self._upper)
+        if self._is_left_open:
+            res ^= hash('is_left_open')
```
mikand, line 230: OK, but why using a string here, you can easily hash booleans as well (booleans are numeric types in python)

unified_planning/model/timing.py
```diff
@@ -212,6 +218,20 @@ def __init__(self, lower: FNode, upper: FNode, is_left_open: bool = False, is_ri
...
+        if self._is_right_open:
+            res ^= hash('is_right_open')
```
mikand, line 232: I appreciate the thought of using XOR here, but there is a better way if you want good hashes (and we generally do not have good hashes in the UP ATM): https://stackoverflow.com/questions/29435556/how-to-combine-hash-codes-in-in-python3

```python
def __hash__(self) -> int:
    return hash((self._lower, self.upper, self._is_left_open, self._is_right_open))
```

However, this is a minor thing

mikand (approved):
@Framba-Luca Excellent work. I left a few very minor comments on the new code, then we can merge

unified_planning/io/pddl_writer.py
```diff
@@ -364,17 +441,21 @@ def _write_problem(self, out: IO[str]):
...
-            name = f'{self.problem.name}'
+            name = _get_pddl_name(self.problem)
         out.write(f'(define (problem {name}-problem)\n')
         out.write(f' (:domain {name}-domain)\n')
         if len(self.problem.user_types) > 0:
-            out.write(' (:objects ')
+            out.write(' (:objects')
             for t in self.problem.user_types:
-                objects = [o for o in self.problem.all_objects if o.type == t]
+                constants_of_this_type: Optional[Set[Object]] = self.domain_objects.get(t, None) # type: ignore
```
mikand, line 450:
```suggestion
                constants_of_this_type = self.domain_objects.get(cast(_UserType, t), None)
```

PR #177, opened 2022-06-23
Infix notation

Framba-Luca:


unified_planning/model/fnode.py
```diff
@@ -270,3 +271,71 @@ def is_le(self) -> bool:
...
+        return get_env().expression_manager.GE(self, right)
+
+    def __lt__(self, right):
+        return get_env().expression_manager.LT(self, right)
+
+    def __le__(self, right):
+        return get_env().expression_manager.LE(self, right)
+
+    def __neg__(self):
+        return get_env().expression_manager.Minus(0, self)
```
mikand, line 322: I think we need to add `__pos__`, and `__invert__` (see https://www.oreilly.com/library/view/python-in-a/0596001886/re22.html)

mikand (approved):
@Framba-Luca Excellent! I added a minor suggestion, for the rest, this looks OK.

unified_planning/model/object.py
```diff
@@ -17,14 +17,16 @@
...
 """
 
+from unified_planning.environment import Environment, get_env
 import unified_planning.model.types
 
 
 class Object:
     """Represents an object."""
-    def __init__(self, name: str, typename: 'unified_planning.model.types.Type'):
+    def __init__(self, name: str, typename: 'unified_planning.model.types.Type', env: Environment = None):
```
mikand, line 26: I think this should be `Optional`, no?

unified_planning/model/parameter.py
```diff
@@ -17,15 +17,17 @@
...
 
 
 class Parameter:
     '''Represents an action parameter or a fluent parameter.
     A parameter has a name, and a type.'''
-    def __init__(self, name: str, typename: 'up.model.types.Type'):
+    def __init__(self, name: str, typename: 'up.model.types.Type', env: Environment = None):
         self._name = name
         self._typename = typename
+        self._env = get_env(env)
```
mikand, line 30: No need o do `get_env`, the `env` parameter is non-optional here

unified_planning/model/variable.py
```diff
@@ -19,19 +19,20 @@
...
 import unified_planning
 import unified_planning.model.walkers as walkers
 import unified_planning.model.operators as op
 
 
-
 class Variable:
     """Represents a varible."""
-    def __init__(self, name: str, typename: 'unified_planning.model.types.Type'):
+    def __init__(self, name: str, typename: 'unified_planning.model.types.Type', env: Environment = None):
```
mikand, line 32: should be `Optional`

mikand (changes requested):
@Framba-Luca Excellent work! I left few very minor comments to fix. In addition I would add a test that triggers the exceptions when one tries to combine two different environments

unified_planning/model/expression.py
```diff
@@ -18,6 +18,7 @@
 are represented by the same object.
 """
 
+from pytest import param
```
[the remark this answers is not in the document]
Framba-Luca: vscode automatic import... My bad

unified_planning/model/parameter.py
```diff
@@ -24,17 +24,17 @@
 class Parameter:
     '''Represents an action parameter or a fluent parameter.
     A parameter has a name, and a type.'''
-    def __init__(self, name: str, typename: 'up.model.types.Type', env: Environment = None):
+    def __init__(self, name: str, typename: 'up.model.types.Type', env: Environment):
```
mikand, line 27: @Framba-Luca Sorry, my bad in the previous code review: can tou rever this change and leave environemnt as `Optional`?

unified_planning/model/problem.py
```diff
@@ -305,6 +305,8 @@ def explicit_initial_values(self) -> Dict['up.model.fnode.FNode', 'up.model.fnod
     def add_timed_goal(self, interval: Union['up.model.timing.Timing', 'up.model.timing.TimeInterval'],
                        goal: Union['up.model.fnode.FNode', 'up.model.fluent.Fluent', bool]):
         '''Adds a timed goal.'''
+        if not isinstance(goal, bool):
+            assert goal.environment == self._env, 'timed_goal does not have the same environment of the problem'
```
mikand, line 309:
```python
assert isinstance(goal, bool) or goal.environment == self._env, '...'
```

mikand (approved):
@Framba-Luca two very minor things before merging

PR #181, opened 2022-06-27
`walkers` package as `model` subpackage

alvalentini:


test_imports.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+set -e
+
+python3 -c "import unified_planning.engines"
```
mikand, line 5: You can make this more robust using `pkgutil`:

```python
import pkgutil

import unified_planning as up

package = up
for importer, modname, ispkg in pkgutil.iter_modules(package.__path__):
    print("Found submodule %s (is a package: %s)" % (modname, ispkg))
```

mikand (approved):
@alvalentini Excellent work! I lesft you an idea on how to make the list of packages dynamic, so that we do not need to maintain the list of packages in `test_imports.sh`

PR #182, opened 2022-06-28
Env to fnode

Framba-Luca:


unified_planning/model/expression.py
```diff
@@ -92,7 +92,7 @@ def create_node(self, node_type: OperatorKind, args: Iterable['up.model.fnode.FN
         if content in self.expressions:
             return self.expressions[content]
         else:
-            n =up.model.fnode.FNode(content, self._next_free_id)
+            n = up.model.fnode.FNode(content, self._next_free_id, self.env)
```
mikand, line 95: At this point, I would assert/check-and-except that all the `args` have the same environment and that it is `self.env`

mikand (approved):
@Framba-Luca Looks good! I would add a check (and a test) for the multiple environments and some utility methods that bypass the environment now that we have it locally

PR #183, opened 2022-06-28
Extend support to not constant actions duration

alvalentini:


unified_planning/io/pddl_reader.py
```diff
@@ -519,25 +519,27 @@ def parse_problem(self, domain_filename: str,
                 dur_act = up.model.DurativeAction(n, a_params, self._env)
                 dur = a['duration'][0]
                 if dur[0] == '=':
+                    dur.pop(0)
+                    dur.pop(0)
```
mikand, line 523:
```suggestion
                    dur = dur[2:]
```

unified_planning/io/pddl_reader.py
```diff
@@ -519,25 +519,27 @@ def parse_problem(self, domain_filename: str,
...
-                        v = Fraction(dur[j][2])
-                        if dur[j][0] == '>=':
-                            if lower is None or v > lower:
-                                lower = v
-                        elif dur[j][0] == '<=':
-                            if upper is None or v < upper:
-                                upper = v
+                        if dur[j][0] == '>=' and lower is None:
+                            dur[j].pop(0)
+                            dur[j].pop(0)
```
mikand, line 532: Same as above

unified_planning/io/pddl_reader.py
```diff
@@ -519,25 +519,27 @@ def parse_problem(self, domain_filename: str,
...
+                            lower = self._parse_exp(problem, dur_act, types_map, {}, dur[j])
+                        elif dur[j][0] == '<=' and upper is None:
+                            dur[j].pop(0)
+                            dur[j].pop(0)
```
mikand, line 536: Same as above

unified_planning/model/action.py
```diff
@@ -427,9 +427,9 @@ def set_duration_constraint(self, duration: 'up.model.timing.DurationInterval'):
         '''Sets the duration interval.'''
         lower, upper = duration.lower, duration.upper
         if not (lower.is_int_constant() or lower.is_real_constant()):
-            raise UPProblemDefinitionError('Duration bound must be constant.')
+            pass
         elif not (upper.is_int_constant() or upper.is_real_constant()):
-            raise UPProblemDefinitionError('Duration bound must be constant.')
+            pass
```
mikand, line 432: Can we remove these checks?

unified_planning/grpc/unified_planning.proto
```diff
@@ -367,39 +367,43 @@ message Problem {
...
-    DURATION_INEQUALITIES = 6;
+    CONTINUOUS_TIME = 2;
+    DISCRETE_TIME = 3;
+    INTERMEDIATE_CONDITIONS_AND_EFFECTS = 4;
+    TIMED_EFFECT = 5;
+    TIMED_GOALS = 6;
+    DURATION_INEQUALITIES = 7;
+    // EXPRESSION_DURATION
+    STATIC_FLUENTS = 8;
+    FLUENTS = 9;
```
mikand, line 380: these names can be confusing, can we call them `DUARTION_FLUENTS` or something like that?

unified_planning/model/problem_kind.py
```diff
@@ -22,6 +22,7 @@
 FEATURES = {
     'PROBLEM_CLASS' : ['ACTION_BASED', 'HIERARCHICAL'],
     'TIME' : ['CONTINUOUS_TIME', 'DISCRETE_TIME', 'INTERMEDIATE_CONDITIONS_AND_EFFECTS', 'TIMED_EFFECT', 'TIMED_GOALS', 'DURATION_INEQUALITIES'],
+    'EXPRESSION_DURATION' : ['STATIC_FLUENTS', 'FLUENTS'],
```
mikand, line 25: Shall we add `CONSTANTS` or we leve it implicit?

mikand (changes requested):
@alvalentini Good work! I left some comments before merging.

unified_planning/grpc/unified_planning.proto
```diff
@@ -367,39 +367,43 @@ message Problem {
...
+    TIMED_EFFECT = 5;
+    TIMED_GOALS = 6;
+    DURATION_INEQUALITIES = 7;
+    // EXPRESSION_DURATION
+    STATIC_FLUENTS_IN_DURATION = 8;
+    FLUENTS_IN_DURATION = 9;
     // NUMBERS
-    CONTINUOUS_NUMBERS = 7;
-    DISCRETE_NUMBERS = 8;
+    CONTINUOUS_NUMBERS = 10;
```
arbimo, line 382: The ids should not be changed as its breaks backward compatibility. They are only useful for compiling to the binary format but not visible by users.
So when adding a new field, you should just pick a new number that has not been previously and leave the other fields untouched.

alvalentini: Ok, thanks for the feedback!

PR #190, opened 2022-07-04
Oversubscription planning and meta engines

alvalentini:


unified_planning/engines/meta_engine.py
```diff
@@ -0,0 +1,62 @@
...
+                return self._supported_kind(engine_class)  # type: ignore
+
+            @staticmethod
+            def supports(problem_kind: ProblemKind) -> bool:
+                return self._supports(problem_kind, engine_class)  # type: ignore
+
+        return MetaEngineImpl
+
+
+class MetaEngine(Engine, metaclass=MetaEngineMeta):
```
mikand, line 44: I think we shall document the public methods :smile:

unified_planning/engines/oversubscription_planner.py
```diff
@@ -0,0 +1,124 @@
...
+from unified_planning.engines.engine import Engine
+from unified_planning.engines.meta_engine import MetaEngine
+from unified_planning.engines.mixins.oneshot_planner import OptimalityGuarantee
+from typing import Type, IO, Callable, Optional, Union, List, Tuple
+from itertools import chain, combinations
+from queue import PriorityQueue
+from fractions import Fraction
+
+
+def powerset(iterable):
```
mikand, line 30: This is the third copy of `powerset`, let's refactor this in a utility module

unified_planning/engines/oversubscription_planner.py
```diff
@@ -0,0 +1,124 @@
...
+        assert isinstance(problem, up.model.Problem)
+        assert isinstance(self.engine, mixins.OneshotPlannerMixin)
+        if len(problem.quality_metrics) == 0:
+            goals: List[Tuple["up.model.FNode", Union[Fraction, int]]] = []
+        else:
+            assert len(problem.quality_metrics) == 1
+            qm = problem.quality_metrics[0]
+            assert isinstance(qm, up.model.metrics.Oversubscription)
+            goals = list(qm.goals.items())
+        q: PriorityQueue = PriorityQueue()
```
mikand, line 105: why using a priority queue? You never mix push and pulls, so a list that is then sorted should be enough

unified_planning/engines/oversubscription_planner.py
```diff
@@ -0,0 +1,124 @@
...
+        for l in powerset(goals):
+            cost: Union[Fraction, int] = 0
+            sg = []
+            for g, c in l:
+                cost += c
+                sg.append(g)
+            q.put((-1 * cost, sg))
```
mikand, line 112: Using tuples (cost, goal) is clever, but beware that it might happen that two goals are compared to break ties (it **should** work, but please check).

unified_planning/engines/parallel.py
```diff
@@ -174,12 +187,14 @@ def _validate(
...
-    EngineClass: type,
+    factory: "up.engines.factory.Factory",
+    engine_name: str,
     options: Dict[str, str],
     signaling_queue: Queue,
     fname: str,
-    *args
+    *args,
 ):
+    EngineClass = factory._engines[engine_name]
```
mikand, line 197: This accesses a private field of the factory. Maybe we can make a public accessor method?

unified_planning/grpc/unified_planning.proto
```diff
@@ -453,6 +463,9 @@ message Metric {
     // This is very awkward to do in this setting where the expression is detached from its scope.
     map<string, Expression> action_costs = 3;
     Expression default_action_cost = 4;
+
+    // List of goals used to define the oversubscription planning problem.
```
mikand, line 467: Add a comment saying that this empty if the kind is not oversubscription

mikand (changes requested):
@alvalentini **Outstanding** work! :champagne: 
I loved the implementation you gave of `MetaEngine`. 

I left a couple of comments to fix. In addition, it would be nice to directly create a notebook showing the use of oversubscription meta-engine. And we need to write a couple of paragraphs in the docs explaining what meta-engines are.

mikand (approved):

PR #195, opened 2022-07-08
Skip protobuf testing if protobuf not installed

Framba-Luca:


unified_planning/test/__init__.py
```diff
@@ -116,5 +117,27 @@ def wrapper(*args, **kwargs):
...
+    """Skip a test if there are no oneshot planner for the given problem kind."""
+
+    def __init__(self, module_name: str):
+        self.module_name = module_name
+
+    def __call__(self, test_fun):
+        msg = f"no module named {self.module_name} installed"
+        cond = False
+        try:
+            importlib.import_module(self.module_name)
```
mikand, line 130: A better option would be to use `find_spec` that does not import the module, but generally OK. 
See: https://docs.python.org/3/library/importlib.html#importlib.util.find_spec

unified_planning/test/test_credits.py
```diff
@@ -56,6 +56,9 @@ def test_long_env_credits(self):
         credits_printed = credits.getvalue()
         # test that every keyword occur the same number of time in the printed result
         number_of_credits_printed = credits_printed.count(credits_keywords[0])
-        for keyword in credits_keywords:
-            self.assertIn(keyword, credits_printed)
-            self.assertEqual(credits_printed.count(keyword), number_of_credits_printed)
+        if number_of_credits_printed > 0:  # If at least one credit was printed
```
mikand, line 59: Why is this fix here?

mikand (approved):
@Framba-Luca excellent work! I left a couple of minors, but this is OK. @alvalentini Go ahead and merge when you want.

unified_planning/test/test_credits.py
Framba-Luca: Because when testing with a clean python environment (on pipenv), this test was failing.

unified_planning/test/__init__.py
```diff
@@ -116,5 +118,27 @@ def wrapper(*args, **kwargs):
...
+    """Skip a test if there are no oneshot planner for the given problem kind."""
+
+    def __init__(self, module_name: str):
+        self.module_name = module_name
+
+    def __call__(self, test_fun):
+        msg = f"no module named {self.module_name} installed"
+        cond = False
+        try:
+            find_spec(self.module_name)
```
mikand, line 131: Nope. this is **not** how `find_spec` is used!

Try in an interpreter:
```python
from importlib.util import find_spec
x = find_spec('mucca')
print(x)
```

mikand (changes requested):
@Framba-Luca By switching to `find_spec` you broke this code, please fix it!

PR #202, opened 2022-07-12
Add "simple_numeric_planning" to problem_kind

Framba-Luca:


unified_planning/model/problem.py
```diff
@@ -538,53 +542,193 @@ def kind(self) -> "up.model.problem_kind.ProblemKind":
...
         seldom as possible."""
+        # Create the needed data structures
+        fluents_to_only_increase: Set["up.model.fnode.FNode"] = set()
+        fluents_to_only_decrease: Set["up.model.fnode.FNode"] = set()
+        static_fluents: Set["up.model.fluent.Fluent"] = self.get_static_fluents()
+
+        # sf_positive_negative is a map from static_fluent to a tuple of 2 booleans, where the first one
+        # says if the static values are only positive and the second one says if every static value is only negative.
+        sf_positive_or_negative: Dict["up.model.fluent.Fluent", Tuple[bool, bool]] = {}
+        for f, v in self.initial_values.items():
```
alvalentini, line 553: Here you should iterate only trough `explicit_initial_values` and `fluents_defaults`, avoiding the full fluents grounding.

unified_planning/model/walkers/linear_checker.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def __init__(self, problem: Optional["up.model.problem.Problem"] = None):
+        DagWalker.__init__(self)
+        self._static_fluents: Set["up.model.fluent.Fluent"] = (
+            problem.get_static_fluents() if problem is not None else set()
+        )
+
+    def get_fluents(
+        self, expression: "up.model.fnode.FNode"
+    ) -> Tuple[bool, Set["up.model.fnode.FNode"]]:
```
alvalentini, line 41: Is it needed to return the set of fluents or would be enough to return another boolean flag saying if it contains fluents?

alvalentini (changes requested):

alvalentini (changes requested):

unified_planning/model/problem.py
```diff
@@ -538,53 +542,193 @@ def kind(self) -> "up.model.problem_kind.ProblemKind":
...
+        linear_checker = up.model.walkers.linear_checker.LinearChecker(self)
         self._kind = up.model.problem_kind.ProblemKind()
         self._kind.set_problem_class("ACTION_BASED")
+        self._kind.set_problem_type("SIMPLE_NUMERIC_PLANNING")
+        for metric in self._metrics:
+            if isinstance(metric, up.model.metrics.MinimizeExpressionOnFinalState):
+                self._kind.set_quality_metrics("FINAL_VALUE")
+                (
+                    is_linear,
+                    fluents_to_only_increase,
```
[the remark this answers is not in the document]
Framba-Luca: You are right, so we need a "stronger" walker? That says which fluents appear with which sign? Or what do you suggest?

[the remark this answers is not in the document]
alvalentini: Yes, for example you can return two sets: `positive_fluents`, `negative_fluents`.

unified_planning/model/problem.py
Framba-Luca: Yes, but we would miss the case where the `fluents_default` don't fit the SNP kind but every fluent is actually initialized explicitly to a value that is OK. Right?

alvalentini: But I think you can do that check without grounding the fluents, right?

alvalentini (changes requested):

unified_planning/test/pddl/enhsp.py
```diff
@@ -68,6 +68,8 @@ def supported_kind() -> ProblemKind:
         supported_kind.set_problem_class("ACTION_BASED")
         supported_kind.set_numbers("DISCRETE_NUMBERS")
         supported_kind.set_numbers("CONTINUOUS_NUMBERS")
+        supported_kind.set_problem_type("SIMPLE_NUMERIC_PLANNING")
+        supported_kind.set_problem_type("GENERAL_NUMERIC_PLANNING")
```
alvalentini, line 72:
```suggestion
```

alvalentini (approved):

unified_planning/model/walkers/linear_checker.py
```diff
@@ -0,0 +1,207 @@
...
+            Tuple[bool, Set["up.model.fnode.FNode"], Set["up.model.fnode.FNode"]]
+        ],
+    ) -> Tuple[bool, Set["up.model.fnode.FNode"], Set["up.model.fnode.FNode"]]:
+        is_linear = True
+        positive_fluents: Set["up.model.fnode.FNode"] = set()
+        negative_fluents: Set["up.model.fnode.FNode"] = set()
+        # If the problem is not given or the fluent is not in the problem's static fluents,
+        # it must be added to the expression fluents
+        if expression.fluent() not in self._static_fluents:
+            positive_fluents.add(expression)
```
alvalentini, line 202:
```suggestion
        positive_fluents.add(expression)
```

unified_planning/model/walkers/linear_checker.py
```diff
@@ -0,0 +1,207 @@
...
+            self._env = get_env(env)
+
+    def get_fluents(
+        self, expression: "up.model.fnode.FNode"
+    ) -> Tuple[bool, Set["up.model.fnode.FNode"], Set["up.model.fnode.FNode"]]:
+        """
+        Returns the tuple containing a flag saying if the expression is linear or not,
+        the set of the fluent_expressions appearing with a positive sign in the expression
+        and the set of the fluent_expressions appearing with a negative sign in the expression ."""
+        return self.walk(self._env.simplifier.simplify(expression))
```
alvalentini, line 62: Here we should use a custom simplifier

alvalentini (changes requested):

PR #205, opened 2022-07-13
Improve PDDL planner

alvalentini:


roeger:
UP Fast Downward in version 0.0.4 already aligns with these changes (cf pull request #218).

PR #206, opened 2022-07-14
Multi agent up

Alee08:


mikand:
Moved development in #207

PR #209, opened 2022-07-18
Add trajectory constraint 

stoppini:
Adding the pddl3 trajectory constraint inside the up.


mikand (commented):
@stoppini @hstairs I had a quick look at the PR. Looks very good!

I left some minor comments here and there (I haven't checked the semantics of the translation, just the code). The only major point I see is the pollution of the FNodes to support the translation and the modeling of constraints.

In particular:
1. Since the trajectory constraints are non-nestable and cannot be used anywhere except for the trajectory constraints, is it worthed to model them as experssions? A simple class containing an the temporal operator, its parameters and the expression it applies to would have been enough, no? To be discussed with @hstairs 
2. I really do not like the `_monitoring_atom_predicate` field you added to `FNode`s. A dictionary in the converter would be much better IMHO, unless I am missing something.

unified_planning/model/action.py
```diff
@@ -241,11 +241,15 @@ def add_precondition(
...
         """
         Adds the given expression to `action's preconditions`.
-
         :param precondition: The expression that must be added to the `action's preconditions`.
         """
         (precondition_exp,) = self._env.expression_manager.auto_promote(precondition)
         assert self._env.type_checker.get_type(precondition_exp).is_bool_type()
+        if precondition_exp.node_type in up.model.operators.TRAJECTORY_CONSTRAINTS:
+            raise UPTypeError(
+                f"Not insert trajectory constraint {precondition_exp} in precondition."
```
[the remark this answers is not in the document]
hstairs: Note that this compilation works for classical planning only. So no durative actions would be allowed.

hstairs:
We took all comments into account, apart from that asking to decoupling trajectory constraints from FNode implementation (this has been discussed in a WP3 Meeting). Is there anything left to be done here, or are we good?:)

unified_planning/model/expression.py
```diff
@@ -318,13 +318,54 @@ def Forall(
...
+                f"Always of expression: {str(expression)} must be created with exactly one parameter."
+            )
+        expressions = tuple(self.auto_promote(expression))
+        return self.create_node(node_type=OperatorKind.ALWAYS, args=expressions)
+
+    def Sometime(self, *expression: BoolExpression) -> "up.model.fnode.FNode":
+        if len(expression) != 1:
+            raise UPExpressionDefinitionError(
+                f"Sometime of expression: {str(expression)} must be created with exactly with one parameter."
+            )
```
Framba-Luca, line 333:
```suggestion
    def Sometime(self, expression: BoolExpression) -> "up.model.fnode.FNode":
```

unified_planning/model/expression.py
```diff
@@ -318,13 +318,54 @@ def Forall(
...
+        expressions = tuple(self.auto_promote(expression))
+        return self.create_node(node_type=OperatorKind.AT_MOST_ONCE, args=expressions)
+
+    def SometimeBefore(self, *expression: BoolExpression) -> "up.model.fnode.FNode":
+        if len(expression) != 2:
+            raise UPExpressionDefinitionError(
+                f"Sometime-Before of expression: {str(expression)} must be created with exactly with 2 parameters."
+            )
+        expressions = tuple(self.auto_promote(expression))
```
Framba-Luca, line 350:
```suggestion
    def SometimeBefore(self, left: BoolExpression, right: BoolExpression) -> "up.model.fnode.FNode":
        expressions = tuple(self.auto_promote(left, right))
```
Here, `left` and `right`, should be more meaningful names, depending on the given semanthic.

unified_planning/model/expression.py
```diff
@@ -318,13 +318,54 @@ def Forall(
...
+        return self.create_node(
+            node_type=OperatorKind.SOMETIME_BEFORE, args=expressions
+        )
+
+    def SometimeAfter(self, *expression: BoolExpression) -> "up.model.fnode.FNode":
```
Framba-Luca, line 355: Same of above.

unified_planning/model/expression.py
```diff
@@ -318,13 +318,54 @@ def Forall(
             node_type=OperatorKind.FORALL, args=expressions, payload=vars
         )
 
+    def Always(self, *expression: BoolExpression) -> "up.model.fnode.FNode":
```
Framba-Luca, line 321: Here, all the added methods must be documented with the format of the other expressions constructors.

Same in the `shortcuts.py` file. (All the changes done here must be done also in the shortcuts file).

unified_planning/model/expression.py
```diff
@@ -318,13 +318,54 @@ def Forall(
...
+        expressions = tuple(self.auto_promote(expression))
+        return self.create_node(node_type=OperatorKind.SOMETIME_AFTER, args=expressions)
+
     def FluentExp(
         self, fluent: "up.model.fluent.Fluent", params: Tuple[Expression, ...] = tuple()
     ) -> "up.model.fnode.FNode":
         """
         Creates an expression for the given `fluent` and `parameters`.
         Restriction: `parameters type` must be compatible with the `Fluent` :func:`signature <unified_planning.model.Fluent.signature>`
-
```
Framba-Luca, line 327: Here this empty row is needed for the correct formatting of the documentation.

unified_planning/model/problem.py
```diff
@@ -558,20 +569,67 @@ def add_goal(
         if goal_exp != self._env.expression_manager.TRUE():
             self._goals.append(goal_exp)
 
+    def add_trajectory_constraint(self, constraint: "up.model.fnode.FNode"):
```
Framba-Luca, line 572:
```suggestion
    def add_trajectory_constraint(self, constraint: "up.model.expression.BoolExpression"):
```
Since you call the `auto_promote` method, you can accept every BoolExpression, even if it is not already an `FNode`

unified_planning/model/problem.py
```diff
@@ -558,20 +569,67 @@ def add_goal(
...
+        (constraint_exp,) = self._env.expression_manager.auto_promote(constraint)
+        new_traj_list.append(
+            simplifier.simplify(
+                self._env.expression_manager.And(
+                    self._trajectory_constraints,
+                    constraint_exp,
+                )
+            )
+        )
+        self._trajectory_constraints = new_traj_list
```
Framba-Luca, line 592: Here it is a bit ugly to create a `Simplifier` and use it for just 1 simplification. Also, this may lead to unwanted behaviors, mainly because the `Simplifier` with a problem at the constructor also simplifies static fluents. But, an Action that modifies said fluent might be added later, thus making this simplification actually wrong.

Is the simplification here really needed?
We don't simplify `goals` etc. Also, here we are missing the check that it is a `trajectory_constraint`.

If it is, I would do it this way:
(Note that you lose simplification over static fluents)
```suggestion
        new_traj_list = []
        (constraint_exp,) = self._env.expression_manager.auto_promote(constraint)
        all_constraints_simplified = self._env.expression_manager.And(
            constraint_exp, 
            *self._trajectory_constraints
        ).simplify()
        self._trajectory_constraints = all_constraints_simplified.args.copy() if all_constraints_simplified.is_and() else [all_constraints_simplified]
```

unified_planning/model/problem.py
```diff
@@ -558,20 +569,67 @@ def add_goal(
...
+
+    def remove_trajectory_constraint(self, constraint: "up.model.fnode.FNode"):
```
Framba-Luca, line 594: We don't offer the capability to remove a single goal, action etc. Is this a common use-case or we can just remove this and leave the `clear_trajectory_constraints` method?

unified_planning/model/problem.py
```diff
@@ -558,20 +569,67 @@ def add_goal(
...
 
+    def clear_trajectory_constraints(self):
+        """Removes the trajectory_constraints."""
+        self._trajectory_constraints = []
+
     def add_quality_metric(self, metric: "up.model.metrics.PlanQualityMetric"):
         """
         Adds the given `quality metric` to the `Problem`; a `quality metric` defines extra requirements that a :class:`~unified_planning.plans.Plan`
         must satisfy in order to be valid.
-
```
Framba-Luca, line 574: Empty row needed.

unified_planning/model/walkers/type_checker.py
```diff
@@ -83,6 +83,46 @@ def walk_param_exp(
         assert len(args) == 0
         return expression.parameter().type
 
+    def walk_always(
+        self, expression: FNode, args: List["unified_planning.model.types.Type"]
+    ) -> "unified_planning.model.types.Type":
+        assert expression is not None
+        assert expression.is_always()
+        assert len(expression.args) == 1
+        return BOOL
```
Framba-Luca, line 92: In all those `return BOOL`, i think you should check what is actually returned in the `args`. For construction this should be equivalent, but it's better to check that the `args` is actually a `BOOL`, because if it is something else, you should return `None`

unified_planning/shortcuts.py
```diff
@@ -145,6 +145,26 @@ def FluentExp(
     return get_env().expression_manager.FluentExp(fluent, params)
 
 
+def Always(expression: BoolExpression) -> FNode:
```
Framba-Luca, line 148: See comments in the `model.expression.py`.

unified_planning/test/test_remove_quantifier_in_traj_constraint.py
```diff
@@ -0,0 +1,126 @@
...
+from unified_planning.model.walkers import Simplifier, ExpressionQuantifiersRemover
+from unified_planning.shortcuts import *
+from unified_planning.test import TestCase
+
+
+class TestRemoveQuantifierInTrajConstraint(TestCase):
+    def setUp(self):
+        TestCase.setUp(self)
+        problem = self.define_problem()
+        self.simplifier = Simplifier(problem.env)
```
Framba-Luca, line 29: Never used.

unified_planning/test/test_remove_quantifier_in_traj_constraint.py
```diff
@@ -0,0 +1,126 @@
...
+        self.problem = problem
+        self.traj_remover = TrajectoryConstraintsRemover()
+        self.traj_remover._expression_quantifier_remover = ExpressionQuantifiersRemover(
+            problem.env
+        )
```
Framba-Luca, line 34: This should not be needed.

unified_planning/test/test_remove_quantifier_in_traj_constraint.py
```diff
@@ -0,0 +1,126 @@
...
+        self.traj_remover._problem = self.problem
```
Framba-Luca, line 35: This is not the correct way to use a `Compiler`, use the `compile` method instead

unified_planning/test/test_trajectory_constraints_remover.py
```diff
@@ -0,0 +1,130 @@
+# Copyright 2021 AIPlan4EU project
```
Framba-Luca, line 1: In this test, I was expecting a full usage of the `Compiler`, how it's interface should be used, how the `map_back_actions_instance` callable works and some testing on the general `CompilerResult` to make sure everything works as intended.

Try to check the tests on other compilers if you need some examples.

unified_planning/engines/compilers/trajectory_constraints_remover.py
```diff
@@ -0,0 +1,500 @@
...
+            return self._env.expression_manager.Or(
+                [self._regression(component, action) for component in phi.args]
+            )
+        elif phi.is_and():
+            return self._env.expression_manager.And(
+                [self._regression(component, action) for component in phi.args]
+            )
+        elif phi.is_not():
+            return self._env.expression_manager.Not(
+                self._regression(phi.args[0], action)
```
Framba-Luca, line 495:
```suggestion
                self._regression(phi.arg(0), action)
```

unified_planning/engines/compilers/trajectory_constraints_remover.py
```diff
@@ -0,0 +1,500 @@
...
+                eff = eff.fluent
+            if literal == eff:
+                if cond.is_true():
+                    return self._env.expression_manager.TRUE()
+                disjunction.append(cond)
+        if not disjunction:
+            return False
+        return self._env.expression_manager.Or(disjunction)
+
+    def _regression(self, phi, action):
```
Framba-Luca, line 480: This function could be a walker

.github/workflows/main.yml
```diff
@@ -216,7 +216,7 @@ jobs:
 
       - name: Install Protobuf compiler
         run: |
-          python3 -m pip install grpcio-tools
+          python3 -m pip install grpcio-tools==1.48.1
```
Framba-Luca, line 219: Is this change needed here?

Framba-Luca (changes requested):
Hi, nice work with the PR!
I left some comments in the code and some notes here:
-  It is common python practice to call variables with all lower case characters; CamelCase is "reserved" for classes and constructors.
- Generally, all methods (even private methods, thus method starting with `_` by python convention) in the UP should be fully typed, this really improves code readability.
- The public methods should also be documented.

If you have any questions, something unclear or we don't agree on something, just answer the comments and we can discuss about it.

Framba-Luca:
Hi @stoppini, the version of mypy changed, so the old master version does not work anymore. For the CI to pass you have to rebase from master.
In the meantime, nice work on the changes! :smiley:

unified_planning/engines/compilers/trajectory_constraints_remover.py
```diff
@@ -0,0 +1,422 @@
...
+        expression_quantifier_remover = ExpressionQuantifiersRemover(problem.env)
+        self._grounding_result = engines.compilers.grounder.Grounder().compile(
+            problem, CompilationKind.GROUNDING
+        )
+        assert isinstance(self._grounding_result.problem, Problem)
+        self._problem = self._grounding_result.problem.clone()
+        self._problem.name = f"{self.name}_{problem.name}"
+        A = self._problem.actions
+        I = self._problem.initial_values
+        C = self._build_constraint_list(
```
mikand, line 125: If you properly represent the constraints as a list (see `problem.py`), this should not be needed (you would only need to remove the quantifiers)...

unified_planning/engines/compilers/trajectory_constraints_remover.py
```diff
@@ -0,0 +1,422 @@
...
+        :param problem: The instance of the `Problem` that contains the trajecotry constraints.
+        :param compilation_kind: The `CompilationKind` that must be applied on the given problem;
+            only `TRAJECTORY_CONSTRAINTS_REMOVING` is supported by this compiler
+        :return: The resulting `CompilerResult` data structure.
+        """
+        assert isinstance(problem, Problem)
+        env = problem.env
+        substituter = Substituter(env)
+        expression_quantifier_remover = ExpressionQuantifiersRemover(problem.env)
+        self._grounding_result = engines.compilers.grounder.Grounder().compile(
```
mikand, line 117: Why does this need to be a field?

(In general, delcaring fields in methods other than `__init__` is a bad idea...)

unified_planning/engines/compilers/trajectory_constraints_remover.py
```diff
@@ -0,0 +1,422 @@
...
+            problem, CompilationKind.GROUNDING
+        )
+        assert isinstance(self._grounding_result.problem, Problem)
+        self._problem = self._grounding_result.problem.clone()
```
mikand, line 121: Why does this need to be a field?

(In general, delcaring fields in methods other than `__init__` is a bad idea...)

unified_planning/engines/compilers/trajectory_constraints_remover.py
```diff
@@ -0,0 +1,422 @@
...
+            env.expression_manager.And(
+                self._remove_quantifier(expression_quantifier_remover, C_list)
+            )
+        ).simplify()
+        return C_to_return.args if C_to_return.is_and() else [C_to_return]
+
+    def _remove_quantifier(self, expression_quantifier_remover, C):
+        new_C = []
+        for c in C:
+            assert c.node_type is not OperatorKind.EXISTS
```
mikand, line 212: Why not?

And also, what about an expression like `And(x, Exists(y, And(x, y))`?

unified_planning/engines/compilers/trajectory_constraints_remover.py
```diff
@@ -0,0 +1,422 @@
...
+            return rho, True
+
+    def _manage_always_compilation(self, env, phi, a):
+        R = (self._regression(env, phi, a)).simplify()
+        if R == phi:
+            return None, False
+        else:
+            return R, True
+
+    def _add_cond_eff(self, env, E, cond, eff):
```
mikand, line 272: I think it would be a good ide to document that you cahnge `E` or make the function returning the things to add to make it reentrant

unified_planning/engines/compilers/trajectory_constraints_remover.py
```diff
@@ -0,0 +1,422 @@
...
+                expression_quantifier_remover.remove_quantifiers(c, self._problem)
+            )
+        return new_C
+
+    def _manage_sa_compilation(self, env, phi, psi, m_atom, a, E):
+        R1 = (self._regression(env, phi, a)).simplify()
+        R2 = (self._regression(env, psi, a)).simplify()
+        if phi != R1 or psi != R2:
+            cond = (
+                env.expression_manager.And([R1, env.expression_manager.Not(R2)])
```
mikand, line 223: No need for the intermediate list:
```suggestion
                env.expression_manager.And(R1, env.expression_manager.Not(R2))
```

unified_planning/engines/compilers/trajectory_constraints_remover.py
```diff
@@ -0,0 +1,422 @@
...
+        if R != phi:
+            self._add_cond_eff(env, E, R, m_atom)
+
+    def _manage_sb_compilation(self, env, phi, psi, m_atom, a, E):
+        R_phi = (self._regression(env, phi, a)).simplify()
+        if R_phi == phi:
+            added_precond = (None, False)
+        else:
+            rho = (
+                env.expression_manager.Or([env.expression_manager.Not(R_phi), m_atom])
```
mikand, line 240:
```suggestion
                env.expression_manager.Or(env.expression_manager.Not(R_phi), m_atom)
```

unified_planning/engines/compilers/trajectory_constraints_remover.py
```diff
@@ -0,0 +1,422 @@
...
+        if R == phi:
+            return None, False
+        else:
+            rho = (
+                env.expression_manager.Or(
+                    [
+                        env.expression_manager.Not(R),
+                        env.expression_manager.Not(m_atom),
+                        phi,
+                    ]
```
mikand, line 259:
```suggestion
                env.expression_manager.Or(
                        env.expression_manager.Not(R),
                        env.expression_manager.Not(m_atom),
                        phi
```

unified_planning/engines/compilers/trajectory_constraints_remover.py
```diff
@@ -0,0 +1,422 @@
...
+            if eff.value.is_false():
+                eff = env.expression_manager.Not(eff.fluent)
+            else:
+                eff = eff.fluent
+            if literal == eff:
+                if cond.is_true():
+                    return env.expression_manager.TRUE()
+                disjunction.append(cond)
+        if not disjunction:
+            return False
```
mikand, line 401: Better:
```suggestion
            return env.expression_manager.FALSE()
```

unified_planning/engines/compilers/trajectory_constraints_remover.py
```diff
@@ -0,0 +1,422 @@
...
+        return env.expression_manager.Or(disjunction)
+
+    def _regression(self, env, phi, action):
+        if phi.is_false() or phi.is_true():
+            return phi
+        elif phi.is_fluent_exp():
+            return self._gamma_substitution(env, phi, action)
+        elif phi.is_or():
+            return env.expression_manager.Or(
+                [self._regression(env, component, action) for component in phi.args]
```
mikand, line 411:
```suggestion
                self._regression(env, component, action) for component in phi.args
```

unified_planning/engines/compilers/trajectory_constraints_remover.py
```diff
@@ -0,0 +1,422 @@
...
+            )
+        elif phi.is_and():
+            return env.expression_manager.And(
+                [self._regression(env, component, action) for component in phi.args]
```
mikand, line 415:
```suggestion
                self._regression(env, component, action) for component in phi.args
```

mikand (changes requested):
@stoppini @hstairs Exceptional job! I left some comments to fix, but this is my last review as the code looks bery good. let me know if something is not clear.

unified_planning/engines/compilers/trajectory_constraints_remover.py
```diff
@@ -0,0 +1,500 @@
...
+            for condition in c.args:
+                relevant_atoms += self._get_all_atoms(condition)
+                relevant_atoms = self._remove_duplicates(relevant_atoms)
+            for atom in relevant_atoms:
+                if atom not in relevancy_dict:
+                    relevancy_dict[atom] = []
+                relevancy_dict[atom].append(c)
+        return relevancy_dict
+
+    def _get_all_atoms(self, condition):
```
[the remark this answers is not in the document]
Framba-Luca: Can this method be deleted?

unified_planning/model/walkers/type_checker.py
```diff
@@ -83,6 +83,48 @@ def walk_param_exp(
         assert len(args) == 0
         return expression.parameter().type
 
+    def walk_always(
```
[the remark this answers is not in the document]
Framba-Luca: Hi @stoppini , I unresolved these conversation because there is an error. In general, in the type_checker, you want to return None when there is a typing error or the correct type of the expression if there are no typing errors.

Here I leave you an example of the `walk_always` for clarity:
```suggestion
    def walk_always(
        self, expression: FNode, args: List["unified_planning.model.types.Type"]
    ) -> "unified_planning.model.types.Type":
        assert expression is not None
        assert expression.is_always()
        assert len(expression.args) == 1
        if args[0] is None or not args[0].is_bool_type(): #type error
            return None
        return args[0] #correct type
```

In particular, you want to check the args given as parameters for the correctness of the types, and not the `expression.args[0].type` field. Note that the `FNode.type` field is basically another call to the `type_checker`.

This change/logic needs to be applied to all the methods of the type_checker! 

Let me know if you have any doubts/questions

mikand:
@stoppini What is the status of this? Just let m know when you are done fixing my code-review comments, so that I can check and merge.

stoppini:
> @stoppini What is the status of this? Just let m know when you are done fixing my code-review comments, so that I can check and merge.

Hi @mikand , i have fixed the code after your code-review.  When you want, can check and merge. 
I remain available for any problem

unified_planning/engines/compilers/trajectory_constraints_remover.py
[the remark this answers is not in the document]
alvalentini: This method seems to be not used

unified_planning/engines/plan_validator.py
```diff
@@ -73,6 +73,7 @@ def supported_kind() -> ProblemKind:
         supported_kind.set_conditions_kind("EQUALITY")
         supported_kind.set_conditions_kind("EXISTENTIAL_CONDITIONS")
         supported_kind.set_conditions_kind("UNIVERSAL_CONDITIONS")
+        supported_kind.set_constraints_kind("TRAJECTORY_CONSTRAINTS")
```
alvalentini, line 76: Are the trajectory constraints really supported by the plan validator?

alvalentini (changes requested):
@stoppini a few code-review comments are still unresolved. Have a look at them and fix them or reply with a comment if you disagree or the comment is not clear.

unified_planning/engines/plan_validator.py
stoppini: hi @alvalentini, I had already fixed it according to the comments made. 
Regarding plan_validator, I forgot to delete the line after trying it, and no, not support native traj_constr

unified_planning/engines/compilers/trajectory_constraints_remover.py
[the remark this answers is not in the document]
hstairs: @stoppini Is this solved?

unified_planning/engines/compilers/trajectory_constraints_remover.py
stoppini: trajectory constraints support Exists only outside.

unified_planning/io/pddl_writer.py
```diff
@@ -362,6 +382,8 @@ def _write_domain(self, out: IO[str]):
                 out.write(" :conditional-effects")
             if self.problem_kind.has_existential_conditions():
                 out.write(" :existential-preconditions")
+            if self.problem_kind.has_trajectory_constraints():
+                out.write(" :constraints")
```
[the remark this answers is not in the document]
stoppini: alot of benckmark problem ie ICAPS-2021 use this keyward inside the domain. 
I entered it to parse those problem

unified_planning/engines/compilers/trajectory_constraints_remover.py
mikand: @hstairs My understanding was that `Exists` is not supported "outside", but "inside" it should be allowed (as it amounts to a simple disjunction).

unified_planning/model/problem.py
```diff
@@ -555,15 +566,38 @@ def add_goal(
...
+    ):
+        """
+        Adds the given `trajectory_constraint` to the `Problem`;
+        a trajectory_constraint is an expression defined as:
+            Always, Sometime, At-Most-Once, Sometime-Before, Sometime-After or
+            defined with quantifiers.
+
+        :param trajectory_constraint: The expression added to the `Problem`.
+        """
+        (constraint_exp,) = self._env.expression_manager.auto_promote(constraint)
```
mikand, line 580: Here we could check that the constraint is syntactically correct (i.e. contains exactly one temporal operator at the top or a forall of temporal operators and no nesting)

unified_planning/engines/compilers/trajectory_constraints_remover.py
stoppini: @mikand yes "inside" is allowed. In my test "_remove_quantifier" remove internal exist with no error.
For example `(always(and (at a)(exists(?x)(at ?x))))` is compiled in a correct way in `Always((at(a) and (at(a) or at(b) or at(c))))`

PR #211, opened 2022-07-19
Universal effects parsing from PDDL in the UP

Framba-Luca:


unified_planning/io/pddl_reader.py
```diff
@@ -373,40 +388,47 @@ def _add_effect(
...
-                    self._parse_exp(problem, act, types_map, {}, exp[2]),
+                    self._parse_exp(problem, act, types_map, {}, exp[1], assignments),
+                    self._parse_exp(problem, act, types_map, {}, exp[2], assignments),
                     cond,
                 )
                 act.add_decrease_effect(*eff if timing is None else (timing, *eff))  # type: ignore
+            elif op == "forall":
+                assert isinstance(
+                    exp, ParseResults
+                )  # NOTE not sure if we always have a ParseResult here, but this implementation assumes it
```
Framba-Luca, line 425: @alvalentini Here, by typing it could also be a string, but I don't know how to handle it

alvalentini: Inside that if statement you are sure it is a `ParseResults`

PR #221, opened 2022-07-26
Documentation

Framba-Luca:


mikand (commented):
@Framba-Luca excellent work, this is EXTREMELY useful!

My only comment is on formatting: the sphinx comments can use RST and you should use the backticks notations when citing methods/fields/functions to create good-looking hyperlinks: https://stackoverflow.com/questions/21289806/link-to-class-method-in-python-docstring

PR #222, opened 2022-07-26
Contingent Planning

alvalentini:


alvalentini:
@shanigue any feedback about this PR?

shanigu:
Did not start working yet, sorry.

shanigu:
I tried to upload files to up-cpor and got an error, due to no permissions.
Can you please take a look?
Here is the error message:
Pushing master
Remote: Permission to aiplan4eu/up-cpor.git denied to shanigu.
Error encountered while pushing to the remote repository: Git failed with a fatal error.
Git failed with a fatal error.
unable to access 'https://github.com/aiplan4eu/up-cpor/': The requested URL returned error: 403

From: גיא שני - הנדסת מערכות מידע
Sent: Monday, August 8, 2022 12:20 PM
To: aiplan4eu/unified-planning ***@***.***>; aiplan4eu/unified-planning ***@***.***>
Cc: Mention ***@***.***>
Subject: RE: [aiplan4eu/unified-planning] Contingent Planning (PR #222)

Did not start working yet, sorry.

mikand:
I added you as a maintainer, see if it works.

shanigu:
Good. I was able to load CPOR.
Next problem: I started checking whether we can load contingent problems in PDDL format. I worked with doors5 example (attached) and it didn’t work.
I used the following code from your documentation:

# Importing the PDDLReader and PDDLWriter
from unified_planning.io import PDDLReader, PDDLWriter

# Creating a PDDL reader
reader = PDDLReader()

# Parsing a PDDL problem from file
problem = reader.parse_problem(
    "./unified_planning/test/pddl/doors5/d.pddl",
    "./unified_planning/test/pddl/doors5/p.pddl",
)

# Creating a PDDL writer
writer = PDDLWriter(problem)

Thanks
Guy

mikand:
@shanigu we still do not have a PDDL parser for these files. We'll put it in the agenda.

Can you send us a link to some benchmarks to use for testing?

shanigu:
I think that it is best to test on the benchmarks by Alex Albore: https://www.upf.edu/documents/7944939/136491029/benchmarks-CLG.zip/d85985a1-32d1-7421-6e70-1a0d282d3cd8

mikand:
@shanigu I implemented a first prototype of PDDL parser so that you are not blocked. Let me know if you encounter any issue.

**NOTE**: for the moment, you need to put a `:contingent` among the PDDL requirements, so that the parser recognizes that you want to construct a `ContingentPlanning` instance. We might want to relax this in the future.

shanigu:
Two issues:
1) There are no current accessors for the constraints and hidden fluents in ContingentProblem. I added locally for now: 

    @property
    def constraints(self) -> List["up.model.fnode.FNode"] :
        return self._initial_constraints

    @property
    def hidden(self) -> Set["up.model.fnode.FNode"]:
        return self._hidden_fluents

2) For oneof constraints, you seem to have compiled them into CNF form. While I understand that oneof is not a standard operator in logic, it is very convenient for doing reasoning. I suggest that you will allow for oneof constraints.

alvalentini:
@shanigu 

To avoid the use of CNF formula I propose to have separate fields to store `or` and `oneof` constraints.

```
@property
def or_constraints(self) -> List[List["up.model.fnode.FNode"]]:
    return self._initial_or_constraints

@property
def oneof_constraints(self) -> List[List["up.model.fnode.FNode"]]:
    return self._initial_oneof_constraints
```

Do you think this can be a reasonable solution?

shanigu:
I don’t think that this is a good suggestion, because it will be difficult to extend with additional operators if need be.
Why is it not sufficient to add another type of OperatorKind?

alvalentini:
Because it has a spread impact to all the `FNode` used in the problem (e.g. action conditions) and to all the [walkers](https://github.com/aiplan4eu/unified-planning/tree/master/unified_planning/model/walkers).

shanigu:
Yes, but this is potentially a good thing. Why not have conditional actions or action preconditions with a oneof statement? One can think of cases where this is appropriate.

mikand:
@shanigu I agree with @alvalentini here: while having oneof as a convenience operator in th library makes theoretical sense, it has a wide impact on the existing library as all the places where formulae are used could be impacted and (most importantly) very few planners will support this kind of constraint natively in e.g. action preconditions. We could revwite the operator in propositional logic, but that would be a hard-to-justify burden for planning engine integrators.

I think handling the operator inside of the `ContingentProblem` is currently the best solution for us. In the future, if we decide to have first-calss support for this operator, we can easily implement the proposed interface over a proper FNode representation.

shanigu:
Ok. This is not critical for us.
________________________________

unified_planning/plans/contingent_plan.py
```diff
@@ -0,0 +1,153 @@
...
+class ContingentPlanNode:
+    """This class represent a node in the tree contingent plan."""
+
+    def __init__(self, action_instance: "plans.plan.ActionInstance"):
+        self.action_instance = action_instance
+        self.children: List[
+            Tuple[Dict["up.model.FNode", "up.model.FNode"], "ContingentPlanNode"]
+        ] = []
+
+    def add_child(
```
Framba-Luca, line 30: This method should check that the `env` of the added node is the same of the current node.

Framba-Luca (commented):
Added some notes, this PR also needs testing!

PR #225, opened 2022-08-01
Anytime Planning

alvalentini:


alvalentini:
@arbimo @franklinselva 
In this PR we created a separated operation mode for anytime planning, instead of using the callback in the oneshot OM.
In this way we can explicitly ask for an engine that supports this feature. What do you think?

This change an impact on your engine integration.

mikand:
@alvalentini As a first approximation I suggest the folowing:

- `AnytimePlanner(problem_kind=p.kind, anytime_guarantee=INCREASING_QUALITY)` : Every plan being generated is strictly better in terms of quality than the previous one. Raises an exception if the problem has no optimality metrics.
- `AnytimePlanner(problem_kind=p.kind, anytime_guarantee=NONE)`: Just an enumeration of plans, no specific guarantee.
- `AnytimePlanner(problem_kind=p.kind, anytime_guarantee=OPTIMAL_PLANS)`: All plans being generated are optimal.

The `anytime_guarantee` parameter is compulsory for this OM.

PR #229, opened 2022-08-05
SequentialSimulator now is a MetaEngine

Framba-Luca:


mikand:
After a quick discussion with @alvalentini, we decided to close this PR as it is more complicated than anticipated. We can use the "internal" grounder assuming its behavior in the Simulator implementation.

PR #234, opened 2022-08-24
Add Aries Engine

franklinselva:


franklinselva:
@mikand Sorry for the late reply! The PR is still in the draft since unified planning requires a GRPC server based on which Aries is built. There is already a base work available at up-aries and I will submit the work upstream.

mikand (approved):
@franklinselva This is OK for me, let me know if/when to merge!

unified_planning/grpc/server.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def _solve(
+        self,
+        problem: "up.model.AbstractProblem",
+        callback: Optional[
+            Callable[["up.engines.results.PlanGenerationResult"], None]
+        ] = None,
+        timeout: Optional[float] = None,
+        output_stream: Optional[IO[str]] = None,
+    ) -> "up.engines.results.PlanGenerationResult":
```
alvalentini, line 107: The signature of this method is changed. Now we have a dedicated OM for anytime planning and we removed the callback from this method. We also added a new parameter to specify a custom heuristic.

Have a look at the OneshotPlannerMixin class for more details.

unified_planning/engines/factory.py
```diff
@@ -40,6 +40,7 @@
     "enhsp": ("up_enhsp.enhsp_planner", "ENHSPSatEngine"),
     "enhsp-opt": ("up_enhsp.enhsp_planner", "ENHSPOptEngine"),
     "tamer": ("up_tamer.engine", "EngineImpl"),
+    "aries": ("up_aries", "Aries"),
```
alvalentini, line 43: Is `up_aries` available in pypi?

And when you add an engine in the `DEFAULT_ENGINES` dict, you also need to add it in the `DEFAULT_ENGINES_PREFERENCE_LIST`.

alvalentini (changes requested):

unified_planning/engines/factory.py
franklinselva: `up_aries` is currently not available in pypi yet.

I have updated the requested changes.

alvalentini: We have the policy to add an engine as a default engine only when it is available in pypi

unified_planning/grpc/server.py
```diff
@@ -0,0 +1,165 @@
...
+        if (
+            response.status == PlanGenerationResultStatus.INTERMEDIATE
+            and heuristic is not None
+        ):
+            # TODO: Implement heuristic
+            pass
+        else:
+            return response
+
+        raise UPException("No response from the server")
```
alvalentini, line 140: The `INTERMEDIATE` result status is not a valid status for a oneshot planning call. It is used only by the anytime planning operation mode.

I would change the code as follow:
```suggestion
        if heuristic is not None:
            warn("Custom heuristic is not used by this planner!")
        return response
```

alvalentini (changes requested):

unified_planning/grpc/server.py
franklinselva: Thanks!

unified_planning/engines/factory.py
franklinselva: @alvalentini I have moved the grpc planner definition to PR #334. Once the PR is merged, I will setup PyPI for `up_aries`

franklinselva:
@alvalentini The planner is now available in [PyPi](https://pypi.org/project/up-aries/).

alvalentini:
> @alvalentini The planner is now available in [PyPi](https://pypi.org/project/up-aries/).

@franklinselva Ok, add it in the setup.py file as an extra requirement

alvalentini (approved):
@arbimo @franklinselva LGTM! Can I proceed with the merge?

arbimo:
Yes !
You're on fire today, I was waiting for CI to end before asking you ;)

alvalentini:
You are on fire with all these PRs :)

PR #237, opened 2022-08-29
Partial Grounding in SequentialSimulator and in SequentialPlanValidator 

Framba-Luca:
The `GroundingSupport` class allows the grounding of only an `Action` and not necessarily an entire `Problem`.

This class is used in the `Grounder` compiler and in the `SequentialSimulator`.

### DIFF
--- a/unified_planning/engines/compilers/grounder.py
+++ b/unified_planning/engines/compilers/grounder.py
@@ -32,18 +32,17 @@ from itertools import product
 from functools import partial
 
 
-class GroundingSupport:
+class GrounderHelper:
     """
-    TODO check
     This class gives the capability of grounding a :class:`~unified_planning.model.Problem` by taking
     it at construction time.
 
-    It offers the capability both of grounding a whole problem, like the :class:`~unified_planning.engines.compilers.Grounder`
-    in the :func:`~unified_planning.engines.compilers.Grounder.compile` method (with the :func:`~unified_planning.engines.compilers.GroundingSupport.ground_problem`
-    method) or offers the capability of grounding a single Action of the Problem given the grounding parameters
-    (with the :func:`~unified_planning.engines.compilers.GroundingSupport.ground_action` method).
+    It offers the capability both of grounding the whole `Problem`, with the :func:`~unified_planning.engines.compilers.GrounderHelper.get_grounded_actions`
+    function or offers the capability of grounding a single `Action` of the `Problem`, given the grounding parameters
+    (with the :func:`~unified_planning.engines.compilers.GrounderHelper.ground_action` function).
 
-    This class also caches the grounded actions so it avoids duplication.
+    Important NOTE: This class caches the grounded actions created to avoid duplication; 2 different calls
+    with the same parameters will return the same object!
     """
 
     def __init__(
@@ -52,7 +51,7 @@ class GroundingSupport:
         grounding_actions_map: Optional[Dict[Action, List[Tuple[FNode, ...]]]] = None,
     ):
         """
-        Creates an instance of the GroundingSupport.
+        Creates an instance of the GrounderHelper.
 
         :param problem: The `Problem` to ground.
         :param grounding_actions_map: Optionally, a map from `Action` to a List of Tuples of expressions.
@@ -87,7 +86,6 @@ class GroundingSupport:
         self, action: Action, parameters: Tuple[FNode, ...] = tuple()
     ) -> Optional[Action]:
         """
-        TODO check
         Grounds the given `action` with the given `parameters`.
         An `Action` is grounded when it has no :func:`parameters <unified_planning.model.Action.parameters>`.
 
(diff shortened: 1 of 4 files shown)


unified_planning/engines/compilers/grounder.py
```diff
@@ -32,6 +32,152 @@
...
+        self._grounded_actions: Dict[
+            Tuple[Action, Tuple[FNode, ...]], Optional[Action]
+        ] = {}
+        env = problem.env
+        self._substituter = Substituter(env)
+        self._simplifier = Simplifier(env, problem)
+        self._grounding_result: Optional[CompilerResult] = None
+
+    @property
+    def name(self):
```
mikand, line 67: Is this proprty needed?

unified_planning/engines/compilers/grounder.py
```diff
@@ -32,6 +32,152 @@
...
+    method) or offers the capability of grounding a single Action of the Problem given the grounding parameters
+    (with the :func:`~unified_planning.engines.compilers.GroundingSupport.ground_action` method).
+
+    This class also caches the grounded actions so it avoids duplication.
+    """
+
+    def __init__(
+        self,
+        problem: Problem,
+        grounding_actions_map: Optional[Dict[Action, List[Tuple[FNode, ...]]]] = None,
```
mikand, line 51: document this parameter, as it is not very obvious

unified_planning/engines/compilers/grounder.py
```diff
@@ -32,6 +32,152 @@
 from functools import partial
 
 
+class GroundingSupport:
```
mikand, line 35: Not a great name.
```suggestion
class GrounderHelper:
```

unified_planning/engines/compilers/grounder.py
```diff
@@ -32,6 +32,152 @@
...
+            subs: Dict[Expression, Expression] = dict(
+                zip(action.parameters, list(parameters))
+            )
+            new_action = create_action_with_given_subs(
+                self._problem, action, self._simplifier, self._substituter, subs
+            )
+            self._grounded_actions[key] = new_action
+            return new_action
+
+    def ground_problem(self) -> CompilerResult:
```
mikand, line 104: This is OK, but a possibly more useful interface could be a function returning an iterator of pairs `<old_action, ground_action>`, as it can be used to easily implement a grounder but also other codes. What do you think?

unified_planning/engines/sequential_simulator.py
```diff
@@ -207,10 +183,52 @@ def _get_applicable_events(self, state: "up.model.ROState") -> Iterator["Event"]
...
+            # total grounding
+            self._all_events_grounded = True
+            grounding_result = self._grounder.ground_problem()
+            grounded_problem: "up.model.Problem" = cast(
+                up.model.Problem, grounding_result.problem
+            )
+            lift_map = grounding_result.map_back_action_instance
+            assert lift_map is not None
+
+            # for every grounded action, translate it in an Event
```
mikand, line 201: As you already recognized in the comment, this code does something reusable, make a function out of it! :smile:

unified_planning/engines/sequential_simulator.py
```diff
@@ -226,14 +244,35 @@ def _get_events(
...
+            if isinstance(grounded_action, up.model.InstantaneousAction):
+                event_list: List[Event] = [
+                    InstantaneousEvent(
+                        grounded_action.preconditions,
+                        grounded_action.effects,
+                        grounded_action.simulated_effect,
+                    )
+                ]
+            else:
+                raise NotImplementedError
```
mikand, line 271: OK, but add a message to explain the reason of this error

mikand (changes requested):
@Framba-Luca Excellent work! I left some comments for improving the code, but it looks very good already!

unified_planning/engines/compilers/grounder.py
```diff
@@ -32,6 +32,152 @@
...
+        :return: The `action` grounded with the given `parameters` or `None` if the grounded
+            action does not have `effects` or the `action conditions` can be easily evaluated as a
+            contradiction.
+        """
+        assert len(action.parameters) == len(
+            parameters
+        ), "The number of given parameters for the grounding is different from the action's parameters"
+        key = (action, tuple(parameters))
+        if key in self._grounded_actions:  # The action is already created
+            return self._grounded_actions[key]
```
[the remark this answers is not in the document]
Framba-Luca: I thought about this, and the way I would access the dictionary only once would be:

```
value = self._grounded_actions.get(key, None)
if value is not None:
     return value
```

The problem with this code is that `None` is actually a valid result for a grounded action. It means that the created grounded action does not have a "meaning" (it has no effects or the preconditions are a contradiction) (and here you also have the answer to the first comment "why Optional?" ) . So, given that `None` is a valid result, i couldn't use the code i wanted to use.. 

But now we have to discuss if `None` is actually a valid result for it's meaning or we want to change that; because as it is now, the method that creates the grounded action (`create_action_with_given_subs`) returns `None` if the action created is meaningless.

mikand (approved):
@Framba-Luca Excellent work! I have no further comments.

PR #238, opened 2022-09-13
Implementation of an ANML parser

Framba-Luca:
The test-file on which the parser was initially developed is included in the commit and must be removed before merge


Framba-Luca:
This PR contains the code and testing to parse an ANML problem, without quantifiers and conditional effects.

unified_planning/model/timing.py
```diff
@@ -108,8 +108,10 @@ def __init__(self, delay: Union[int, Fraction], timepoint: Timepoint):
     def __repr__(self):
         if self._delay == 0:
             return f"{self._timepoint}"
-        else:
+        elif self._delay > 0:
```
alvalentini, line 111: delay is always > 0

unified_planning/io/anml_reader.py
```diff
@@ -0,0 +1,618 @@
...
+
+        if timing_type == TK_START and is_global:
+            return up.model.GlobalStartTiming(delay.constant_value())
+        elif timing_type == TK_START:
+            return up.model.StartTiming(delay.constant_value())
+        elif timing_type == TK_END and is_global:
+            return up.model.GlobalEndTiming(delay.constant_value())
+        elif timing_type == TK_END:
+            return up.model.EndTiming(delay.constant_value())
+        else:
```
alvalentini, line 471: delay must be positive

alvalentini (changes requested):

alvalentini:
@Framba-Luca can you list here the current unsupported (by the parser) features that could be modeled by the UP and the assumptions under which the parser works?

Framba-Luca:
> @Framba-Luca can you list here the current unsupported (by the parser) features that could be modeled by the UP and the assumptions under which the parser works?
Under these conditions the current parser should work
1. statements containing the duration of an action are not mixed with other statements ( `duration == 3 and at(l_from);` ).
2. the action duration can be set with: 
     - `duration == expression;`
     - `duration := expression;`
     - `duration CT expression and duration CT expression`, where CT are Compare Tokens, so `>`, `>=`, `<` and `<=`.
     All the other ways are not supported.
3. Statements containing both conditions and effects are not supported ( `(at(l_from) == true) := false);` or `(at(l_from) == true) and (at(l_from) := false);` ).
4. Quantifiers and conditional effects are not supported yet.

Then, when adding the support for quantifiers and conditional effects, i was thinking about those assumptions:
1. Quantifier body does not support intervals, they can only be defined outside
2. Conditional effects are not supported inside an expression block

Let me know if you have any feedback about those.

unified_planning/io/anml_reader.py
```diff
@@ -0,0 +1,810 @@
...
+
+    def _parse_interval(
+        self,
+        interval_res: ParseResults,
+        types_map: Dict[str, "Type"],
+        is_global: bool = False,
+    ) -> Union["Timing", "TimeInterval"]:
+        if len(interval_res) == 0:
+            if is_global:
+                # TODO: decide/understand if an undefined interval outside of an action really means GlobalStartTiming or not
```
Framba-Luca, line 481: Note this missing TODO

unified_planning/io/anml_reader.py
```diff
@@ -0,0 +1,810 @@
...
+        for interval_and_exp in action_body_res:
+            relevant_words = {TK_DURATION, TK_ASSIGN, TK_INCREASE, TK_DECREASE, TK_WHEN}
+            contained_words = count_instances(interval_and_exp, relevant_words)
+            c_duration = contained_words[TK_DURATION]
+            c_when = contained_words[TK_WHEN]
+            c_assign = contained_words[TK_ASSIGN]
+            c_increase = contained_words[TK_INCREASE]
+            c_decrease = contained_words[TK_DECREASE]
+
+            if c_duration > 0:  # handle duration
```
alvalentini, line 396: If a duration expression is not in a form that is supported by the parser we should raise an exception with a clear error message

unified_planning/io/anml_reader.py
alvalentini: It is not allowed to have expressions with an undefined interval, except for constants initialization (currently not supported). If it happens we should raise an exception.

unified_planning/io/anml_reader.py
```diff
@@ -0,0 +1,810 @@
...
+        elif len(interval_res) == 3:
+            l_par = interval_res[0]
+            timing_and_exp = interval_res[1]
+            r_par = interval_res[2]
+            timing_type = timing_and_exp[0]
+
+            assert l_par in (TK_L_PARENTHESIS, TK_L_BRACKET) and r_par in (
+                TK_R_PARENTHESIS,
+                TK_R_BRACKET,
+            ), "parsing error"
```
alvalentini, line 495: For parsing error we can create a custom exception (if not already present) and use it in all these situations.

unified_planning/io/anml_reader.py
```diff
@@ -0,0 +1,810 @@
...
+            expression = [expression]
+        stack: List[Tuple[Union[ParseResults, List], bool, Dict[str, "Variable"]]] = [
+            (expression, False, {})
+        ]
+        solved: List[FNode] = []
+        while 1:
+            try:
+                exp, already_expanded, vars = stack.pop()
+            except IndexError:
+                break
```
alvalentini, line 653: Why not checking the stack size in the while condition?

unified_planning/io/anml_reader.py
```diff
@@ -0,0 +1,810 @@
...
+    except ValueError:
+        return False
+
+
+def count_instances(
+    result: Union[ParseResults, List], strings: Set[str]
+) -> Dict[str, int]:
+    stack: List[Union[ParseResults, List]] = [result]
+    counters: Dict[str, int] = {s: 0 for s in strings}
+    while 1:
```
alvalentini, line 797: Same as above

unified_planning/io/anml_reader.py
```diff
@@ -0,0 +1,810 @@
...
+
+def is_float(string: str) -> bool:
+    try:
+        float(string)
+        return True
+    except ValueError:
+        return False
+
+
+def count_instances(
```
alvalentini, line 792: Do you really need the exact number of occurrences?

alvalentini (changes requested):

unified_planning/io/anml_reader.py
Framba-Luca: I thought that the `len` time complexity was O(n), but I checked and it actually is O(1).

unified_planning/io/anml_reader.py
Framba-Luca: Yes, sometimes it is useful to validate the input without having to re-check all the parsed result.

unified_planning/io/anml_reader.py
mikand: @Framba-Luca why this is still an assert?

unified_planning/io/anml_grammar.py
```diff
@@ -0,0 +1,435 @@
...
+            | type_decl
+            | fluent_decl
+            | action_decl
+            | standalone_expression_block
+            | goal_decl
+            | standalone_timed_expression
+        )
+
+        anml_body = OneOrMore(Group(anml_stmt + Suppress(TK_SEMI)))
+        anml_body.ignore(TK_COMMENT + restOfLine)
```
mikand, line 350: Are we sure this is enough? what about 
```
action x(Location l1) // This is a comment in the middle of an action
{
...
}
```

mikand: Add some test checking comments

mikand (commented):
@Framba-Luca Excellent work! Looks good to me!

mikand:
@arbimo any feedback on this? At the moment it does not support hierarchical problems and also the syntax is likely a bit different from FAPE. Shall we try to integrate these changes here or merge it as a "beta feature" and work on this from master?

unified_planning/test/anml/hydrone.anml
```diff
@@ -0,0 +1,145 @@
...
+  [end] visited(destination) := true;
+};
+
+instance Location start_point, a1, a2, b1, b2, b3, b4, b5, b6;
+
+[start] distance(start_point, b1) := 70;
+[start] distance(start_point, b2) := 2;
+[start] distance(start_point, b3) := 30;
+[start] distance(start_point, b6) := 200;
+[start] distance(start_point, start_point) := -1;
```
arbimo, line 33: ANML normally allows not specifying a value of a state variable/constant fluent. So this line should not be necessary.

In also explicitly support marking value a unknown (with an effect to the `UNKNOWN` value).
This is in my view a crucial feature to combine state variable and temporal planning, with the transition notation `[start,end] loc == kitchen :-> bedroom;`.

arbimo (commented):
This looks like a good starting point to me :+1: 
I just went though the examples and added two comments on the syntax used.Not necessarily things to fix immediately, but at least the support of unknown/undefined value are crucial for the ANML language.

unified_planning/io/anml_reader.py
Framba-Luca: Because, for the grammar we defined, it should be impossible that here we have something different. Let me know if you prefer something different

PR #240, opened 2022-09-26
Complete HDDL parser

arbimo:
This PR completes the support for HDDL parsing in the UP. together with minor improvements and bugfixes to the parser (case insensitivity of types, bugs in equality type checking, ...).

In the PR is also includes 30 (out of 32) domains for the latest IPC on hierarchical planning that are all parsed successfully (the two remaining ones are blocked due to limitation in the UP itself, not the parser).
I believe it would be great to test parsing of many domains in an automated fashion (also for PDDL) however I am unsure where to integrate those domains are they are mainly intended to test the parser, not an integration.

## Current limitations of the parser 

(that this PR does not address but are worth mentioning while I have them in mind):

- we associate a special meaning to the `object` type

I think the root of the problem is that (1) we always need to assign a type to a variable and (2) we do not have a fixed top-type (the root of type hierarchy).
As a result, if no type was given we assume that the user would have wanted to assign to an "object" type and create it for him.

While "object" is often used as a root type in PDDL domains, I believe this is just a convention. It is dangerous to enforce this as "object" is extremely common in domains.
The solution would be to have our own top-type (limited to PDDL ?) for these situations. I can do a follow-up PR for this if that's what we want (that would also fix the TODO left).

- No handling of case insensitivity

This PR adds it for types (because some domains needed it) but general support is still missing.

 - limited error reporting. 

Currently, if a failure occurs during parsing (i.e. the part that is delegated to pyparsing), we would only report a very approximate column/line number that is hidden in the middle of the error text. 
The location (line/col) can be made more precise by forcing the library not to backtrack (with the `-` combinator instead of the `+` combinator). I added this in a few places but it remains to be improved.

If the error occurs in our own code, we just throw an error message without mentioning the location (file+line+column) where the error occurred.
One way to improve this is to have the parser hand out tokens that are annotated with the location in the input file of the string. (i.e. like the CaseInsensitiveToken in this PR but with additional metadata about the string).

To give an idea, this is the kind of messages that can be printed out when you have this information (taken from aries): 
```
> target/release/lcp problems/hddl/ipc/2020-po-Rover/instance.1.pb.hddl
Error: /mnt/data/work/aries/problems/hddl/ipc/2020-po-Rover/domain.hddl:102:14
        (not (visite ?mid))
              ^^^^^^ Unknown atom
```

It would be hard to retrofit completely in the PDDL parser but we might want to consider this early on for the ANML parser.

- The parser would fail to parse actions with conflicting effects (and in particular ones with wiht delete-before-add pattern)
- the parser would fail to parse problem where there is multiple type inheritance (that's probably OK, as I only saw this features used for the first time yesterday)


unified_planning/model/mixins/fluents_set.py
```diff
@@ -135,8 +135,10 @@ def add_fluent(
             fluent = up.model.fluent.Fluent(
                 fluent_or_name, typename, None, env=self.env, **kwargs
             )
-        if self._has_name_method(fluent.name):
-            raise UPProblemDefinitionError("Name " + fluent.name + " already defined!")
+        if self.has_fluent(fluent.name):
+            raise UPProblemDefinitionError(
+                "Fluent with name " + fluent.name + " already defined!"
+            )
```
alvalentini, line 141: Why do we need this change?

unified_planning/model/mixins/user_types_set.py
```diff
@@ -41,7 +41,7 @@ def _add_user_type(self, type: "up.model.types.Type"):
         assert type.is_user_type()
         if type not in self._user_types:
             t = cast(_UserType, type)
-            if self._has_name_method(t.name):
+            if self.has_type(t.name):
```
alvalentini, line 44: Same as above, why do we need this?

alvalentini (commented):
@arbimo I do not have many comments, the changes are good for me! I've just added two small questions directly in the code.

I also agree that the limitations you mentioned should be fixed/improved in a future PR.

unified_planning/model/mixins/fluents_set.py
arbimo: The previous implementation disallowed to have two things with the same name in the UP even if they are non-colliding (e.g. a type and an action).
While this is generally bad practice, it is not forbidden in the PDDL standard and some domains rely on it (in particular, at least one domain in the HTN IPC relies on it).

Maybe we could issue a warning in this case, but he exception would rule out valid, unambiguous, domains.

unified_planning/model/mixins/user_types_set.py
arbimo: Same (with a different domain requiring this one).

unified_planning/io/pddl_reader.py
```diff
@@ -582,19 +640,20 @@ def _parse_subtasks(
...
         for g in domain_res.get("constants", []):
-            if len(g) <= 1 or g[1] == "object":
+            if len(g) <= 1 or g[1] == Object:
                 return True
         for a in domain_res.get("actions", []):
             for g in a.get("params", []):
-                if len(g) <= 1 or g[1] == "object":
+                if len(g) <= 1 or g[1] == Object:
                     return True
+        # TODO probably missing spots there + is this something we really want
```
arbimo, line 656: This is something that should be handled as well. Either by :
 - completing this (to include methods, tasks ...). This would not fix the issue of the incorrect special treatment of "object" though.
 - having a special top type that is the root of the type hierarchy (often called the "Any"/"Top" type in programming languages). This could substantially simplify parsing by having the `parameter` parser always provide a type (no explicit type meaning the top type which would remove the need for the statements like `g[1] if len(g) > 1 else "object"`).

Option 2 should be the object of a distinct PR.

unified_planning/model/mixins/fluents_set.py
arbimo: Yes, you're right. I missed this possible ambiguity. An option would be to have a more fine grained check there.

alvalentini: I agree that we should have a more fine grained check. I propose to undo these changes in this PR, remove the problematic domains from the tests and open a dedicated issue.

Then we will handle this in a new PR.

unified_planning/io/pddl_reader.py
alvalentini: I propose to go with option 1 in this PR (if it is something easy to do) and, as you said, go with option 2 in a new PR.

arbimo: Commited option 1. I can draft a PR implementing option 2 once this one is merged.

unified_planning/model/mixins/fluents_set.py
alvalentini: @arbimo what do you think about my proposal here?

arbimo: I am unsure about this one. In python, we never manipulate the name but instead use the previously constructed python objects, so I do not see any risk of ambiguity there. There I believe it would make sense to remove the check entirely (which would also simplify this error prone part of the code with cross-mixins dependencies).

In the PDDL parser, we can detect and raise the ambiguity at its point of occurrence : when parsing expressions. This would also give the opportunity to provide better error messages.

alvalentini: I agree that we should relax these checks, even removing them entirely, but I would do it in a separate PR.
Removing them entirely also means to update the ANML/PDDL writers to handle the possible ambiguities.

alvalentini: @arbimo if we relax/remove the checks in a different PR and we undo the changes from here, I think we can merge the PR.

arbimo: Yes, I just pushed the changes, and removed the three impacted domains that now failed to parse.

unified_planning/model/mixins/user_types_set.py
arbimo: Handled together with the previous issue.

PR #246, opened 2022-10-17
Support nested quantifiers in PDDL parsing

### LINKED ISSUE #239: Parsing error with nested quantifiers
We noticed that in the pddl_reader the support for nested quantifiers seems broken. More precisely we tried to parse a problem involving the goal **Forall((?x)(and (at ?x) (Forall (?y) (connected ?x ?y))))**, and got a parsing error from the library as follows:

**".venv/lib/python3.10/site-packages/unified_planning/io/pddl_reader.py", line 376, in _parse_exp_
    assert act is not None
AssertionError**

The issue can be reproduced by parsing the domain and the problem attached to this issue.
[operator_test.zip](https://github.com/aiplan4eu/unified-planning/files/9611079/operator_test.zip)

Framba-Luca:
Extended PDDL parser code to support nested quantifers when the quantifier does not contain an assignment.

Fixes #239


unified_planning/test/pddl/operator/domain.pddl
```diff
@@ -0,0 +1,17 @@
+;Header and description
+
+(define (domain operator_test_domain)
+
+(:predicates ;todo: define predicates here
```
alvalentini, line 5: We can remove this TODO comment
```suggestion
(:predicates 
```

unified_planning/test/pddl/operator/problem.pddl
```diff
@@ -0,0 +1,22 @@
...
+(:goal
+  (and
+    (forall 
+      (?x)
+      (and
+        (at ?x)
+        (forall (?y) (connected ?x ?y))
+      )
+    )
+  )
```
alvalentini, line 20: It seems an impossible problem, can we update it with a problem that makes sense and has nested quantifiers?

alvalentini:
@Framba-Luca Rebase needed!

mikand (commented):
@Framba-Luca This looks OK, but the test does not seem to have a nested quantifier, I was expecting a forall with a nested exists or something like that?

PR #251, opened 2022-10-24
Optimality semantic

alvalentini:


mikand (commented):
LGTM, but we need to add some documentation explaining the expected behavior in all the cases.

mikand:
@alvalentini For me this can be merged.
@hstairs @roeger Can we merge this or shall we wait for FD and ENHSP to be updated?

roeger:
Don't wait for Fast Downward, I can update it quickly (~1 day) once this is merged.

PR #255, opened 2022-10-28
Multi-agent goals

Alee08:
Added the goals in the agent class


alvalentini:
@Alee08 The PR looks good! I have only two comments:
- you need to update the `kind` method of the `MultiAgentProblem` class to consider also the `Agent`'s goals;
- I think it would be useful to have an explanation of the semantics of these new `Agent`'s goals.

karpase:
@Alee08 
I think you also need to update the clone method of multiagent_problem to also clone the agent's goals.
Even better - implement a clone method for the agent, and then just use that in multiagent_problem.

Issue #232, opened 2022-08-19
oversubscription[fast-downward] claims to be optimal on suboptimal solution


roeger:
Problem [depots2.py.txt](https://github.com/aiplan4eu/unified-planning/files/9381430/depots2.py.txt) has no metric set and the optimal solution length is 15.  Still oversubscription[fast-downward] gives a plan with 16 actions and PlanGenerationResultStatus.SOLVED_OPTIMALLY.

Is this a bug or do I misinterpret the optimality guarantee?

roeger:
I had a look into the code of `oversubscription_planner.py`. If the subsolver solves the problem with all subgoals then oversubscription sets the result status to SOLVED_OPTIMALLY, independently of whether the subsolver gives this guarantee. It should be a one-liner to add the additionally required check.

roeger:
Ah, correct, it depends on the metric. I am not sure whether SOLVED_SATISFICING is the right choice in all cases without metric. Do you assume there is no such thing as an optimal plan in an instance without metric? In the classical setting, PDDL defaults to minimizing plan length if no metric is specified.

roeger:
No, I don't want to reopen the discussion but it means that Ipotentially need to adapt the Fast Downward planning engine and the test cases. Where are the intended semantics documented?

alvalentini:
Adding the semantics documentation will be part of #251.

PR #256, opened 2022-10-31
Only signal optimal solution if subengine does so.

roeger:
This should fix issue #232.


roeger:
I am closing this pull request because it is subsumed by PR #251 (which was not cross-linked with issue #232).

PR #258, opened 2022-11-09
Improve DNF compiler when it creates an invalid action.

Framba-Luca:
Now, in the DNF_remover, if an action is invalid (False precondition or no effects), it is not added to the compiled problem.

### DIFF
--- a/unified_planning/io/pddl_writer.py
+++ b/unified_planning/io/pddl_writer.py
@@ -181,10 +181,11 @@ class ConverterToPDDLString(walkers.DagWalker):
     ):
         walkers.DagWalker.__init__(self)
         self.get_mangled_name = get_mangled_name
+        self.simplifier = env.simplifier
 
     def convert(self, expression):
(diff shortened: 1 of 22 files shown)


unified_planning/io/pddl_writer.py
```diff
@@ -181,11 +181,10 @@ def __init__(
     ):
         walkers.DagWalker.__init__(self)
         self.get_mangled_name = get_mangled_name
-        self.simplifier = env.simplifier
 
     def convert(self, expression):
         """Converts the given expression to a PDDL string."""
-        return self.walk(self.simplifier.simplify(expression))
+        return self.walk(expression.simplify())
```
mikand, line 187: Why this change? In this way you are forcing the use of the "global" simplifier instead of the one in the environment used to construct this converter, no?

mikand (changes requested):
LGTM except for the comment I left

PR #259, opened 2022-11-09
Enforced pyparsing version to be >= 3

### LINKED ISSUE #253: Restict pyparsing to versions >=3 ?
The PDDL parser currently fails with old pyparsing versions (version 2.4.2 on this machine).

```
  File "/home/pherail/Saved_Code/method_learner/python_scripts/extract_sequences_lcp_out_upf.py", line 31, in <module>
    reader = PDDLReader()
  File "/home/pherail/.local/lib/python3.8/site-packages/unified_planning/io/pddl_reader.py", line 316, in __init__
    grammar = PDDLGrammar()
  File "/home/pherail/.local/lib/python3.8/site-packages/unified_planning/io/pddl_reader.py", line 72, in __init__
    tpe = name.copy().add_parse_action(lambda t: CaseInsensitiveToken(t))
AttributeError: '_WordRegex' object has no attribute 'add_parse_action'
```

Do we have any reason to support pyparsing versions `< 3` ? If yes, then we should probably test it in the CI to avoid such bugs to enter.
Otherwise, we can enforce the version in the requirements.

Framba-Luca:
closes #253

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-pyparsing
+pyparsing >= 3
 networkx
--- a/unified_planning/io/pddl_reader.py
+++ b/unified_planning/io/pddl_reader.py
@@ -17,7 +17,6 @@ from itertools import product
 import unified_planning as up
 import unified_planning.model.htn as htn
 import unified_planning.model.walkers
-import pyparsing
 import typing
 from unified_planning.environment import Environment, get_env
 from unified_planning.exceptions import UPUsageError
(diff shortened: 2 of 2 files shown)


mikand (changes requested):
OK, but also update the pddlreader module to import only only pyparsing 3 and add an assertion to explain why we will not execute with version 2.

mikand:
Another thing, if we properly use pyparsing 3 we should use the snake_case method versions instead of the camelCase ones...

Issue #254, opened 2022-10-27
Performance problem of PDDL parser


arbimo:
It seems that the PDDL parser has some important performance problems when parsing non trivial problems. Below is the runtime in seconds to parse some HDDL problems. Runtimes are often above one second and even go up to 23 seconds (!) for instance [23].
This instance in particular has a fairly big (though not completely unusual) initial state expression.

```
=== [0]  2020-po-Monroe-Partially-Observable === 1.3065104484558105
=== [1]  2020-to-Blocksworld-HPDDL === 0.1640629768371582
=== [2]  2020-to-Transport === 0.143265962600708
=== [3]  2020-to-Robot === 0.1581273078918457
=== [4]  2020-po-Monroe-Fully-Observable === 1.7055234909057617
=== [5]  2020-to-Towers === 0.12245368957519531
=== [6]  2020-to-Blocksworld-GTOHP === 0.20174217224121094
=== [7]  2020-to-Logistics-Learned-ECAI-16 === 0.7226729393005371
=== [8]  2020-po-Satellite === 0.23903155326843262
=== [9]  2020-to-Monroe-Fully-Observable === 2.0467567443847656
=== [10] 2020-to-Childsnack === 0.3639533519744873
=== [11] 2020-to-Depots === 0.27066755294799805
=== [12] 2020-to-AssemblyHierarchical === 0.5382301807403564
=== [13] 2020-to-Minecraft-Regular === 1.166816234588623
=== [14] 2020-to-Entertainment === 1.3003723621368408
=== [15] 2020-to-Multiarm-Blocksworld === 0.23566651344299316
=== [16] 2020-po-PCP === 0.31250834465026855
=== [17] 2020-to-Factories-simple === 0.28209948539733887
=== [18] 2020-to-Hiking === 0.5998499393463135
=== [19] 2020-to-Elevator-Learned-ECAI-16 === 0.4780910015106201
=== [20] 2020-po-Transport === 0.1967909336090088
=== [21] 2020-to-Snake === 0.3077516555786133
=== [22] 2020-to-Woodworking === 2.1901118755340576
=== [23] 2020-to-Minecraft-Player === 23.85342025756836
=== [24] 2020-to-Monroe-Partially-Observable === 3.194375514984131
=== [25] 2020-po-Rover === 0.7088096141815186
=== [26] 2020-to-Rover-GTOHP === 0.865910530090332
=== [27] 2020-to-Satellite-GTOHP === 0.3189351558685303
```

[...]

PR #260, opened 2022-11-09
[pddl] Performance improvement in PDDL parser.

arbimo:
This is a straightforward adaptation of the alternative implementation of `pyparsing.nestedExpr` proposed by @mikand in #254.

On my machines, it reduces the time to parse all 27 hddl instances of  `test/hddl` from 42s to 5 seconds, making it an excellent mitigation. I would not say it fixes the issue, but it certainly makes it less urgent to handle.

### DIFF
--- a/unified_planning/io/pddl_reader.py
+++ b/unified_planning/io/pddl_reader.py
@@ -66,14 +66,21 @@ Object = CaseInsensitiveToken("object")
 TypesMap = Dict[CaseInsensitiveToken, unified_planning.model.Type]
 
 
-def nested_expr():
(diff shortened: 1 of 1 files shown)


arbimo:
Thanks for trying these out @mikand ! 
The latter optimization (restricting the recursion depth) does not seem to provide any improvement on my machine (but is not detrimental either). Unless you have seen some benefits elsewhere we might to leave it out?

Otherwise I think could go in, it does provide a very substantial improvement for large models.

mikand:
Merged as a mitigation fix

PR #261, opened 2022-11-10
Default of mypy now also checks untyped functions

Framba-Luca:


.mypy.ini
```diff
@@ -1,8 +1,12 @@
 [mypy]
+check_untyped_defs = True
 
 [mypy-unified_planning.grpc.generated.*]
 ignore_errors = True
 
+[mypy-unified_planning.test.*]
+ignore_errors = True
```
mikand, line 8: Can we avoid this? I had a quick look and it seems that several errors are meaningful (e.g. you call a compiler and you do not check that the result problem is not None) having such checks seems useful, no?

mikand (changes requested):

.mypy.ini
Framba-Luca: Yes, there were actually a lot of things to adjust. Some checks were useful, others were a little bit borderline, but it helped to find some inconsistencies trough the library.

mikand (changes requested):
@Framba-Luca Excellent work!

PR #263, opened 2022-11-14
Added a method to parse PDDL from strings and some testing

### LINKED ISSUE #262: Allow PDDLParser to parse strings directly
It would be good to have PDDLReader take not only file names, but also the PDDL string directly and pass it.
Something like: 
reader=PDDLReader()
problem=reader.parse_problem_string(domain_string,problem_string)

This would make calling one planner from another easier, at least for us at the CPOR branch.

Framba-Luca:
Closes #262

### DIFF
--- a/unified_planning/io/pddl_reader.py
+++ b/unified_planning/io/pddl_reader.py
@@ -20,10 +20,9 @@ import unified_planning.model.walkers
 import typing
 from unified_planning.environment import Environment, get_env
 from unified_planning.exceptions import UPUsageError
-from unified_planning.model import FNode
 from collections import OrderedDict
 from fractions import Fraction
-from typing import Dict, Tuple, Union, Callable, List, cast
+from typing import Dict, Union, Callable, List, cast
 
 
 import pyparsing
@@ -1192,10 +1191,9 @@ class PDDLReader:
         """
         domain_res = self._pp_domain.parse_file(domain_filename, parse_all=True)
 
+        problem_res = None
         if problem_filename is not None:
             problem_res = self._pp_problem.parse_file(problem_filename, parse_all=True)
-        else:
-            problem_res = None
 
         return self._parse_problem(domain_res, problem_res)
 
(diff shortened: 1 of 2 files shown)


unified_planning/io/pddl_reader.py
```diff
@@ -23,7 +23,7 @@
 from unified_planning.model import FNode
 from collections import OrderedDict
 from fractions import Fraction
-from typing import Dict, Union, Callable, List, cast
+from typing import Dict, Tuple, Union, Callable, List, cast
```
mikand, line 26: Is this useful?

unified_planning/io/pddl_reader.py
```diff
@@ -1190,3 +1176,46 @@ def declare_type(
...
+        :param domain_filename: The path to the file containing the `PDDL` domain.
+        :param problem_filename: Optionally the path to the file containing the `PDDL` problem.
+        :return: The `Problem` parsed from the given pddl domain + problem.
+        """
+        domain_res = self._pp_domain.parse_file(domain_filename, parse_all=True)
+
+        if problem_filename is not None:
+            problem_res = self._pp_problem.parse_file(problem_filename, parse_all=True)
+        else:
+            problem_res = None
```
mikand, line 1198:
```suggestion
        problem_res = None
        if problem_filename is not None:
            problem_res = self._pp_problem.parse_file(problem_filename, parse_all=True)
```

unified_planning/test/test_pddl_io.py
```diff
@@ -323,6 +323,14 @@ def test_depot_reader(self):
         self.assertEqual(len(problem.actions), 5)
         self.assertEqual(len(list(problem.objects(problem.user_type("object")))), 13)
 
+        with open(domain_filename, "r") as file:
```
mikand, line 326: Just to be sure:
```suggestion
        with open(domain_filename, "r", encoding="utf-8") as file:
```

mikand (approved):
LGTM! I left a couple of very minor comments for your consideration

PR #266, opened 2022-11-15
MA-PDDL writer

Alee08:


unified_planning/io/ma_pddl_writer.py
```diff
@@ -0,0 +1,1066 @@
...
+    def add_agent_prefix(self, string: str, fluents_list: List["Fluent"]):
+        match = re.search(r"\((\w+)\s\?", string)
+        if match:
+            found = match.group(1)
+        else:
+            return string
+        for obj in fluents_list:
+            if obj.name == found:
+                return re.sub(r"\((\w+)\s\?", r"(a_\1 ?", string)
+        return string
```
mikand, line 1045: 1. These methods are very strange for a printer: why do you need to build new expressions? I think it would be MUCH simpler to add an optional `agent` parameter to the `ConverterToMAPDDLString` and add the agent name where needed. For example:
```python
class ConverterToMAPDDLString(ConverterToPDDLString):
    """Expression converter to a MA-PDDL string."""

    def __init__(
        self,
        env: "up.environment.Environment",
        get_mangled_name: Callable[
            [
                Union[
                    "up.model.Type",
                    "up.model.Action",
                    "up.model.Fluent",
                    "up.model.Object",
                    "up.model.multi_agent.Agent",
                ]
            ],
            str,
        ],
        agent : Optional["up.model.multi_agent.Agent"]
    ):
        ConverterToPDDLString.__init__(self, env, get_mangled_name)
        self.agent = agent

    def walk_dot(self, expression, args):
        agent = expression.agent()
        fluent = expression.args[0].fluent()
        objects = expression.args[0].args
        return f'({self.get_mangled_name(fluent)} {self.get_mangled_name(agent)} {" ".join([self.convert(obj) for obj in objects])})'

    def walk_fluent_exp(self, expression, args):
        fluent = expression.fluent()
        return f'({self.get_mangled_name(fluent)} {self.get_mangled_name(self.agent) if self.agent is not None and fluent in self.agent.fluents else ""}{" " if len(args) > 0 else ""}{" ".join(args)})'
```
In this way, you can simply create different `ConverterToMAPDDLString` objects for different agents and use them. This should greatly simplify the printing, because you simply call the appropriate printer on a problem expression, without creating new FNodes.

2. these functions seems seriously buggy: why adding a `Not` if an expression is a `Not`? What happens if an expression has more than two arguments?  `sub_exp` is never used....

unified_planning/io/ma_pddl_writer.py
```diff
@@ -0,0 +1,1066 @@
...
+                "up.model.Object",
+                "up.model.Parameter",
+                "up.model.Variable",
+                "up.model.multi_agent.Agent",
+            ],
+        ] = {}
+        # those 2 maps are "simmetrical", meaning that "(otn[k] == v) implies (nto[v] == k)"
+        self.domain_objects: Optional[Dict[_UserType, Set[Object]]] = None
+        self.domain_objects_agents: Dict[up.model.multi_agent.Agent, str]
+        self.domain_fluents_agents: Dict[up.model.multi_agent.Agent, List[Fluent]]
```
mikand, line 129: What is the purpose of this state? The Printer class should ideally be reusable across multiple problems and adding state requires cleanup (that seems to be missing)

unified_planning/io/ma_pddl_writer.py
```diff
@@ -0,0 +1,1066 @@
...
+        :return: The `MA-PDDL` name of the given item.
+        """
+        try:
+            return self.otn_renamings[item]
+        except KeyError:
+            raise UPException(
+                f"The item {item} does not correspond to any item renamed."
+            )
+
+    def _populate_domain_objects(
```
mikand, line 919: What is the purpose of this function? Please an an explanation comment

unified_planning/io/ma_pddl_writer.py
```diff
@@ -0,0 +1,1066 @@
...
+                    )
+                else:
+                    out.write(f" {converter.convert(p)}")
+            out.write(f"))")
+            out.write("\n)")
+            out.seek(0)
+            ag_problem = out.read()
+            ag_problems[ag.name] = ag_problem
+            out.truncate(0)
+            out.seek(0)
```
mikand, line 746: WTF?
Why do you take `out` in input at all? You can simply use a [`StringIO`](https://docs.python.org/3/library/io.html#io.StringIO) for every agent and return their `.getvalue()`...

mikand: BTW, you are not guaranteed to have `seek` and `truncate` for every type of `IO[str]`, only for files and buffers...

unified_planning/io/ma_pddl_writer.py
```diff
@@ -0,0 +1,1066 @@
...
+                        out.write(")")
+                    out.write(")\n")
+                else:
+                    raise NotImplementedError
+            out.write(")\n")
+            out.seek(0)
+            ag_domain = out.read()
+            ag_domains[ag.name] = ag_domain
+            out.truncate(0)
+            out.seek(0)
```
mikand, line 651: See comment below: this is very awkward...

mikand (changes requested):
@Alee08 @fabio-patrizi Unfortunately, I think this requires a bit more work: see my comments for the details but from the high-level POV you need to avoid creating needless expressions just for printing (and I think there are bugs in the current creation of such expressions) and fix the interface with streams.

Moreover, as a general point, you need to rebase the code to the current master and check the use of `get_mangled_name` that is not always used on the agents names.

unified_planning/io/ma_pddl_writer.py
Alee08: Thanks for the feedback, I understand that using one StringIO for each agent and returning their .getvalue() is much easier. Initially I wanted the multi-agent _write_domain() to take the output as a parameter and thus look similar to the single-agent one. 
Done!

unified_planning/io/ma_pddl_writer.py
Alee08: The simplify() method aims to simplify the expression but does not take into account the "Not" operator. This is why I check if the FNode is of type "Not" and if it is, I remove it using `self.problem.env.expression_manager.Not(arg)`. This function will remove the "Not" operator if it is present, otherwise, it adds it.
I agree with your suggestion to use `ConverterToMAPDDLString` as it simplifies the printing process.
Done!

unified_planning/io/ma_pddl_writer.py
Alee08: The states `domain_objects_agents` and `domain_fluents_agents` are used to keep track of fluents and objects of other agents that are referenced using the DOT operator in the preconditions. They allow the addition of these fluents and objects to the public predicates and constants in the MA-PDDL domain of the current agent. This is necessary since these fluents are not present in the current agent, and the current agent's fluents do not use the objects.

unified_planning/io/ma_pddl_writer.py
Alee08: The purpose of this function is to populate the `domain_objects`, `domain_objects_agents`, and `domain_fluents_agents` class variables. These variables are used to keep track of the objects and fluents of other agents that are referenced using the DOT operator in the preconditions of the actions of the current agent. The function iterates through the actions of the current agent and checks the preconditions for any references to other agents and fluents (DOT). 
If found, it updates the class variables with the relevant information. This information is later used to add the fluents and objects to the public predicates and constants in the MA-PDDL domain of the current agent. This function is also present in pddl_writer.py and allows you to update the "domain_objects" (objects present in the preconditions of the actions).

unified_planning/io/ma_pddl_writer.py
Alee08: Done!

mikand:
@Alee08 It seems that you had some problems with the rebase and several unrelated commits entered the branch.
Let us know if you cannot fix this and we will give a try.

Alee08:
> @Alee08 It seems that you had some problems with the rebase and several unrelated commits entered the branch. Let us know if you cannot fix this and we will give a try.

@mikand Yes, there is a unit test that continues to fail.

alvalentini:
@Alee08 It seems that there are still several unrelated commits (747 commits in total) in this PR. Let us know if you want that we will a try to fix that.

Alee08:
> @Alee08 It seems that there are still several unrelated commits (747 commits in total) in this PR. Let us know if you want that we will give a try to fix that.

Yes thanks. I created a new PR [MulitAgent-io #324](https://github.com/aiplan4eu/unified-planning/pull/324) equal to this [MA-PDDL writer #266](https://github.com/aiplan4eu/unified-planning/pull/266), if the current branch is more difficult to merge.

Alee08:
> > @Alee08 It seems that there are still several unrelated commits (747 commits in total) in this PR. Let us know if you want that we will give a try to fix that.
> 
> @alvalentini yes thanks. I created a new PR [MulitAgent-io #324](https://github.com/aiplan4eu/unified-planning/pull/324) equal to this [MA-PDDL writer #266](https://github.com/aiplan4eu/unified-planning/pull/266), if the current branch is more difficult to merge.

mikand:
@Alee08 It seems that you are working on #324, can we close this one?

Alee08:
> @Alee08 It seems that you are working on #324, can we close this one?

@mikand Yes, thank you!

PR #267, opened 2022-11-18
Add epsilon timing to temporal examples

Shi-Raida:
The plan of two temporal examples lacked an espilon timing between them so that the effects are applied before the next action.


arbimo:
I believe these are also part of #297 
(In fact they were discovered when working on another temporal validator as well). 
@Framba-Luca @Shi-Raida : is it OK to close this while waiting for the larger PR ? If #297 is not expected to land soon, we might just as well merge this one.

mikand:
@Framba-Luca I would rebase and merge this, as merging #297 will take a while.

PR #268, opened 2022-11-18
Now FluentExp method accepts a general Iterable and not only a Tuple.

Framba-Luca:
I don't see any reasons why the `FluentExp` method must accept only a tuple and not every `Iterable`.


mikand (changes requested):
OK! Maybe add an example showing the use-case?

PR #278, opened 2022-12-01
Disable checks on simulator

Framba-Luca:
This PR allows to disable automatic checks on the simulators at construction time (when the checks are done for the simulators)


Framba-Luca:
Closes #278

PR #291, opened 2022-12-16
Portfolio Interface

Framba-Luca:
Created a PortfolioSelectorMixin, created a PortfolioSelector Operation Mode and documented

### DIFF
--- a/unified_planning/engines/engine.py
+++ b/unified_planning/engines/engine.py
@@ -14,30 +14,35 @@
 #
 """This module defines the engine interface."""
 
+from enum import Enum
 from unified_planning.model import ProblemKind
 from unified_planning.engines.credits import Credits
 from typing import Optional
 
 
+class OperationMode(Enum):
+    ONESHOT_PLANNER = "oneshot_planner"
+    ANYTIME_PLANNER = "anytime_planner"
+    PLAN_VALIDATOR = "plan_validator"
+    PORTFOLIO_SELECTOR = "portfolio_selector"
+    COMPILER = "compiler"
+    SIMULATOR = "simulator"
+    REPLANNER = "replanner"
+
+
 class EngineMeta(type):
     def __new__(cls, name, bases, dct):
         obj = type.__new__(cls, name, bases, dct)
-        oms = [
-            "oneshot_planner",
-            "anytime_planner",
-            "plan_validator",
-            "portfolio_selector",
-            "compiler",
-            "simulator",
-            "replanner",
-        ]
         for base in bases:
-            for om in oms:
-                if hasattr(base, "is_" + om) and getattr(base, "is_" + om)():
-                    setattr(obj, "is_" + om, staticmethod(lambda: True))
-        for om in oms:
-            if not hasattr(obj, "is_" + om):
-                setattr(obj, "is_" + om, staticmethod(lambda: False))
+            for om in OperationMode:
+                if (
+                    hasattr(base, "is_" + om.value)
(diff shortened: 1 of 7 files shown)


unified_planning/engines/engine.py
```diff
@@ -14,29 +14,35 @@
 #
 """This module defines the engine interface."""
 
+from enum import Enum
 from unified_planning.model import ProblemKind
 from unified_planning.engines.credits import Credits
 from typing import Optional
 
 
+class OperationMode(Enum):
```
mikand, line 23: Missing doc

unified_planning/engines/factory.py
```diff
@@ -362,13 +364,14 @@ def _add_meta_engine(
 
     def _get_engine_class(
         self,
-        engine_kind: str,
+        engine_kind: "OperationMode",
```
mikand, line 367: At this point I would rename this:
```suggestion
        operation_mode: "OperationMode",
```

unified_planning/engines/factory.py
```diff
@@ -362,13 +364,14 @@ def _add_meta_engine(
...
         name: Optional[str] = None,
         problem_kind: ProblemKind = ProblemKind(),
         optimality_guarantee: Optional["OptimalityGuarantee"] = None,
         compilation_kind: Optional["CompilationKind"] = None,
         plan_kind: Optional["PlanKind"] = None,
         anytime_guarantee: Optional["AnytimeGuarantee"] = None,
+        operation_mode: Optional["OperationMode"] = None,
```
mikand, line 374: I would rename this, for the inetrnal methods, while for the public ones it's OK: 
```suggestion
        target_operation_mode: Optional["OperationMode"] = None,
```

unified_planning/engines/factory.py
```diff
@@ -796,6 +829,37 @@ def Replanner(
...
+            raise up.exceptions.UPUsageError(
+                "For PortfolioSelector Operation Mode, at least one",
+                " of name and operation_mode kwargs must be set.",
+            )
+        return self._get_engine(
+            OperationMode.PORTFOLIO_SELECTOR,
+            name=name,
+            params=params,
+            problem_kind=problem_kind,
+            operation_mode=operation_mode,
```
mikand, line 860:
```suggestion
            target_operation_mode=operation_mode,
```

mikand (changes requested):
@Framba-Luca Excellent work. I left some very minor comments, then we can merge!

unified_planning/engines/factory.py
```diff
@@ -800,6 +838,37 @@ def Replanner(
...
         )
 
+    def PortfolioSelector(
+        self,
+        *,
+        name: Optional[str] = None,
+        params: Optional[Dict[str, Any]] = None,
+        problem_kind: ProblemKind = ProblemKind(),
+        operation_mode: Optional[Union["OperationMode", str]] = None,
+    ) -> "up.engines.engine.Engine":
```
alvalentini, line 848: I think that other parameters could be useful, for example `optimality_guarantee` in case the operation mode is `OperationMode.ONESHOT_PLANNER`.

mikand:
@alvalentini @Framba-Luca After a bit of discussion, we decided to remove the `operation_mode` parameter from the `PortfolioSelector` OM to keep things simpler. Moreover we need to add the `optimality_guarantee` parameter to mimik the `OneshotPlanning` OM.

@alvalentini Can you do the fix and then merge?

PR #297, opened 2022-12-19
Temporal Validator

Framba-Luca:

PR #303, opened 2023-01-09
Implemented grounder on quality metrics

Framba-Luca:


unified_planning/engines/compilers/grounder.py
```diff
@@ -82,6 +92,14 @@ def __init__(
...
         self._simplifier = Simplifier(env, problem)
 
+    @property
+    def substituter(self) -> Substituter:
+        return self._substituter
+
+    @property
+    def simplifier(self) -> Simplifier:
+        return self._simplifier
+
```
alvalentini, line 102: Why is it needed to expose these class attributes?

Framba-Luca: Because they are used outside just to ground 1 expression. I thought about a `ground_expression` method, but it just calls the substituter and simplifier over the expression.

unified_planning/engines/sequential_simulator.py
```diff
@@ -283,6 +289,11 @@ def supported_kind() -> "up.model.ProblemKind":
         supported_kind.set_effects_kind("INCREASE_EFFECTS")
         supported_kind.set_effects_kind("DECREASE_EFFECTS")
         supported_kind.set_simulated_entities("SIMULATED_EFFECTS")
+        supported_kind.set_quality_metrics("ACTIONS_COST")
+        supported_kind.set_quality_metrics("PLAN_LENGTH")
+        supported_kind.set_quality_metrics("OVERSUBSCRIPTION")
+        supported_kind.set_quality_metrics("MAKESPAN")
+        supported_kind.set_quality_metrics("FINAL_VALUE")
```
alvalentini, line 296: This change is related to the comment in the #297

Framba-Luca: Yes, I think this also has to be discussed in a meeting: probably the simulator will ignore quality metrics, but the validator I have no idea, still has to be discussed at all

alvalentini: If we add the quality metrics here, we should add them also in the sequential plan validator

PR #307, opened 2023-01-11
Plan Repair

alvalentini:


mikand:
@arbimo Have a look at the basic `repair` method and its result: any feedback?

arbimo:
The API look fine to me and in fact I don't think we can provide a more advanced API that is general enough for all kinds of planning (in particular classical planning). We might want go beyond that when considering it jointly with execution (but that would be a later step).

Something that could be clarified is the expected behavior of the API. I think what we have in mind is:
 1. the plan is invalid -> a valid plan
 2. the pan is valid -> the same plan (or another valid one)
 3. the plan is invalid and an optimality metric is provided -> an optimal plan
 4. the plan is valid and an optimality metric is provided -> an optimal plan

The *repair* semantically mostly cover the 3 first cases above (even if the third is a bit strange). It make sense to support all four with the *repair* method as it shouldn't cost more in the implementation. Note that in all these cases, it is OK to just ignore the provided plan and call a one-shot planner.

However case (4.) is a very useful thing in itself, often called warm-starting in MILP/CP solvers (typical usage, a greedy method provides a reasonable solution that is then improved by the systematic solver). I don't expect anybody to search for it in the repair API. We could provide a base plan as an optional parameter in the optimal planning API. Again, the solver would be perfectly allowed to ignore it.

Also one thing that would not fit in the proposed API is providing a dedicated metric (e.g. difference with previous plan).
I am unsure this something we want to do as there are no obvious choice of a metric (I have not look into the literature on this for some time but I don't recall a metric particularly standing out).
So I think the three options are:
 - no dedicated metric
 - dedicated metric are provided inside the problem class (e.g., a "plan distance" metric that contains the entire plan that we want to mimic). This makes them available to any operation mode.
 - metric is provided as an additional parameter in the repair method that supersedes the metric field in the problem.

Do you have something in mind for a plan-repair metric ? I certainly can go in a later PR but we need to make sure that we don't paint ourselves into a corner.

unified_planning/shortcuts.py
```diff
@@ -636,6 +636,30 @@ def Replanner(
...
+) -> "up.engines.engine.Engine":
+    """
+    Returns a plan repairer. There are two ways to call this method:
+    - using 'name' (the name of a plan repairer) and eventually 'params'.
+      e.g. PlanRepairer(name='xxx')
+    - using 'problem_kind' and 'optimality_guarantee'.
+      e.g. PlanRepairer(problem_kind=problem.kind, optimality_guarantee=SOLVED_OPTIMALLY)
+    """
+    if isinstance(optimality_guarantee, str):
+        optimality_guarantee = OptimalityGuarantee[optimality_guarantee]
```
Framba-Luca, line 654: This is already done in the Factory, doing it also here might be redundant

unified_planning/engines/factory.py
```diff
@@ -838,6 +844,31 @@ def Replanner(
...
+    ) -> "up.engines.engine.Engine":
+        """
+        Returns a plan repairer. There are two ways to call this method:
+        - using 'name' (the name of a plan repairer) and eventually 'params'.
+          e.g. PlanRepairer(name='xxx')
+        - using 'problem_kind' and 'optimality_guarantee'.
+          e.g. PlanRepairer(problem_kind=problem.kind, optimality_guarantee=SOLVED_OPTIMALLY)
+        """
+        if isinstance(optimality_guarantee, str):
+            optimality_guarantee = OptimalityGuarantee[optimality_guarantee]
```
Framba-Luca, line 863: I am thinking about what happens when a user writes something that is not a valid `OptimalityGuaranteee`; I guess he gets a `KeyError`. 

To improve the user experience, we might want to encapsulate this code in a try-except block that catches the `KeyError` and re-raises a `UPUsageError` stating that the given string is not a valid `OptimalityGuarantee` and contains the possible `OptimalityGuarantees`.

This is a general comment, not necessarily to do in this PR, as it should be done also in other Operation Modes; but it's such a small change that can also be done here. As you prefer!

Framba-Luca (approved):
Good job, LGTM!

I only left 2 minor comments.

unified_planning/engines/factory.py
alvalentini: I think it is better to handle it in a different PR

alvalentini:
@arbimo we decided to go with the first option (no dedicated metric), at least for the moment.

PR #308, opened 2023-01-11
all properties called env are now environment

Framba-Luca:


mikand:
@Framba-Luca Apart for rebasing, what is the impact of this on the exising engines? Did you analyze it?

Framba-Luca:
> @Framba-Luca Apart for rebasing, what is the impact of this on the exising engines? Did you analyze it?

Yes, so, PDDL Solvers are not impacted by this, while in `up-tamer` and `up-pyperplan` there was 1 line to change, where the env property was called. 

I expect that this change needs to be done on almost every existing engine that interfaces directly with the UP data structure but I don't expect any other impact besides than that.

alvalentini:
@aiplan4eu/engine-developers this PR may have a small impact on engine integrations.

e.g. to get the environment from a problem:
- old code: `problem.env`
- new code: `problem.environment`

PR #314, opened 2023-01-16
use Fast Downward integration 0.1.1, supporting grounding

roeger:


unified_planning/engines/factory.py
```diff
@@ -73,6 +73,11 @@
         "unified_planning.engines.compilers.tarski_grounder",
         "TarskiGrounder",
     ),
+    "fast-downward-grounder": ("up_fast_downward", "FastDownwardGrounder"),
+    "fast-downward-reachability-grounder": (
```
alvalentini, line 77: They should be added also in the `DEFAULT_ENGINES_PREFERENCE_LIST`

alvalentini (changes requested):

unified_planning/engines/factory.py
roeger: I added them after the tarski grounder (without any insight which one is better).  I am sure that it is in general a better choice than the up_grounder.

alvalentini (approved):

PR #318, opened 2023-01-19
Update tamer version

alvalentini:


Framba-Luca (approved):
Good work! 

I was thinking about a test where you explicitly use 2 heuristics, one good and one bad, to show that actually the good one works better than the bad one.

This would prove that the given heuristic is actually used (for what concerns the test, the heuristic might be simply ignored); however, the only way I can find to measure this is the time elapsed, which might be unpredictable. 

Consider this and then decide how to proceed, LGTM anyway! :smile:

PR #320, opened 2023-01-23
Proto: write and read timing delay

Shi-Raida:


unified_planning/grpc/proto_writer.py
```diff
@@ -175,12 +175,15 @@ def walk_timing_exp(
             atom=proto.Atom(symbol=fn),
             kind=proto.ExpressionKind.Value("FUNCTION_SYMBOL"),
         )
+        dl_exp = proto.Expression(
+            atom=proto.Atom(int=timing.delay),
```
arbimo, line 179: `timing.delay` could be a fraction as well.

unified_planning/test/test_protobuf_io.py
```diff
@@ -276,6 +276,13 @@ def test_validation_result(self):
 
                 self.assertEqual(validation_result, validation_result_up)
 
+    def test_temporal_hierarchical_goal(self):
```
arbimo, line 279: Great to have a complete example for mixed hierarchical and temporal.

It might be good to have a `test_timing` function that only test the different kind of timing expressions (start/end/global-start-global-end + 0/int/fraction).

arbimo (changes requested):
Thanks for the PR. I have let two comments regarding minor things.

Currently, the PR essentially encode timings as follow:
 - `StartTiming(3, move)` as  `(up:start 3 move)` 
 - `StartTiming(3)` as `(up:start 3)`
 - `StartTiming(0, move)` as `(up:start 0 move)`

Note that this translation only concerns timings that appear in expressions (e.g. ordering constraints in HTNs). Timings appearing on conditions/effects have a dedicated object.

Given that these timing appear in arbitrary expression, on which we will eventually allow additions (with constants or variables) the current encoding is probably sub-optimal. The following is probably more future proof:

 - `StartTiming(3, move)` as  `(up:add (up:start move) 3)` 
 - `StartTiming(3)` as `(up:add (up:start) 3)`
 - `StartTiming(0, move)` as `(up:start move)`

arbimo (approved):
Thanks for the changes, I have made some minor improvements, rebased and updated the CI scripts to avoid failing when uploading code coverage (which only work on the main repository).

arbimo:
@alvalentini Could you have a look at this PR. The changes look fine to me, fixing a limitation of the protobuf encoding that did not support delays on timepoints in constraint expressions.

arbimo:
No that's not it, I added this as workaround so that the tests (that do run on Shi-Raida's  repository) do not fail there.
So the tests are running, but are not directly visible to us : https://github.com/Shi-Raida/unified-planning/actions/runs/4065614438
They also do not have access to our environment variables (hence the fail on codecoverage that required a secret of our own).

mikand (approved):

PR #324, opened 2023-01-27
MulitAgent-io  

Alee08:
This branch is clean restart based on the code in https://github.com/aiplan4eu/unified-planning/pull/266.


Framba-Luca:
Hi @Alee08 , I modified 2 things:

- Remove the MAProblem field of the Agent because it was creating a circular dependency. (And it was the reason the tests were failing)
- Removed fmap from the default engines in the factory because up-fmap is not yet on pypi; when up-fmap will be on pypi another PR can be done that adds it to the default engines.

Let me know if you see any problems with these changes!

Alee08:
> Hi @Alee08 , I modified 2 things:
> 
> * Remove the MAProblem field of the Agent because it was creating a circular dependency. (And it was the reason the tests were failing)
> * Removed fmap from the default engines in the factory because up-fmap is not yet on pypi; when up-fmap will be on pypi another PR can be done that adds it to the default engines.
> 
> Let me know if you see any problems with these changes!

Hi @Framba-Luca, thank you, I don't have problems with these changes.

However, I added the public fluents in Agent and now obtain same problem.
I would like every agent to have their `public fluents` and `private fluents`. I added a new `add_public_fluent()` method in Agent to not modify the `FluentsSetMixin` class by adding a boolean attribute "example: `private = True/False`" to the `add_fluent() `method.

I have to investigate why the tests fail. Probably the new `public fluents` methods in Agent conflict with the `FluentsSetMixin` methods.
If you have an answer, I would be grateful.

Framba-Luca:
I think the problem is that you added the `self._ma_problem` field to the Agent class again; try to remove that and see if it works!

Alee08:
> I think the problem is that you added the `self._ma_problem` field to the Agent class again; try to remove that and see if it works!

Yes you are right, thank you! I forgot to remove it locally!

unified_planning/model/multi_agent/agent.py
```diff
@@ -71,13 +75,102 @@ def env(self) -> "up.Environment":
...
+        :param fluents: The `list` of `public fluents` that must be added to the `problem`.
+        """
+        for fluent in fluents:
+            self.add_public_fluent(fluent)
+
+    @property
+    def public_fluents(self) -> List["up.model.fluent.Fluent"]:
+        """Returns the `fluents` currently in the `problem`."""
+        return self._public_fluents
+
```
alvalentini, line 162: @Alee08 To avoid code duplication I propose to update this code in this way:

```suggestion
    def add_public_fluent(
        self,
        fluent_or_name: Union["up.model.fluent.Fluent", str],
        typename: Optional["up.model.types.Type"] = None,
        *,
        default_initial_value: Optional["ConstantExpression"] = None,
        **kwargs: "up.model.types.Type",
    ) -> "up.model.fluent.Fluent":
        """Adds the given `public fluent` to the `problem`.

        If the first parameter is not a `Fluent`, the parameters will be passed to the `Fluent` constructor to create it.

        :param fluent_or_name: `Fluent` instance or `name` of the `fluent` to be constructed.
        :param typename: If only the `name` of the `fluent` is given, this is the `fluent's type` (passed to the `Fluent` constructor).
        :param default_initial_value: If provided, defines the default value taken in initial state by
                                      a state variable of this `fluent` that has no explicit value.
        :param kwargs: If only the `name` of the `fluent` is given, these are the `fluent's parameters` (passed to the `Fluent` constructor).
        :return: The `fluent` passed or constructed.
        """
        fluent = self.add_fluent(fluent_or_name, typename, default_initial_value=default_initial_value, **kwargs)
        self._public_fluents.append(fluent)
        return fluent

    def add_private_fluent(
        self,
        fluent_or_name: Union["up.model.fluent.Fluent", str],
        typename: Optional["up.model.types.Type"] = None,
        *,
        default_initial_value: Optional["ConstantExpression"] = None,
        **kwargs: "up.model.types.Type",
    ) -> "up.model.fluent.Fluent":
        """Adds the given `private fluent` to the `problem`.

        If the first parameter is not a `Fluent`, the parameters will be passed to the `Fluent` constructor to create it.

        :param fluent_or_name: `Fluent` instance or `name` of the `fluent` to be constructed.
        :param typename: If only the `name` of the `fluent` is given, this is the `fluent's type` (passed to the `Fluent` constructor).
        :param default_initial_value: If provided, defines the default value taken in initial state by
                                      a state variable of this `fluent` that has no explicit value.
        :param kwargs: If only the `name` of the `fluent` is given, these are the `fluent's parameters` (passed to the `Fluent` constructor).
        :return: The `fluent` passed or constructed.
        """
        return self.add_fluent(fluent_or_name, typename, default_initial_value=default_initial_value, **kwargs)
        
    def add_public_fluents(self, fluents: List["up.model.fluent.Fluent"]):
        """
        Adds the given `list` of `public fluents` to the `problem`.

        :param fluents: The `list` of `public fluents` that must be added to the `problem`.
        """
        for fluent in fluents:
            self.add_public_fluent(fluent)
        
    def add_private_fluents(self, fluents: List["up.model.fluent.Fluent"]):
        """
        Adds the given `list` of `private fluents` to the `problem`.

        :param fluents: The `list` of `private fluents` that must be added to the `problem`.
        """
        for fluent in fluents:
            self.add_private_fluent(fluent)

    @property
    def public_fluents(self) -> List["up.model.fluent.Fluent"]:
        """Returns the `fluents` currently in the `problem`."""
        return self._public_fluents

    @property
    def private_fluents(self) -> List["up.model.fluent.Fluent"]:
        """Returns the `fluents` currently in the `problem`."""
        return [f for f in self._fluents if f not in self._public_fluents]
```

Alee08: @alvalentini I think it is perfect. Thank you!

PR #332, opened 2023-02-13
Add FMAP

Alee08:


.github/workflows/main.yml
```diff
@@ -224,6 +230,9 @@ jobs:
           cd ENHSP-Public; git checkout enhsp20-0.9.5; bash ./compile; cd ..
           bash -c "mkdir .planners; mv ENHSP-Public/enhsp-dist .planners/enhsp-20; rm -rf ENHSP-Public"
 
+      - name: Download and install FMAP
+        run: python3 -m pip install up-fmap
+
```
alvalentini, line 235: The changes in this file are not needed since FMAP is not used in the tests.

setup.py
```diff
@@ -30,13 +30,15 @@
...
         "lpg": ["up-lpg==0.0.4.9"],
+        "fmap": ["up-fmap==0.0.2"],
         "engines": [
             "tarski[arithmetic]",
             "up-pyperplan==0.3.0.2.dev1",
             "up-tamer==0.3.1.3.dev1",
             "up-enhsp==0.0.9",
             "up-fast-downward==0.1.2",
             "up-lpg==0.0.4.9",
+            "fmap==up-fmap==0.0.2",
```
alvalentini, line 41:
```suggestion
            "up-fmap==0.0.2",
```

alvalentini (changes requested):

.github/workflows/main.yml
Alee08: Ok, I've removed this for now! Do you suggest adding a test for FMAP?

setup.py
Alee08: Done!

PR #333, opened 2023-02-14
[htn] Provide more detailed HTN kinds

arbimo:
This PR primarily adds problem features for HTN (total-order, partial-order, constraints, variables).

It also improves the API for dealing with task ordering in task networks (allowing to get a view of the constraint as a total/partial order when possible).

Side effects:
 - impl +/- operators for `Timing` and `Timepoint`
 - add a new expression walker that can check whether any of subexpressions matches a predicate (used to find the presence of timepoints in constraint expressions)
 - provide `AbstractTaskNetwork` that contains common methods for `(Intitial)TaskNetwork` and `Method`
 - regroup HTN tests into a single file (they were previously scattered in several places)


alvalentini (approved):
@arbimo LGTM!

PR #334, opened 2023-02-15
GRPC Server Definition

franklinselva:


franklinselva:
@arbimo Do you think that the changes here are still required/valid?

alvalentini (approved):
LGTM! Mark it as ready when you want to merge it in master.

arbimo:
@franklinselva No, this is now outdated (e.g. missing engine interfaces) and not directly needed for Aries.
Aries come with its [own implementation](https://github.com/plaans/aries/blob/8b9d290da5cec861bc3a08d75e33cbf2d37b67df/planning/unified/plugin/up_aries/solver.py) of the gRPC Server that allow for a tighter integration (e.g. one solver process per problem).

We might need something like this in the future but I am closing this for now.

PR #337, opened 2023-02-22
Implementing the add-after-delete semantic

Framba-Luca:
PR linked to the discussion #191 

Before merging this PR, the engines that don't implement the **add-after-delete** semantic (i know about tamer and aries) should never return **UNSOLVABLE_PROVEN** but **UNSOLVABLE_INCOMPLETELY**. 

Maybe, a **conservative** check can be done to see if the **add-after-delete** semantic **never impacts the problem**. 
In this case, the engine **might** return **UNSOLVABLE_PROVEN**.

### DIFF
--- a/unified_planning/model/action.py
+++ b/unified_planning/model/action.py
@@ -444,6 +444,7 @@ class InstantaneousAction(Action):
         up.model.effect.check_conflicting_simulated_effects(
             simulated_effect,
             None,
+            self._fluents_assigned,
             self._fluents_inc_dec,
             "action",
         )
--- a/unified_planning/model/action.py
+++ b/unified_planning/model/action.py
@@ -28,7 +28,7 @@ from unified_planning.exceptions import (
     UPConflictingEffectsException,
 )
 from fractions import Fraction
-from typing import List, Set, Union, Optional
+from typing import Dict, List, Set, Union, Optional
 from collections import OrderedDict
 
 from unified_planning.model.mixins.timed_conds_effs import TimedCondsEffs
--- a/unified_planning/model/action.py
+++ b/unified_planning/model/action.py
@@ -26,6 +26,7 @@ from unified_planning.exceptions import (
     UPUnboundedVariablesError,
     UPProblemDefinitionError,
     UPConflictingEffectsException,
+    UPUsageError,
 )
 from fractions import Fraction
 from typing import Dict, List, Set, Union, Optional
@@ -318,7 +319,10 @@ class InstantaneousAction(Action):
             value_exp,
             condition_exp,
         ) = self._environment.expression_manager.auto_promote(fluent, value, condition)
-        assert fluent_exp.is_fluent_exp()
+        if not fluent_exp.is_fluent_exp():
+            raise UPUsageError(
+                "fluent field of add_effect must be a Fluent or a FluentExp"
+            )
         if not self._environment.type_checker.get_type(condition_exp).is_bool_type():
             raise UPTypeError("Effect condition is not a Boolean condition!")
         if not fluent_exp.type.is_compatible(value_exp.type):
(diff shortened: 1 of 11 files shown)


unified_planning/model/action.py
```diff
@@ -402,16 +418,38 @@ def _add_effect_instance(self, effect: "up.model.effect.Effect"):
...
+                        )
+                        or (
+                            assigned_value.is_real_constant()
+                            and effect.value.is_real_constant()
+                        )
+                        or (
+                            assigned_value.is_object_exp()
+                            and effect.value.is_object_exp()
+                        )
+                    )
```
alvalentini, line 440:
```suggestion
                    and assigned_value.is_constant()
                    and effect.value.is_constant()
```

unified_planning/model/action.py
```diff
@@ -402,16 +418,38 @@ def _add_effect_instance(self, effect: "up.model.effect.Effect"):
...
                 ):
                     raise UPConflictingEffectsException(
                         f"The effect {effect} is in conflict with the effects already in the action."
                     )
-                self._fluents_assigned.add(effect.fluent)
+                if (
+                    assigned_value is None
+                    or effect.value.is_int_constant()
+                    or effect.value.is_real_constant()
+                    or effect.value.is_object_exp()
```
alvalentini, line 450:
```suggestion
                    or effect.value.is_constant()
```

unified_planning/model/mixins/timed_conds_effs.py
```diff
@@ -301,31 +319,56 @@ def _add_effect_instance(
...
+                        )
+                        or (
+                            assigned_value.is_real_constant()
+                            and effect.value.is_real_constant()
+                        )
+                        or (
+                            assigned_value.is_object_exp()
+                            and effect.value.is_object_exp()
+                        )
+                    )
```
alvalentini, line 343: Same

unified_planning/model/mixins/timed_conds_effs.py
```diff
@@ -301,31 +319,56 @@ def _add_effect_instance(
...
                 ):
                     raise UPConflictingEffectsException(
                         f"The effect {effect} at timing {timing} is in conflict with the effects already in the action."
                     )
-                fluents_assigned.add(effect.fluent)
+                if (
+                    assigned_value is None
+                    or effect.value.is_int_constant()
+                    or effect.value.is_real_constant()
+                    or effect.value.is_object_exp()
```
alvalentini, line 353: Same

Framba-Luca:
@aiplan4eu/engine-developers 
Hi everyone, after this PR will be merged in master the UP will implement the `add-after-delete` semantic.

If your engine does not support it, the `PlanGenerationResult` returned -or anything returned by the engine to the user- might need some modifications; for example if the status is `UNSOLVABLE_PROVEN`, a valid plan might exist when enforcing the `add-after-delete` semantic, therefore it must be changed to `UNSOLVABLE_INCOMPLETELY`.

Should you have any question or issue just let me know by commenting here or via email!

unified_planning/model/action.py
```diff
@@ -402,16 +419,28 @@ def _add_effect_instance(self, effect: "up.model.effect.Effect"):
...
+                        assigned_value.is_constant()
+                        and effect.value.is_constant()
+                        and assigned_value.constant_value()
+                        == effect.value.constant_value()
+                    ):
+                        raise UPConflictingEffectsException(
+                            f"The effect {effect} is in conflict with the effects already in the action."
+                        )
+                else:
+                    self._fluents_assigned[effect.fluent] = effect.value
```
alvalentini, line 443: @Framba-Luca This piece of code is duplicated in 3 different files. Can we refactor it?

PR #338, opened 2023-02-22
fix: Remove the possibility of specifying negative delays from end timings.

arbimo:
`EndTiming(4)` is now disallowed and should be written `EndTiming() - 4`.

The implementation was not according to documentation (delay was reversed).
All usage and implementations that relied on it were wrong.

I removed the culprit parameter as it was inconsistent with other Timing delays.


arbimo:
@alvalentini This is ready for review and I think we agreed in a previous meeting that this was the way to go.  Note that this break backward compatibility but was very sparsely used in the UP and probably only impacting aries and tamer.

@mikand suggested to also extend this to `StartTiming` but I would suggest having this in another PR since this one is primarily about fixing the incorrect implementation. In the same line, we could stop using `Timing` and use `Timepoint`s instead as they are also converted to a Timing when additioned with a delay.

@Shi-Raida Once this gets in, the the plan of `timed-connected-location` test case should be valid.

alvalentini:
@arbimo I added a new problem feature to represent out-of-action conditions and effects. If you agree with the changes I can proceed with the merge.

arbimo:
@alvalentini LGTM. Thanks for adding the new feature, I forgot about this part of the discussion :/

PR #339, opened 2023-02-23
Support for hierarchical plans 

arbimo:
This creates a `HierarchicalPlan` class that contains:
 - a non-hierarchical plan (either `TimeTriggeredPlan` or `SequentialPlan`)
 - decomposition information that maps an initial task network down to primitives actions


arbimo:
@alvalentini This is ready for review now. We have used internally for some time without any problem.

We intend to submit a plan validator that leverages it quickly. It might also be useful for the integration of Siadex from the open calls.

alvalentini (approved):
@arbimo LGTM!

PR #343, opened 2023-02-28
fix(typing): Fix incorrect type assignment check.

arbimo:
Type checking of hierarchical types was reversed.

The type checking query was also reversed in the particular case of fluent-parameter checking.


Framba-Luca (approved):
Nice catch on this and thanks for the fix!

PR #344, opened 2023-02-28
feat(model): Allow treating action arguments as attributes.

arbimo:
This allows using subscript syntax to access parameters.

In a nutshell, it avoids the need for all the `action.parameter("...")` calls to enable the following syntax:

```py
move = InstantaneousAction("move", truck=Truck, src=Loc, tgt=Loc)
move.add_precondition(Equals(city(move.src), city(move.tgt)))
move.add_precondition(Equals(at(move.truck), move.src))
move.add_effect(at(move.truck), move.tgt)
```

While I am in general in favor of quite explicit code. The access to parameters has been quite pain point in my experience.
In particular, reextracting the parameters and storing them into intermediate variables was a common source of typos either when giving the name of the parameter to extract or when selecting the intermediate python variable (python scoping rules meaning that you have many such variables in scope).

### DIFF
--- a/unified_planning/model/action.py
+++ b/unified_planning/model/action.py
@@ -110,8 +110,8 @@ class Action:
         >>> move.parameter("target")
         Location target
 
-        If a parameter's name does not conflict with an existing attribute of an action, it can also be accessed as
-        if it was an attribute of the action. For instance:
+        If a parameter's name (1) does not conflict with an existing attribute of `Action` and (2) does not start with '_'
+        it can also be accessed as if it was an attribute of the action. For instance:
 
         >>> move.source
         Location source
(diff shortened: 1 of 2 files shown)


arbimo:
@mikand @alvalentini Could you let me know what's you're feeling about this PR ? You can ignore the failing test for now

mikand:
@arbimo I love it :heart:

arbimo:
@mikand Great! Now that I've had a taste of it, I couldn't do without it :wink: 

@alvalentini I have implemented the same for `Method` and handled a corner case where pickling that was messing around with attributes. This is now ready for review/merge.

alvalentini (approved):

PR #345, opened 2023-02-28
Epsilon separation in temporal problems

Framba-Luca:
This PR aims to introduce the epsilon field in the problem, in order to define the minimum amount of time that must expire between 2 events in a proposed solution.


unified_planning/model/problem.py
```diff
@@ -300,6 +305,38 @@ def get_static_fluents(self) -> Set["up.model.fluent.Fluent"]:
...
+    @epsilon.setter
+    def epsilon(self, new_value: Optional[Union[Real, Fraction]]):
+        if new_value is None:
+            pass
+        elif new_value < 0:
+            raise UPProblemDefinitionError(
+                "A negative epsilon is not valid and has no meaning!"
+            )
+        elif not isinstance(new_value, Fraction):
+            new_value = Fraction(float(new_value))
```
arbimo, line 330: Bad idea to go through float as an intermediate representation, as some fractionals would not be representable exactly:

```
>>>> print(Fraction(float(Fraction(1,10))))
3602879701896397/36028797018963968
```

unified_planning/model/problem.py
```diff
@@ -300,6 +305,38 @@ def get_static_fluents(self) -> Set["up.model.fluent.Fluent"]:
...
+        be, for example, the start of an action, the end of an action, an intermediate
+        step of an action, a timed effect of the problem.
+
+        When None, it means that this minimum step is chosen by the Engine the Problem
+        is given to.
+        """
+        return self._epsilon
+
+    @epsilon.setter
+    def epsilon(self, new_value: Optional[Union[Real, Fraction]]):
```
arbimo, line 322: We may want to accept integers as well (just to support setting to `1`.

arbimo:
Is `Problem` really the place to implement this? I certainly would like to use it for `SchedulingProblem` (with a default of 1?) so it may go either to `AbstractProblem` or in a temporal-stuff mixin.

A temporal-stuff mixin might be a good idea regardless. In Scheduling problem notably, all timed goals and timed initial effects are group in an object inheriting from `TimedCondsEffs` to support something similar and `Problem` could use the same thing as well.

Framba-Luca:
Hi @arbimo , thanks for the review and sorry for the late replay. Yes, I think that making a mixin that contains the Problem's temporal stuff is really a good idea!  After that, both the `SchedulingProblem` and the `Problem` will inherit from it.

It might be a good idea to also add to that mixin the `add_timed_effect` etc, but I see that in the `SchedulingProblem` they have a slightly different meaning.

Do you see an easy way to include also these methods to the `TemporalStuff` mixin or you only see the epsilon as a common ground between the `Problem` and the `SchedulingProblem`?

unified_planning/plans/time_triggered_plan.py
```diff
@@ -151,3 +152,64 @@ def convert_to(
...
+
+    :param relative_time: The timing in the action.
+    :param start: the starting time of the action.
+    :param duration: The duration of the action.
+    :return: The absolute time of the given timing.
+    """
+    if relative_time.is_from_start():
+        return start + relative_time.delay
+    else:
+        return start + duration - relative_time.delay
```
alvalentini, line 215:
```suggestion
        return start + duration + relative_time.delay
```

alvalentini:
@aiplan4eu/engine-developers 

This PR is important for planners that support time, in particular LPG (@IvanSerina @stoppini @AfkValerik) and Aries (@arbimo).

For LPG (that extends `PDDLPlanner`) it is needed to implement the new method [`_get_engine_epsilon`](https://github.com/aiplan4eu/unified-planning/blob/epsilon-separation/unified_planning/engines/pddl_planner.py#L90).

For Aries you can do something similar to [this](https://github.com/aiplan4eu/up-tamer/blob/epsilon-separation/up_tamer/engine.py#L189).

PR #350, opened 2023-03-06
feat(scheduling): Initial draft of the solution to a scheduling problem.

arbimo:


arbimo:
@mikand @alvalentini This is ready for review now.

The PR has three main contributions:
 - Add a plan representation for scheduling
 - Implement kind computation for scheduling problems
 - Implement conversion of scheduling problems from/into protobuf

At this point, the scheduling API is mostly feature complete. It will require one follow up PR to polish the public API (there are some inconsistencies left), add additional tests/examples and bootstrap the documentation.

The most impactful change is with respect to kind computation that I pulled almost completely out of `Problem` to allow reuse from sibling classes. Hopefully, the code is a bit better structured now.

alvalentini (approved):

PR #354, opened 2023-03-14
PDDL case-sensitivity problems

Framba-Luca:
This PR aims to eliminate PDDL case insensitivity related problems by setting the domain and problem string to lower before the parsing.

### DIFF
--- a/unified_planning/io/pddl_reader.py
+++ b/unified_planning/io/pddl_reader.py
@@ -314,7 +314,9 @@ class PDDLReader:
     """
     Parse a `PDDL` domain file and, optionally, a `PDDL` problem file and generate the equivalent :class:`~unified_planning.model.Problem`.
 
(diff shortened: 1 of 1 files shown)


Framba-Luca:
Hi @arbimo , since PDDL is totally case-insensitive we decided just to call the `lower()` method before parsing the domain and problem strings. This removes the need of the `CaseInsensitiveToken` class. 

We are aware that the error message will be written in lower-case, even if the original text is in upper-case. This is mitigated by the fact that the user gets line and column number of the error.

Do you see any issue with that or with this PR in general?

arbimo:
Yes, I think it makes sense to do this :+1: 
Even though I am annoyed not to have the exact string in the resulting `Problem`, the complexity of handling it fully (currently, case insensitivity was only supported for types I think ?) is probably not worth it. We may want to consider it if we ever switch our parser implementation though.

PR #355, opened 2023-03-14
fix(examples): duration of move in the htn-go-temporal problem

Shi-Raida:
The duration was set to 10 in the problem and 1 in the plan.
Therefore, the plan was not valid.


arbimo (approved):
LGTM, thanks for catching this!

@alvalentini This fixes an inconsistency I introduced between the plan and the problem in this example. The resulting plan was invalid.

PR #356, opened 2023-03-15
feat(validation): Assign correct boolean semantics to `PlanValidationResult`.

arbimo:
This allows using a validation result directly as a boolean.
Or more exactly when used as a boolean (e.g. in an if condition),
VALID would be interpreted as True and INVALID as False.

Previously, the code below was perfectly legal but would print ok
even if the plan was invalid.

```py
if validator.validate(pb, plan):
  print("ok")
else:
  print("ko")
```


alvalentini (approved):
Is it ready to merge?

arbimo:
Yes, it is ready to merge.

PR #357, opened 2023-03-16
feat(simulator): Simulator interface simplification.

Framba-Luca:


unified_planning/model/expression.py
```diff
@@ -68,24 +68,22 @@ def __init__(self, environment: "up.environment.Environment"):
...
-            if isinstance(p, Iterable):
-                res.extend(list(p))
+        for a in args:
+            if isinstance(a, Iterable):
+                for p in a:
+                    yield p
             else:
-                res.append(p)
-        return tuple(res)
+                yield a
```
alvalentini, line 86: Why this change?

unified_planning/engines/sequential_simulator.py
```diff
@@ -435,47 +486,105 @@ def supported_kind() -> "up.model.ProblemKind":
...
+
+def evaluate_quality_metric(
+    simulator: SequentialSimulatorMixin,
+    quality_metric: "up.model.PlanQualityMetric",
+    metric_value: "up.model.FNode",
+    state: "up.model.State",
+    action: "up.model.Action",
+    parameters: Tuple["up.model.FNode", ...],
+    next_state: "up.model.State",
+) -> "up.model.FNode":
```
alvalentini, line 500: I would return a `Fraction` or an `int`

PR #363, opened 2023-03-22
Fix MinimizeActionCosts metric on compilers

Framba-Luca:


unified_planning/engines/compilers/disjunctive_conditions_remover.py
```diff
@@ -209,7 +210,14 @@ def _compile(
             new_problem.add_timed_goal(t, goal_to_add)
 
         for qm in problem.quality_metrics:
-            if isinstance(qm, Oversubscription):
+            if isinstance(qm, MinimizeActionCosts):
+                new_costs = {
+                    new_act: qm.get_action_cost(old_act)
+                    for new_act, old_act in new_to_old.items()
+                    if old_act is not None
```
alvalentini, line 217: You have to specify a cost for every Action, or to specify a default.

alvalentini (changes requested):

PR #365, opened 2023-03-23
New ProblemKind's flags

### LINKED ISSUE #323: PDDL output error when right side of a boolean assignment is not a constant
When dumping a problem to PDDL with the PDDLWriter, an effect on a boolean fluent with a value that is not a boolean constant, is translated with an assign; this generates wrong PDDL syntax.

Framba-Luca:
This PR adds the effects kind that discriminates between constant and non constant assignments, and fixes the PDDLWriter when writing non-constant boolean assignments

Closes #323


unified_planning/model/problem.py
```diff
@@ -280,36 +280,93 @@ def _replace_action_instance(
...
+                for c in a.preconditions:
+                    remove_used_fluents(c)
                 for e in a.effects:
+                    remove_used_fluents(e.fluent)
+                    remove_used_fluents(e.value)
+                    remove_used_fluents(e.condition)
                     static_fluents.discard(e.fluent.fluent())
                 if a.simulated_effect is not None:
+                    # empty the set because a simulated effect reads all the fluents
+                    unused_fluents.difference_update(unused_fluents)
```
alvalentini, line 310:
```suggestion
                    unused_fluents.clear()
```

unified_planning/model/problem.py
```diff
@@ -280,36 +280,93 @@ def _replace_action_instance(
...
                     for f in a.simulated_effect.fluents:
                         static_fluents.discard(f.fluent())
             elif isinstance(a, up.model.action.DurativeAction):
+                remove_used_fluents(a.duration.lower)
+                remove_used_fluents(a.duration.upper)
```
alvalentini, line 315:
```suggestion
```

unified_planning/model/problem.py
```diff
@@ -280,36 +280,93 @@ def _replace_action_instance(
...
+            (f.fluent() for f in fve.get(exp))
+        )
         for a in self._actions:
             if isinstance(a, up.model.action.InstantaneousAction):
+                for c in a.preconditions:
+                    remove_used_fluents(c)
                 for e in a.effects:
+                    remove_used_fluents(e.fluent)
+                    remove_used_fluents(e.value)
+                    remove_used_fluents(e.condition)
```
alvalentini, line 306:
```suggestion
                    remove_used_fluents(e.fluent, e.value, e.condition)
```

Same below

alvalentini (changes requested):

unified_planning/model/problem.py
Framba-Luca: Ah yes, I fixed the one at line 318

unified_planning/engines/pddl_planner.py
```diff
@@ -192,7 +199,9 @@ def _solve(
         output_stream: Optional[Union[Tuple[IO[str], IO[str]], IO[str]]] = None,
     ) -> "up.engines.results.PlanGenerationResult":
         assert isinstance(problem, up.model.Problem)
-        self._writer = PDDLWriter(problem, self._needs_requirements)
+        self._writer = PDDLWriter(
+            problem, self._needs_requirements, rewrite_bool_assignment=True
```
alvalentini, line 203:
```suggestion
            problem, self._needs_requirements, self._rewrite_bool_assignment
```

roeger:
Hi, I am not sure whether this is the right place for this comment but since this pull request should also enable the action-costs requirement for classical planners, maybe yes.

I tried to solve with the Fast Downward integration (including the changes from the pull request) a task that uses numeric fluents for determining action costs. This would already be covered by the :actions-costs requirement but the PDDLWriter also adds requirement :numeric-fluents, so Fast Downward refuses to attempt the task (on the PDDL level, not the problem kind level). Is it possible to narrow down the PDDLWriter to not include this requirement in such situations?

alvalentini:
@roeger thanks for highlight the point, you are right! Now the PDDLWriter should not add :numeric-fluents in these situations.

PR #375, opened 2023-03-31
Remove constraint on pyparsing version

alvalentini:


g-gemignani (commented):
@alvalentini thank you for the quick PR. With this change, you are reverting the code to the syntax used in pyparsing 2. Wouldn't it make sense to keep the syntax of pyparsing 3 and just add a try except for the imports that in case of failure import the pyparsing 2 modules named as the pyparsing 3? Also, I think it would make sense to change the github workflow to test with both versions of the library.

alvalentini:
@g-gemignani now the code should work fine with both pyparsing 2 and 3.

g-gemignani:
@alvalentini yep, looks good to me. Thank you so much for the quick fix!

g-gemignani (approved):
Looks good to me. However, I do not know what are the adopted policies for commit guidelines or conventions. Also, I am not an expert of the package. Thus, let's wait the review of someone else before merging. Thank you again!

PR #377, opened 2023-04-04
Enhancement PlanValidationResult with invalid plans

Framba-Luca:
With this PR we add:
- an `Enum` representing the possible reasons of a failed plan (for now `INAPPLICABLE_ACTION` or `UNSATISFIED_GOAL`)
- an `Optional` field to the `PlanValidationResult` called `reason`, that is an instance of the `Enum` described above
- an `Optional` field to the `PlanValidationResult` called `inapplicable_action`, that might contain the failed action instance when the reason of the failed plan is `INAPPLICABLE_ACTION`

### DIFF
--- a/unified_planning/engines/plan_validator.py
+++ b/unified_planning/engines/plan_validator.py
@@ -133,6 +133,7 @@ class SequentialPlanValidator(engines.engine.Engine, mixins.PlanValidatorMixin):
         if metric is not None:
             metric_value = evaluate_quality_metric_in_initial_state(simulator, metric)
         msg = None
+        last_executed_action = None
         for i, ai in enumerate(plan.actions):
             assert prev_state is not None
             try:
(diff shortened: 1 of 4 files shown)


unified_planning/engines/results.py
```diff
@@ -45,6 +45,13 @@ def __bool__(self):
             return False
 
 
+class FailedValidationReasons(Enum):
```
alvalentini, line 48:
```suggestion
class FailedValidationReason(Enum):
```

unified_planning/engines/plan_validator.py
```diff
@@ -141,14 +143,23 @@ def _validate(
                 msg = f"{str(i)}-th action instance {str(ai)} creates conflicting effects: {str(e)}"
             except UPUsageError as e:
                 msg = f"{str(i)}-th action instance {str(ai)} creates a UsageError: {str(e)}"
+            except UPInvalidActionError as e:
+                msg = f"{str(i)}-th action instance {str(ai)} creates an Invalid Action: {str(e)}"
             if unsat_conds:
                 msg = f"Preconditions {unsat_conds} of {str(i)}-th action instance {str(ai)} are not satisfied."
```
alvalentini, line 149: To fix

PR #379, opened 2023-04-05
Multi-agent Notebooks

Alee08:


alvalentini:
@Alee08 thanks for the notebooks. Do you think you can split some of the code cells adding some text describing what the code is doing?

Alee08:
> @Alee08 thanks for the notebooks. Do you think you can split some of the code cells adding some text describing what the code is doing?

I added some text describing what the code does. Do you think it's okay?

PR #382, opened 2023-04-07
Small qol fixes

Framba-Luca:


unified_planning/model/metrics.py
```diff
@@ -39,27 +46,69 @@ class MinimizeActionCosts(PlanQualityMetric):
...
     ):
-        self.costs = costs
-        self.default = default
+        PlanQualityMetric.__init__(self, environment)
+        em = self._env.expression_manager
+        self._costs: Dict["up.model.Action", Optional["up.model.FNode"]] = {}
+        for action, cost in costs.items():
+            cost_exp: Optional["up.model.FNode"] = None
+            if cost is not None:
+                cost_exp = em.auto_promote(cost)[0]
```
alvalentini, line 59:
```suggestion
                cost_exp, = em.auto_promote(cost)
```

unified_planning/model/metrics.py
```diff
@@ -39,27 +46,69 @@ class MinimizeActionCosts(PlanQualityMetric):
 
     def __init__(
         self,
-        costs: Dict["up.model.Action", Optional["up.model.FNode"]],
-        default: Optional["up.model.FNode"] = None,
+        costs: Dict["up.model.Action", Optional["up.model.Expression"]],
```
alvalentini, line 49: I would remove `Optional` here

```suggestion
        costs: Dict["up.model.Action", "up.model.Expression"],
```

unified_planning/model/metrics.py
```diff
@@ -39,27 +46,69 @@ class MinimizeActionCosts(PlanQualityMetric):
...
+                        f"The cost expression {cost_exp} and the metric don't have the same environment"
+                    )
+            if action.environment != self._env:
+                raise UPProblemDefinitionError(
+                    f"The action {action.name} and the metric don't have the same environment"
+                )
+            self._costs[action] = cost_exp
+        self._default: Optional["up.model.FNode"] = None
+        if default is not None:
+            default_exp: "up.model.FNode" = em.auto_promote(default)[0]
```
alvalentini, line 77: Same

unified_planning/model/metrics.py
```diff
@@ -108,8 +157,25 @@ class MinimizeExpressionOnFinalState(PlanQualityMetric):
...
 
-    def __init__(self, expression: "up.model.FNode"):
-        self.expression = expression
+    def __init__(
+        self,
+        expression: "up.model.Expression",
+        environment: Optional[Environment] = None,
+    ):
+        PlanQualityMetric.__init__(self, environment)
+        self.expression: "up.model.FNode" = self._env.expression_manager.auto_promote(
```
alvalentini, line 166: Same here and in all similar situations

unified_planning/model/metrics.py
```diff
@@ -179,14 +286,31 @@ class TemporalOversubscription(PlanQualityMetric):
...
+            if g_exp.environment != self._env:
+                raise UPProblemDefinitionError(
+                    f"goal {g_exp} does not have the same environment given to the metric."
+                )
+            goal_keys.append((interval, g_exp))
+            if gain.denominator == 1:
+                gains.append(gain.numerator)
+            else:
+                gains.append(gain)
+        self._goals = dict(zip(goal_keys, gains))
```
alvalentini, line 313: Why not directly create a dict? Same above

unified_planning/engines/compilers/grounder.py
```diff
@@ -327,7 +327,7 @@ def _compile(
         for qm in problem.quality_metrics:
             if isinstance(qm, MinimizeActionCosts):
                 simplifier = grounder_helper.simplifier
-                new_costs: Dict[Action, Optional[FNode]] = {}
+                new_costs: Dict["up.model.Action", Optional["up.model.Expression"]] = {}
```
alvalentini, line 330: Is this change due to a mypy problem? Because it should be correct also in the previous version.

alvalentini (changes requested):

unified_planning/engines/compilers/grounder.py
Framba-Luca: Yes. 
Mypy, inside the `Dict` typing, does not recognize an `FNode` as an `Expression`. It has something to do with `Dict` being an invariant type. 

Maybe if we use `Mapping` instead of `Dict` this problem should be solved, but I haven't actually tried. Otherwise we have to type it to be the exact type requested (actually, same issue is with the `Substituter`).

PR #384, opened 2023-04-12
Uniformed numeric types

Framba-Luca:


unified_planning/model/walkers/type_checker.py
```diff
@@ -230,6 +230,7 @@ def walk_plus(self, expression, args):
             assert upper is None or isinstance(upper, Fraction)
             return self.environment.type_manager.RealType(lower, upper)
         else:
+            print(lower)
```
alvalentini, line 233: Leftover

unified_planning/model/walkers/type_checker.py
```diff
@@ -284,9 +285,13 @@ def walk_times(self, expression, args):
             else:
                 lower = min(lower * l, lower * u, upper * l, upper * u)
                 upper = max(lower * l, lower * u, upper * l, upper * u)
-        if lower == -float("inf"):
+        if (
+            lower == -float("inf") or lower != lower
+        ):  # lower != lower is True only if lower is NaN
```
alvalentini, line 290: You can explicitly check that lower is NaN

alvalentini (changes requested):

PR #385, opened 2023-04-13
Implemented correct substituter shadowing

Framba-Luca:


unified_planning/test/test_substituter.py
```diff
@@ -98,6 +98,57 @@ def test_substitution(self):
             e6 = Int(1)
             s.substitute(e6, subst)
 
+    def test_shadowing(self):
```
mikand, line 101: Add a bit of comments to explain what you are testing: this will save your life in the future :)

mikand (approved):
Very good! Just add a few comments before merge

PR #386, opened 2023-04-13
fix(expression): use isintance to check the value type

Shi-Raida:
The value can be an `int` without having the type `int`, *e.g.* `IntEnum` which is a subclass of `int`.


Shi-Raida:
@arbimo can you review it please?

arbimo:
LGTM but I am not sure whether there was any reason for having an exact type check in the original version. Maybe @alvalentini can help here (you were git-blamed) ?

alvalentini (approved):
I'm fine with these changes!

PR #390, opened 2023-04-17
Added Dot condition in add_effect()

Alee08:


alvalentini:
@Alee08 can you update the error messages according to the changes and do the same change for the `add_increase_effect` and `add_decrease_effect` methods?

PR #398, opened 2023-05-08
Added self_overlapping flag to TimeModelMixin

Framba-Luca:


alvalentini:
@arbimo @IvanSerina 

In this PR we added the possibility to specify if a temporal problem allows or not self overlapping.

If your engines support self overlapping, you can add it in the supported kind.

PR #413, opened 2023-05-24
Added two planning engines: "symk" and "symk-opt"

speckdavid:
As part of the 1st sprint of the "Symbolic Search for Diverse Plans and Maximum Utility" project, we added two search engines to the UP that can be installed using the [up-symk](https://github.com/aiplan4eu/up-symk/tree/master) repo. Both engines can generate a single or multiple/all plans using the OneshotPlanner and AnytimePlanner interfaces. The two differ in that "symk-opt" has an optimality guarantee for both settings, generating a single or multiple/all optimal plans.

You can find a usage example [here](https://github.com/aiplan4eu/up-symk/blob/master/notebooks/symk_usage.ipynb).

Let me know if you want any changes.

Best
David

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,6 +32,7 @@ setup(
         "lpg": ["up-lpg==0.0.6.3"],
         "fmap": ["up-fmap==0.0.6"],
         "aries": ["up-aries>=0.0.8"],
+        "symk": ["up-symk>=0.0.3"],
         "engines": [
             "tarski[arithmetic]",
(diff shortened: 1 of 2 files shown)


speckdavid:
@alvalentini Thanks, I added `symk` to the [setup.py](https://github.com/aiplan4eu/unified-planning/pull/413/commits/4b11993aa353a426412a842787d4da70b522ab02) file.

PR #414, opened 2023-05-26
Added UNKNOW to ValidationResultStatus

Framba-Luca:
Added UNKNOWN to ValidationResultStatus, added epsilon, self_overlapping and discrete_time to grpc and added a check that in an assignment there are no fluents in the arguments of the main fluent


unified_planning/grpc/unified_planning.proto
```diff
@@ -765,6 +773,8 @@ message ValidationResult {
         VALID = 0;
         // The Plan is not valid for the Problem.
         INVALID = 1;
+        // The engine can't determine if the plan is VALID or INVALID for the Problem.
+        UNKNOWN = 2;
```
alvalentini, line 777: This is not handle in the proto reader and writer

Issue #416, opened 2023-05-29
Negative Conditions Removing not working with the pddls I use


aymeric75:
**Describe the bug**
When using the negative preconditions removing, I get this error:
```
  File "/usr/local/lib/python3.8/dist-packages/unified_planning/model/fnode.py", line 200, in fluent
    assert self.is_fluent_exp()
AssertionError
```

**To Reproduce**
Here is my code:
```
import unified_planning
from unified_planning.shortcuts import *
from unified_planning.io import PDDLReader, PDDLWriter


reader = PDDLReader()
pddl_problem = reader.parse_problem('./domain.pddl', './problem.pddl')
print(pddl_problem)

# Get the compiler from the factory

dcr_kind = pddl_problem.kind

dcr_problem = pddl_problem

with Compiler(
        problem_kind = dcr_kind,
        compilation_kind = CompilationKind.NEGATIVE_CONDITIONS_REMOVING
    ) as negative_conditions_remover:
    # After we have the compiler, we get the compilation result
    ncr_result = negative_conditions_remover.compile(
        dcr_problem,
        CompilationKind.NEGATIVE_CONDITIONS_REMOVING
    )
    ncr_problem = ncr_result.problem
    ncr_kind = ncr_problem.kind

    # Check the result of the compilation
    assert original_problem_kind.has_negative_conditions()
    assert qr_kind.has_negative_conditions()
    assert cer_kind.has_negative_conditions()
    assert dcr_kind.has_negative_conditions()
    assert not ncr_kind.has_negative_conditions()
```

Please find attache the domain.pddl and the problem.pddl

Thanks

[domain.txt](https://github.com/aiplan4eu/unified-planning/files/11591056/domain.txt)
[problem.txt](https://github.com/aiplan4eu/unified-planning/files/11591057/problem.txt)

Framba-Luca:
Hi @aymeric75 , thank you for the bug report.

From your PDDL files I see that you also initialize false predicates (for example `(:init (not (z0))`).

This is wrong PDDL, as predicates are `False` by default, and you set them `True` in the initial state by writing  `(:init (z0)`.

This caused the bug in the Compiler, (which is more a silent parsing error than a parser/compiler bug).

We are still deciding if we want to support this syntax or raise the exception in the parser; in the meantime you can simply delete the `not` init in the problem and your code should work without any issue.

aymeric75:
Indeed thanks!

PR #417, opened 2023-05-29
Added support in PDDL to fluents initialized with not

Framba-Luca:


unified_planning/io/pddl_reader.py
```diff
@@ -1535,13 +1535,14 @@ def declare_type(
...
-                        self._parse_exp(
-                            problem, None, types_map, {}, init, problem_str
-                        ),
-                        self._em.TRUE(),
+                    exp = self._parse_exp(
+                        problem, None, types_map, {}, init, problem_str
                     )
-
+                    if not exp.is_not():
+                        problem.set_initial_value(
```
alvalentini, line 1542: Add a check that exp is a fluent expression here

alvalentini (changes requested):

Framba-Luca:
solves #416

PR #422, opened 2023-06-01
open-source project management documents and tools 

mikand:
Adding missing files, documentation and procedures to safely allow for external contributions and to institute a governance mechanism.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -87,7 +87,7 @@ Once it's filed:
 
 ### Suggesting Enhancements
 
-This section guides you through submitting an enhancement suggestion for Unified Planning Library, **including completely new features and minor improvements to existing functionality**. Following these guidelines will help maintainers and the community to understand your suggestion and find related suggestions.
+This section guides you through submitting an enhancement suggestion for the Unified Planning Library, **including completely new features and minor improvements to existing functionality**. Following these guidelines will help maintainers and the community to understand your suggestion and find related suggestions.
 
 <!-- omit in toc -->
 #### Before Submitting an Enhancement
@@ -110,7 +110,7 @@ Enhancement suggestions are tracked as [GitHub issues](https://github.com/aiplan
 
 ### Code Contribution
 
-Before start coding, open an issue or a discussion on the project [GitHub space]((https://github.com/aiplan4eu/unified-planning) or get in contact with a maintainer (you can use the project mailing list: unified-planning@googlegroups.com)! It is possible that someone else is already working on the same or a related fix or feature!
+Before you start coding, open an issue or a discussion on the project [GitHub space]((https://github.com/aiplan4eu/unified-planning) or get in contact with a maintainer (you can use the project mailing list: unified-planning@googlegroups.com)! It is possible that someone else is already working on the same or a related fix or feature!
 
 Moreover, we have to keep a reasonable project scope, so if your contribution is wildly beyond the current capabilities of the library, it is better to discuss the idea with the community and the maintainers to avoid unpleasant situations where we have to reject your code.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,7 +7,7 @@ All types of contributions are encouraged and valued. See the [Table of Contents
 
 > And if you like the project, but just don't have time to contribute, that's fine. There are other easy ways to support the project and show your appreciation, which we would also be very happy about:
 > - Star the project
-> - Tweet about it
+> - Post on it on social networks
 > - Refer this project in your project's readme
 > - Mention the project at local meetups and tell your friends/colleagues
 
@@ -51,7 +51,7 @@ We will then take care of the issue as soon as possible.
 ## I Want To Contribute
 
 > ### Legal Notice <!-- omit in toc -->
-> When contributing to this project, you must agree that you have authored 100% of the content, that you have the necessary rights to the content and that the content you contribute may be provided under the project license.
+> When contributing to this project, you must agree to the [Developer Certificate of Origin (DCO)](DCO.txt): all commits must be signed-off with your real name to signal the adherence to the DCO legal statement.
 
 ### Reporting Bugs
 
@@ -74,7 +74,7 @@ A good bug report shouldn't leave others needing to chase you up for more inform
 <!-- omit in toc -->
 #### How Do I Submit a Good Bug Report?
 
-> You must never report security related issues, vulnerabilities or bugs including sensitive information to the issue tracker, or elsewhere in public. Instead sensitive bugs must be sent by email to <unified-planning-maintainers@googlegroups.com>.
+> You should never report security related issues, vulnerabilities or bugs including sensitive information to the issue tracker, or elsewhere in public. Instead sensitive bugs should be sent by email to <unified-planning-maintainers@googlegroups.com>.
 
 We use GitHub issues to track bugs and errors. If you run into an issue with the project:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,21 +48,6 @@ If you then still feel the need to ask a question and need clarification, we rec
 
 We will then take care of the issue as soon as possible.
 
-<!--
-You might want to create a separate issue tag for questions and include it in this description. People should then tag their issues accordingly.
-
-Depending on how large the project is, you may want to outsource the questioning, e.g. to Stack Overflow or Gitter. You may add additional contact and information possibilities:
-- IRC
-- Slack
-- Gitter
-- Stack Overflow tag
-- Blog
-- FAQ
-- Roadmap
-- E-Mail List
-- Forum
--->
-
 ## I Want To Contribute
 
 > ### Legal Notice <!-- omit in toc -->
@@ -90,7 +75,6 @@ A good bug report shouldn't leave others needing to chase you up for more inform
 #### How Do I Submit a Good Bug Report?
 
 > You must never report security related issues, vulnerabilities or bugs including sensitive information to the issue tracker, or elsewhere in public. Instead sensitive bugs must be sent by email to <unified-planning-maintainers@googlegroups.com>.
-<!-- You may add a PGP key to allow the messages to be sent encrypted as well. -->
 
 We use GitHub issues to track bugs and errors. If you run into an issue with the project:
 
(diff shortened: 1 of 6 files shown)


GOVERNANCE.md
```diff
@@ -0,0 +1,93 @@
...
+
+
+# Communication Channels
+
+We establish two official communication channels for the UP project.
+
+- The **Maintainers Mailing List** (unified-planning-maintainers@googlegroups.com): all and only the active Maintainers are part of this list. The mailing list is not publicly readable but anyone can send messages to it. It is the main and official communication channel among maintainers and to communicate to the board of maintainers.
+
+- The **Public Mailing List** (unified-planning@googlegroups.com): is a general-purpose communication channel to be used for both technical matters and public announcements. the list is publicly readable and anyone can join it. All decisions and votes of the Board of Maintainers are publicly communicated through this list.
+
```
roeger, line 57: Does this mean that when the board of maintainers declines a pull request, there should be an email to this mailing list? Does this make sense? These decisions are transparent because they are visible in the pull requests.

GOVERNANCE.md
```diff
@@ -0,0 +1,93 @@
...
+The following rules govern the promotion and demotion of Committers and Maintainers.
+
+- Any Maintainer can autonomously nominate (and give the needed permissions) any number of Committers. (The rationale is that being a Committer does not grant any specific power over the project, but simplifies the work for people who often contribute to the library).
+
+- Any Maintainer can step down (effective immediately) from being a Maintainer by informing the Board of Maintainers.
+
+- Only a majority vote of the Board of Maintainers can nominate new Maintainers. The number rof Maintainers is not fixed.
+
+- Only a full consensus vote (meaning that all the Maintainers agree, with no abstentions) of the Board of Maintainers can modify the governance rules layed out in this document.
```
roeger, line 88: This is already mentioned in the technical rules, but I think it still makes sense to repeat it here because it is not only a technical issue but this explicit statement makes clear that the board of all maintainers has the right to change the governance rules.

CONTRIBUTING.md
```diff
@@ -0,0 +1,139 @@
+<!-- omit in toc -->
+# Contributing to Unified Planning Library
+
+First off, thanks for taking the time to contribute! ❤️
+
+All types of contributions are encouraged and valued. See the [Table of Contents](#table-of-contents) for different ways to help and details about how this project handles them. Please make sure to read the relevant section before making your contribution. It will make it a lot easier for us maintainers and smooth out the experience for all involved. The community looks forward to your contributions. 🎉
+
+> And if you like the project, but just don't have time to contribute, that's fine. There are other easy ways to support the project and show your appreciation, which we would also be very happy about:
+> - Star the project
+> - Tweet about it
```
roeger, line 10: Do we want to refer to twitter here? Not sure whether I would like to support it. What about "post on it on social networks"?

CONTRIBUTING.md
```diff
@@ -0,0 +1,139 @@
...
+- Open an [Issue](https://github.com/aiplan4eu/unified-planning/issues/new).
+- Provide as much context as you can about what you're running into.
+- Provide project and platform versions (nodejs, npm, etc), depending on what seems relevant.
+
+We will then take care of the issue as soon as possible.
+
+## I Want To Contribute
+
+> ### Legal Notice <!-- omit in toc -->
+> When contributing to this project, you must agree that you have authored 100% of the content, that you have the necessary rights to the content and that the content you contribute may be provided under the project license.
```
roeger, line 54: Is this really meant as a conjunction? Then no-one may use existing material whose license would allow an integration.

CONTRIBUTING.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Reporting Bugs
+
+<!-- omit in toc -->
+#### Before Submitting a Bug Report
+
+A good bug report shouldn't leave others needing to chase you up for more information. Therefore, we ask you to investigate carefully, collect information and describe the issue in detail in your report. Please complete the following steps in advance to help us fix any potential bug as fast as possible.
+
+- Make sure that you are using the latest version.
+- Determine if your bug is really a bug and not an error on your side e.g. using incompatible environment components/versions (Make sure that you have read the [documentation](https://unified-planning.readthedocs.io/en/latest/). If you are looking for support, you might want to check [this section](#i-have-a-question)).
```
roeger, line 64: In that section we write that the way to get support is to open an issue. What's the exact line between a bug report and an issue?

CONTRIBUTING.md
```diff
@@ -0,0 +1,139 @@
...
+  - Stack trace (Traceback)
+  - OS, Platform and Version (Windows, Linux, macOS, x86, ARM)
+  - Version of the interpreter, compiler, SDK, runtime environment, package manager, depending on what seems relevant.
+  - Possibly your input and the output
+  - Can you reliably reproduce the issue? And can you also reproduce it with older versions?
+
+<!-- omit in toc -->
+#### How Do I Submit a Good Bug Report?
+
+> You must never report security related issues, vulnerabilities or bugs including sensitive information to the issue tracker, or elsewhere in public. Instead sensitive bugs must be sent by email to <unified-planning-maintainers@googlegroups.com>.
```
roeger, line 77: This "must never" and "must be sent" sounds very harsh and strict (and we cannot really forbid it in practice, especially regarding reporting in public). I would go for "should" instead.

CONTRIBUTING.md
roeger: In the DCO it's a discjunction.

roeger (commented):
Overall the documents look good to me. I only left some small comments.

CONTRIBUTING.md
mikand: I am not sure how to address this, any suggestions?

mikand:
@roeger @arbimo @alvalentini @hstairs As future maintainers, please have a careful look at the docs. Any feedback/edit is very welcome.

CONTRIBUTING.md
```diff
@@ -0,0 +1,140 @@
...
+
+> If you want to ask a question, we assume that you have read the available [Documentation](https://unified-planning.readthedocs.io/en/latest/).
+
+Before you ask a question, it is best to search for existing [Issues](https://github.com/aiplan4eu/unified-planning/issues) that might help you. In case you have found a suitable issue and still need clarification, you can write your question in this issue. It is also advisable to search the internet for answers first.
+
+If you then still feel the need to ask a question and need clarification, we recommend the following:
+
+- Open an [Issue](https://github.com/aiplan4eu/unified-planning/issues/new).
+- Provide as much context as you can about what you're running into.
+- Provide project and platform versions (nodejs, npm, etc), depending on what seems relevant.
```
hstairs, line 47: Minor, but why do we have nodejs, npm as examples here?

GOVERNANCE.md
hstairs: I can't see where this is implied, sorry!:(

GOVERNANCE.md
```diff
@@ -0,0 +1,94 @@
...
+
+9.  Any Maintainer can autonomously nominate (and give the needed permissions) any number of Committers. (The rationale is that being a Committer does not grant any specific power over the project, but simplifies the work for people who often contribute to the library).
+
+10. Any Maintainer can step down (effective immediately) from being a Maintainer by informing the Board of Maintainers.
+
+11. Only a majority vote of the Board of Maintainers can nominate new Maintainers. The number of Maintainers is not fixed.
+
+12. Only a full consensus vote (meaning that all the Maintainers agree, with no abstentions) of the Board of Maintainers can modify the governance rules layed out in this document.
+
+13. In order to remove an Maintainer unwilling to leave his/her position, a 75% majority vote of the other Maintainers is needed.
```
hstairs, line 94: I think it is "at least a 75%"

GOVERNANCE.md
```diff
@@ -0,0 +1,94 @@
...
+The **Board of Maintainers** is the main decision body of the UP project and is composed of all the current Maintainers.
+
+
+# Communication Channels
+
+We establish two official communication channels for the UP project in addition to the GitHub facilities (issues, discussions, PRs).
+
+- The **Maintainers Mailing List** (unified-planning-maintainers@googlegroups.com): all and only the active Maintainers are part of this list. The mailing list is not publicly readable but anyone can send messages to it. It is the main and official communication channel among maintainers and to communicate to the board of maintainers.
+
+- The **Public Mailing List** (unified-planning@googlegroups.com): is a general-purpose communication channel to be used for both technical matters and public announcements. The list is publicly readable and anyone can join it. The outcome of all the votes of the Board of Maintainers (required for the situations indicated in the following) are publicly communicated through this list.
```
hstairs, line 56: And also anyone can send message to it? If anyone can send messages, as I suppose it is, are messages supposed to be accepted/rejected by someone when received before getting public?

GOVERNANCE.md
```diff
@@ -0,0 +1,94 @@
...
+
+The following rules apply for the technical day-to-day governance of the project.
+
+1. Any Maintainer can approve and commit to "master" a Pull-Request from any Committer, External Contribution or **another** Maintainer.
+
+2. It is forbidden to anyone to self-approve and commit a Pull-Request.
+
+3. Any Maintainer can reject a Pull-Request from any Committer or External Contribution or Maintainer by publicly motivating the reason in the PR discussion page. Reasons should be of technical nature or concerning the scope of the library and any effort should be made to not waste anybody's work in a constructive manner.
+
+4. If a Maintainer disagrees with another Maintainer decision of accepting or rejecting a Pull-Request, he/she can uphold the final decision to a simple majority vote of the  Board of Maintainers.
```
alvalentini, line 71: Does this mean that an accepted Pull-Request can be reverted?

GOVERNANCE.md
mikand: Good point. I was thinking on a list where anyone can subscribe and post without moderation (except spam), but then we need a ban mechanism for trolls... We can discuss this at the enxt meeting

GOVERNANCE.md
mikand: I think that in the (hopefully rare) ocasions where a formal vote is needed because of disagreement, it does not hurt to publish the vote result also on the mailing list. In all the other cases we can rely on the comment messages on the PRs (where I expect a technical discussion, even between maintainers)

GOVERNANCE.md
mikand: Good point, added to discussion points

roeger (approved):
I left some comments and approve merging this pull request (also without addressing the comments).

PR #424, opened 2023-06-06
Fix of PDDL Anytime Planner refactoring

Framba-Luca:
Added a method to modify the last result returned by the PDDLAnytimePlanner

### DIFF
--- a/unified_planning/engines/pddl_anytime_planner.py
+++ b/unified_planning/engines/pddl_anytime_planner.py
@@ -174,8 +174,7 @@ class PDDLAnytimePlanner(engines.pddl_planner.PDDLPlanner, mixins.AnytimePlanner
     def _generate_last_result(
         self,
         solve_result: PlanGenerationResult,
-        last_plan_found: Optional[Plan],
-        last_status: PlanGenerationResultStatus,
+        last_result: Optional[PlanGenerationResult],
     ) -> PlanGenerationResult:
         """
         IMPORTANT FOR ENGINES IMPLEMENTING THIS CLASS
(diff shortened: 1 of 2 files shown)


Framba-Luca:
Hi @roeger , I looked into your issue a little bit more in depth. The main difference between your integration and the ENHSP one seems to be that the fast-downward anytime does not write the final plan to a file (correct me if I am wrong).

This causes the `_solve` method to return an `UNSOLVABLE_INCOMPLETELY` status and a `None` plan (tested on the `miconic` problem inside the `unified_planning/test/pddl` folder).

So, the idea of this PR, is to give an easy way to modify the `_solve` result with the `_generate_last_result` method. 

The name of the method will probably change, as there are for sure better options :)

Let me know what is your feedback on this solution!

roeger:
@Framba-Luca Fast Downward follows the IPC setup, writing all found plans into files with an increasing suffix (sas_plan.1  sas_plan.2  sas_plan.3  sas_plan.4, ...). In the IPC, the last fully written plan counts as the final solution.

Since here we would have to verify that the last found plan is complete, taking the last stored plan from the writer (as with a timeout) instead makes sense to me. How do you plan to determine the result status?

Framba-Luca:
Good point! With the code I added the status passed to the new function is the same as if the last plan was parsed from file. 

If an engine needs more granularity, I can easily add a new `_get_anytime_result_status` to differ from the `_get_result_status` of the oneshot planner.

alvalentini:
Hi @roeger! In the last commits @Framba-Luca updated the code to return the last plan found when the `_solve` method returns a `PlanGenerationResult` without a plan. As a default behavior, if the status of the `PlanGenerationResult`returned by the `_solve` method is `UNSOLVABLE`, we replace it with `SOLVED_SATISFICING`.

This behavior is implemented in a method (`_generate_last_result`) that can be re-implemented by the engine if needed.

Do you think it works well for your engine? Any feedback?

roeger:
Hi, sorry that this rested so long. This works for me and could be merged. It's not perfectly beautiful because it somehow puts a plaster on a weird behavior of the overall interaction. If I am not mistaken, the PlanGenerationResult with the UNSOLVABLE status comes from the return value of the _solve method and the fact that Fast Downward appends a suffix to the found plans, so it does not write the expected plan file.

I would find it more natural, if the Writer (parsing the planner output) would instead also look for some string that signals that there won't be any further plans found and created the Result with the last  there.

alvalentini:
Thanks @roeger for the feedback!

I see your point and I agree that for fast-downward better solutions are possible, but I am not sure that they work fine also for the other engines that support anytime planning. We tried to find a solution that works for all the planners.

PR #432, opened 2023-06-20
Documentation improvements

arbimo:
The most important change at this point is the introduction of a list of classes that should appear in the API references.
The list of classes (currently very partial) lies in the  `docs/generate_api_docs.py` script.

This list does not require to be exhaustive but instead should contain the classes that we want to highlight (to help discovery) or refer to in the main documentation.


arbimo:
@alvalentini This is ready I believe. I have reviewed your changes and its seems perfect to me. (thanks again)

PR #433, opened 2023-06-21
Run tests on multiple python versions

alvalentini:


.github/workflows/test.yml
```diff
@@ -0,0 +1,421 @@
...
+        uses: actions/checkout@v2
+
+      - name: Set up Python
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{inputs.python-version}}
+
+      - name: Install Protobuf compiler
+        run: |
+          python3 -m pip install grpcio-tools==1.54.2
```
alvalentini, line 347: Hi @arbimo! I updated the grpcio-tools version from `1.48.1` to `1.54.2` because `1.48.1` does not contain wheels for Python 3.11 and we want to run all the tests for all the supported version (Python 3.8, 3.9, 3.10, 3.11) before every new release. Do you see any issues?

arbimo: No problem, fixing the version is only a matter of making the process reproducible. There should not be any down side of upgrading the version.

Issue #435, opened 2023-06-26
Wrong kind of problem


Shi-Raida:
The following problems are using fluent with `RealType` but the problem does not have the kind `CONTINUOUS_NUMBERS`. Is it not a bug?

- up:robot_with_static_fluents_duration
- up:robot_with_static_fluents_duration_timed_goals
- up:matchcellar_static_duration
- up:locations_connected_cost_minimize

Framba-Luca:
Hi @Shi-Raida. We decided to remove the `CONTINUOUS_NUMBERS` flag when the `RealType` fluent is not used as a number in the expressions but only as a "placeholder" for actions durations and action costs (the 2 examples here).

Ineed, those problems you mentioned don't have the `CONTINUOUS_NUMBERS` flag but have the `STATIC_FLUENTS_IN_DURATIONS` (the first 3) and `STATIC_FLUENTS_IN_ACTIONS_COST` the 4th one.

A small example: you want to write the duration of an action `move(l_from, l_to)` as something depending from `l_from` and `l_to`. You need to define a fluent (for example `distance`), you set the duration as `distance(l_from, l_to)` (or any expression containing it) and  then assign a value to each possible distance (e.g. `distance(l1, l2)`, `distance(l2, l3)` etc.). This example is different from handling numbers in expressions as, after the grounding, you can just assign a fixed number to the different durations of the `move` action (because the distance fluent is static). 

So, if a planner can handle this particular situation, but not numeric fluents in general, only the `STATIC_FLUENTS_IN_DURATIONS` flag will be specified.

arbimo:
> ... This example is different from handling numbers in expressions as, **after the grounding**, you can just assign a fixed number to the different durations of the `move` action (because the distance fluent is static). ...

I think this is where we have a problem in this case: the problem kind should apply to the problem as it is currently, not as it may appear after an hypothetical transformation. In particular, one should not assume that the problem will be grounded.

Framba-Luca:
I get that. The idea behind the decision is that `CONTINUOUS_NUMBERS` and `STATIC_FLUENTS_IN_DURATIONS` are 2 different features of the problem, and **might** be handled differently by an engine; this offers more granularity for the declared `supported_kind`.

The grouding was just to make an example

alvalentini:
The motivation under this choice is more that with `CONTINUOUS_NUMBERS` we mean continuous numbers in numeric planning.

[...]

PR #446, opened 2023-06-30
Support backward compatible PlanKind extensions

arbimo:
Signed-off-by: Arthur Bit-Monnot <abitmonnot@laas.fr>


arbimo:
@mikand @alvalentini This very minimal PR simply add the possibility of adding a version indication when constructing a ProblemKind structure.

It is just a place holder but I suspect that without this, would be blocked in our ability to add new problem kinds in the future without always breaking compatibility.

Without this in particular #435, could not be fixed or mitigated without breaking backward compatibility, and I suspect many other instances will show up.

PR #461, opened 2023-07-25
Added GraphvizGenerator

Alee08:
I added class GraphvizGenerator to make POP graphs more readable for multiagent case. Let me know if you think it's ready to merge

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ setup(
             "up-aries>=0.0.8",
             "up-symk>=0.0.3",
         ],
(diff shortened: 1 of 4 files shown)


Alee08:
Hi @alvalentini! I moved GraphvizGenerator to [plan_plot.py file](https://github.com/aiplan4eu/unified-planning/blob/master/unified_planning/plot/plan_plot.py).
For now I've created a separate _graphviz_partial_order_plan_ , but we could move the code into _plot_partial_order_plan_ if you think it's better.

PR #467, opened 2023-08-01
Improve printing of interval in the ANMLWriter

alvalentini:


Framba-Luca (approved):
LGTM! It only needs an update in tests where the result of the writer must match an hard-coded string.

PR #474, opened 2023-08-16
Added exception when replanner remove_{action/goal} fails

Framba-Luca:
Now the integrated meta-engine replanner does not silently fail when a remove_action or remove_goal does not change the problem.


Framba-Luca:
Thinking about this feature, we could insert a flag that determines if the user wants an exception when the method fails (default behavior) or he prefers a silent fail.

PR #475, opened 2023-08-16
Interchangable order of functions and predicates in reading

Framba-Luca:


unified_planning/io/pddl_grammar.py
```diff
@@ -0,0 +1,349 @@
...
+else:
+    from pyparsing import one_of
+    from pyparsing import rest_of_line
+
+
+Object = "object"
+TypesMap = Dict[str, unified_planning.model.Type]
+
+
+def nested_expr():
```
mikand, line 42: @Framba-Luca I am wondering, would it change anything if we avoid to call this function avery time (createing separate objects) and instead we "recycle" the same object in different points of the grammar?

unified_planning/io/pddl_grammar.py
```diff
@@ -0,0 +1,349 @@
...
+
+
+class PDDLGrammar:
+    def __init__(self):
+        # Domain data structures to populate while parsing
+        self.name: str = ""
+        self.requirements: List[ParseResults] = []
+        self.types: List[ParseResults] = []
+        self.constants: List[ParseResults] = []
+        self.predicates: List[ParseResults] = []
```
mikand, line 73: I cannot find the point where you check that this list has length 1 and you throw an error otherwise... You did this for the problem part, but what about the domain? I think only actions can be a "real" list.

unified_planning/io/pddl_reader.py
```diff
@@ -878,27 +625,33 @@ def _parse_subtasks(
                 f"Could not parse the subtasks list: {e}, from line: {start_line}, col {start_col} to line: {end_line}, col {end_col}"
             )
 
-    def _check_if_object_type_is_needed(self, domain_res) -> bool:
-        for p in domain_res.get("predicates", []):
+    def _check_if_object_type_is_needed(self, domain_res, grammar: PDDLGrammar) -> bool:
+        empty_parse_res = ParseResults()
+        predicates = grammar.predicates[0][0] if grammar.predicates else empty_parse_res
```
mikand, line 630: This indexing is **very** ugly. Can't we have a grammar property instead?

unified_planning/io/pddl_grammar.py
mikand: One alternative to using `Lists` here would be an `Optional` and throwing an error in the patrsing action if the value is not `None` before replacing...

mikand (changes requested):
@Framba-Luca please try the proposed suggestions. Also see if recycling the nextedExpr object makes any change in performance

unified_planning/io/pddl_grammar.py
Framba-Luca: I created an `append_unique` method that enforces this behavior directly at parsing time. I tried the `Optional` and `None` approach, creating a unique method that used `getattr` and `setattr` but it was slower than the current approach.

unified_planning/io/pddl_reader.py
Framba-Luca: Managed to clan it up by removing obsolete `Group` in the grammar.

unified_planning/io/pddl_grammar.py
Framba-Luca: Actually, recycling the same object actually worsen performance by a lot. For now I didn't include it in the commit, but it can be investigated more deeply.

PR #481, opened 2023-08-29
merged master plus added new version up-enshp

hstairs:


hstairs:
Hi all,

there seems to be something broken here, but doesn't seem to be related on my part. It seems an error in some parsing of a tamer file...not sure how to overcome this. LMK

Enrico

alvalentini:
The problem is that this PR removes some files

hstairs:
Ok I see. Closing this and opening a new one

PR #489, opened 2023-09-11
Tests: add a basic numeric example

Shi-Raida:


unified_planning/test/test_problem.py
```diff
@@ -538,6 +539,7 @@ def test_simple_numeric_planning_kind(self):
                     problem.kind.has_simple_numeric_planning(), problem.name
                 )
             else:
+                print(problem.name)
```
alvalentini, line 542: Can you remove this `print`?

PR #493, opened 2023-09-27
Adding Support for Axioms and Derived Predicates

speckdavid:
As part of the 3rd Sprint of the "Symbolic Search for Diverse Plans and Maximum Utility" project, we added support for **Axioms + Derived Predicates** to the UP. So that these can be represented internally as well as read from and written to PDDL. 

We also created a notebook as a usage example:  [Axioms and Derived Predicates in the Unified Planning Library](https://github.com/speckdavid/up-symk/blob/master/notebooks/axioms_usage.ipynb).

Best
David

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ setup(
         "lpg": ["up-lpg==0.0.7"],
         "fmap": ["up-fmap==0.0.7"],
         "aries": ["up-aries>=0.0.8"],
-        "symk": ["up-symk>=0.0.3"],
+        "symk": ["up-symk>=1.0.0"],
         "engines": [
             "tarski[arithmetic]",
             "up-pyperplan==1.0.0.1.dev1",
@@ -40,7 +40,7 @@ setup(
             "up-lpg==0.0.7",
             "up-fmap==0.0.7",
             "up-aries>=0.0.8",
-            "up-symk>=0.0.3",
+            "up-symk>=1.0.0",
         ],
         "plot": [
             "plotly",


RobertL31:
Would appreciate this feature in my actual project, looking forward to it :D !

unified_planning/model/axiom.py
```diff
@@ -0,0 +1,112 @@
...
+"""
+This module defines the base class `Axiom'.
+An `Axiom' has a `head' which is a single `Parameter' and a `body' which is a single `Condition'.
+"""
+
+from unified_planning.model.action import *
+from unified_planning.environment import get_environment, Environment
+
+
+class Axiom(InstantaneousAction):
```
mikand, line 24: Why deriving from `InstantaneousAction`? This makes it possible to have `Axioms` passed to the `add_action` methods and causes some hacky code below. Why not a dedicated class?

unified_planning/model/types.py
```diff
@@ -75,6 +79,17 @@ def is_bool_type(self) -> bool:
         return True
 
 
+class _DerivedBoolType(Type):
```
mikand, line 82: Does this information needs to be in the typing system or we can keep a list/map/set of predicates that are derived in the `Problem`?

mikand (changes requested):
@speckdavid Many thanks for this PR.
@roeger I know very little about derived predicates and I left only a couple of comments. Can you do a more detailed code review?

unified_planning/model/types.py
speckdavid: In the beginning, I simply wanted to use the Boolean type, but I ran into some problems. For example, when a planning problem is created, an assertion checks whether all variables have an initial value. For derived predicates, however, we cannot set such a value. 

I think your suggestion of having a list/set of variables that are derived would also work if we can access it at the appropriate places. It will definitely simplify the code as the typing system is not touched. I will check it.

unified_planning/model/axiom.py
hstairs: Can you have a look at the internal structure of the InstantaneousActiona and see what is that you really need? Maybe this could be a call to a small refactoring aimed at factorising what is needed to both instantaneous actions and axioms.

jendrikseipp:
I just stumbled over this as well: I had thought that the UP supports derived predicates already. Would be great to have this in the library!

PR #498, opened 2023-10-09
Move Test cases in the up_test_cases folder

Framba-Luca:
This PR aims to collect, refactor and expand the scattered code of the planning test cases.


arbimo:
I have another python version problem when running this one in the CI of Aries:

```
  File "/home/runner/work/aries/aries/planning/unified/deps/unified-planning/up_test_cases/utils.py", line 68, in _get_pddl_test_cases
    pddl_files = glob("*.pddl", root_dir=pddl_files_path)
TypeError: glob() got an unexpected keyword argument 'root_dir'
```

This parameter was introduced in python 3.10. It is strange that this is not cached in the CI

arbimo:
This looks fantastic to me! (the annoying part being that it tells that I still have a few problems to fix :/)

In my workflow with report, I had an additional script `run.py` that would take as command line parameters the name of a planner and of a test case and run the solver in verbose mode on this problem. (essentially boiling down to the code below)

This was very useful to get detailed information on the test case that was problematic with report. It may be overkill for our purposes but having a way to get the planner's logs would be very useful to actually debug things. One bottleneck there is to get from the name of the test case to the actual `Problem` object.

```py
test_cases = report.all_test_cases()

planner_name = sys.argv[1]
problem_name = sys.argv[2]

problem = [test_case.problem for test_case in test_cases if test_case.name == problem_name][0]

print(problem)
print(problem.kind)
planner = OneshotPlanner(name=planner_name)
print("supported:", planner.supports(problem.kind))

result = planner.solve(problem, output_stream=sys.stdout, timeout=1)



print("======= UP RESULT =====")
print(result)

val = PlanValidator(name="aries-val")
print(val.validate(problem, result.plan))
```

Framba-Luca:
Thanks again for reporting the mistake, now your tests should pass.

It probably isn't caught by the CI because it is tested on python 3.10 (and we test on other versions only after the merge on master). It might also be because the report file isn't really tested in the CI, but just by hand (for now).

Regarding the `run.py` script, it's a really cool and useful functionality; would have saved me quite some time in doing the errors report! We are considering if it should be it's own file or just a specific configuration of the `report.py` script.

alvalentini:
We will add the verbose mode in another PR.

PR #507, opened 2023-10-26
new version of enhsp

hstairs:


arbimo:
General question: don't we want to avoid specifying exact version. Since we (developers) are working with the master branch exact version work ok. But with the current schema, if a user wants an updated version of (e.g.) ENHSP with a bug fix, we need to cut a new release of the UP (with updated version numbers) to comply with the dependency requirements.

I use `>=` for versions of Aries. This can be demanding for maintaining backward compatibilty. Note that I can as well require a minimum version of the UP on Aries side.
An other less demanding option is the `~=` specifier that lets pip select semantically compatible versions (https://peps.python.org/pep-0440/#compatible-release)

hstairs:
@arbimo Can you point me where in the code of Aries you enforce a minimum version of the UP in Aries?

PR #509, opened 2023-10-27
Feat: Create STN from a TTP

anissaab:


unified_planning/plans/ttp_to_stn.py
```diff
@@ -0,0 +1,270 @@
...
+from unified_planning.plans.sequential_plan import SequentialPlan
+from unified_planning.plans.partial_order_plan import PartialOrderPlan
+from unified_planning.plot.plan_plot import plot_partial_order_plan
+
+EPSILON = Fraction(1, 1000)
+MAX_TIME = Fraction(5000, 1)
+
+import operator
+
+ops = {"+": operator.add, "-": operator.sub}
```
Framba-Luca, line 19: Never used?

unified_planning/plans/ttp_to_stn.py
```diff
@@ -0,0 +1,270 @@
...
+                    ]
+                else:
+                    ai_next_time = [
+                        item[0]
+                        for item in self.events_sorted
+                        if item[1] == ai_next.action
+                    ][0]
+                    time_edge = ai_next_time - ai_current_time
+                    graph[ai_current][ai_next]["interval"] = [time_edge, time_edge]
+
```
Framba-Luca, line 202: It's not perfectly clear to me what is happening here. Maybe we can have a separate chat about this!

Framba-Luca (commented):
Very nice work. :smiley: 

It still needs some interface work before merging, but I discovered I will probably have to do it so I wrote mostly general tips.

unified_planning/plans/ttp_to_stn.py
```diff
@@ -0,0 +1,270 @@
...
+        """
+        Stn with Start and End nodes
+        Return the stn from the partial order plan.
+        """
+        graph = self.partial_order_plan._graph
+
+        for ai_current, l_next_ai in self.partial_order_plan.get_adjacency_list.items():
+            ai_current_time = [
+                item[0] for item in self.events_sorted if item[1] == ai_current.action
+            ][0]
```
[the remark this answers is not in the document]
anissaab: Because of adjacency list is a dict of ActionInstance and not a dict of Instantaneous action in the other way around I need to cross the adj list also.

nicola-gigante:
Hello! This would be very valuable for me in a project I'm starting. Will this work on `TimeTriggeredPlan`s obtained from a `HierarchicalProblem` with durative actions?

anissaab:
> Hello! This would be very valuable for me in a project I'm starting. Will this work on `TimeTriggeredPlan`s obtained from a `HierarchicalProblem` with durative actions?

Hi ! It shouldn't be a problem ! If you use a `TimeTriggeredPlan` and a problem with durative actions you should obtains a STN. The Problem is only used when converting the TTP to a `PartialOrderPlan`, with the already implemented function `_to_partial_order_plan`.

anissaab:
> Hi @anissaab, thanks for answering!
> 
> I'm not familiar with how you extract the STN, but I suppose you need to compare the plan with problems' the durative actions, right?
> 
> But, in a `HierarchicalProblem`, there are some constraints that are not represented in the actions but come from the task network.
> 
> Suppose that two durative actions `A1` and `A2` implement two subtasks `T1` and `T2`, and we have an ordering constraint `T1 < T2`. Then, `A1` will come before `A2` in the TTP. Now, even if the durative actions themselves do not have any conflicting conditions, the STN extracted from the TTP has to still guarantee that `A2` starts after the end of `A1`.
> 
> Is this something you can handle already? Or am I making things more complex than needed?

Indeed, for now the algorithm currently does not take into consideration task precedence. But it shouldn't be complicated to insert it for `HierarchicalProblem` !

mikand:
@nicola-gigante Handling HTN is a cool idea! I would keep the scope of this PR to "flat" problems, but it is surely an interesting extension! If you have some CPU to work on this, maybe you can start a new PR (I would suggest to start from the non-temporal case and work on the partial order plan construction). I am sure @arbimo will have nice suggestions! :smile:

nicola-gigante:
Hi @mikand!

Yes, I have some CPU do work on this :)

@arbimo is my intuition correct that to support a `HierarchicalProblem` we only need to refine the STN to account for task precedence constraints?

arbimo:
@nicola-gigante Yes that is correct. The only (minor) annoyance being that constraints are expressed between a mix of abstracts and primitive tasks, so there is some transitive reasoning to do.

The `TaskNetwork` class provides some method to extract the temporal constraints and is a good entry point. Note that a "temporal constraints" is just a boolean expression that refers to a timepoint. It might be nested into others expression. The total-order and partial-order cases (and corresponding problem kinds) are the cases you probably want to focus first on.  It should be easy to add a TASK_ORDER_STN kind that would require temporal constraints to form an STN (in the most expressive model they could for instance form a DTP).

PR #523, opened 2023-11-17
fix(htn): Remove `ACTION_BASED` class from HTN planning problem kind.

arbimo:
While it historically made sense to have it there, it prevented pure HTN planners from declaring
that they do not support non-hierarchical planning.

Note that this is purely backward compatible (surprisingly).


alvalentini (approved):
LGTM

PR #528, opened 2023-11-24
Add agriculture use-case to documentation

sebastianstock:
This adds the description for the agricultural use-case as a showcase to the documentation.


arbimo:
@sebastianstock I have changed the image links from html to markdown syntax, otherwise they are not picked up by sphynx when building the doc. (You can look in the contributors's guide section of the doc to see how to preview changes).

If you want to rescale the images, the best option is to rescale them directly (it is also better for bandwidth). But you can also submit the document in reStructuredText that has more options to configure rendering.

Let me know if you want to update the PR or if it should go in as it is.

arbimo (approved):

sebastianstock:
Dear @arbimo, thanks for fixing the links and the hint about the image size. I have rescaled the images and checked it locally. Now it looks fine.

PR #531, opened 2023-11-24
Update 01_available_engines.rst

fabio-patrizi:
Added fmap to available engines


arbimo (changes requested):
Thanks @fabio-patrizi 

However, I think MA planning should go its own setction (like hierarchical planning), otherwise we will quickly overload the tables with to much information.

arbimo (approved):
Perfect, thanks!

PR #538, opened 2023-12-04
Fix Problem's timed effects in Protobuf Writer

### LINKED ISSUE #536: Crash of `aries` engine integration code when adding timed effects
**Describe the bug**
The `aries` engine integration crashes when a *timed effect* is added to a problem.

**To Reproduce**
Run the following:

```
import unified_planning as up
from unified_planning.shortcuts import *
from unified_planning.model.htn import *

problem = HierarchicalProblem()

done = Fluent('done', BoolType())
problem.add_fluent(done)
problem.set_initial_value(done, False)

problem.add_timed_effect(
    Timing(60, Timepoint(TimepointKind.GLOBAL_START)), done, True
)

print(problem.timed_effects)

with OneshotPlanner(name='aries') as planner:
    res = planner.solve(problem)
    print(res)
```

The output is the following:

```
{start + 60: [done := true]}
  File "/usr/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/ngigante/projects/telaio/backend/planning/models/issue.py", line 19, in <module>
    res = planner.solve(problem)
  File "/home/ngigante/projects/telaio/.venv/lib/python3.9/site-packages/unified_planning/engines/mixins/oneshot_planner.py", line 80, in solve
    return self._solve(problem, heuristic, timeout, output_stream)
  File "/home/ngigante/projects/telaio/.venv/lib/python3.9/site-packages/up_aries/solver.py", line 347, in _solve
    server, req = self._prepare_solving(problem, heuristic, timeout, output_stream)
  File "/home/ngigante/projects/telaio/.venv/lib/python3.9/site-packages/up_aries/solver.py", line 310, in _prepare_solving
    proto_problem = self._writer.convert(problem)
  File "/home/ngigante/projects/telaio/.venv/lib/python3.9/site-packages/unified_planning/grpc/converter.py", line 37, in convert
    return f(element, *args)
  File "/home/ngigante/projects/telaio/.venv/lib/python3.9/site-packages/unified_planning/grpc/proto_writer.py", line 587, in _convert_problem
    return proto.Problem(
TypeError: Parameter to MergeFrom() must be instance of same class: expected <class 'unified_planning_pb2.TimedEffect'> got <class 'unified_planning_pb2.Timing'>.
```

Since `MergeFrom` is a `protobuf` function and the `aries` integration uses `protobuf` to communicate to the backend, I suspect this is a problem in the conversion of the UP `Problem` class to the `unified_planning.grpc.generated.unified_planning_pb2.Problem` class.

Note that the fact that I'm creating a `HierarchicalProblem` here is irrelevant.
The same happens if I create a simple `Problem`.

**Expected behavior**
I would expect to get whatever result the planning problem has, instead of an exception.

**Additional context**
I'm on Python 3.9 on Debian Linux.

Framba-Luca:
Fix #536

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ on:
         type: string
 
 env:
-  up_tamer_commit: "4d179b5c2722758127bf684d732a6a44431a1732"
+  up_tamer_commit: "1b543623ad2ebd9795529abe812352db8f4ec014"
   up_pyperplan_commit: "6cec570b88d62ca5c4b8029a60c704414fe7f6a9"
   up_fast_downward_commit: "090977461c6a55812bc16d5e2df55e556dbbdfb5"
   up_enhsp_commit: "af93cd799974626ac17b8aaf1a83975f75c073ab"
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ setup(
         "fast-downward": ["up-fast-downward==0.4.0"],
         "lpg": ["up-lpg==0.0.10"],
         "fmap": ["up-fmap==0.0.12"],
-        "aries": ["up-aries>=0.3.1"],
+        "aries": ["up-aries>=0.3.2"],
         "symk": ["up-symk>=1.0.1"],
         "engines": [
             "tarski[arithmetic]",
@@ -39,7 +39,7 @@ setup(
             "up-fast-downward==0.4.0",
             "up-lpg==0.0.10",
             "up-fmap==0.0.12",
-            "up-aries>=0.3.1",
+            "up-aries>=0.3.2",
             "up-symk>=1.0.1",
         ],
         "plot": [
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ setup(
         "grpc": ["grpcio", "grpcio-tools", "grpc-stubs"],
         "tarski": ["tarski[arithmetic]"],
         "pyperplan": ["up-pyperplan==1.0.0.6.dev1"],
-        "tamer": ["up-tamer==1.0.0.6.dev1"],
+        "tamer": ["up-tamer==1.0.0.8.dev1"],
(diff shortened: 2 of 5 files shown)


arbimo:
I just expanded a bit the example to better test the compliance of validators.

It seems however that we now hit a limitation of the PDDL writer that does not support TILs...

Framba-Luca:
Thanks for expanding. I did a quick fix to mypy, so now we see the PDDL test failing. I think I just have to add a skip if there are TILs to the problem in the PDDL testing file

alvalentini:
@arbimo I'm going to add TILs support in the PDDLWriter in a different PR.

arbimo:
One last thing. The test cases trigger a corner case of the temporal planning semantics: when should the goals hold in the presence of timed initial literals ? The PDDL 2.2 semantics require the goals to hold after all TILs are computed (the real final state).
Another possible interpretation would be "immediately after the end of the last action".

Looking at report it seems that:
 - tamer and aries-val adheres to the PDDL 2.2 semantics
 - aries and tamer-val adheres to the second one

I believe we should enforce the PDDL2.2 semantics as they are the least surprising ones (and most established). This is the one enforced in the test-cases added in this PR.

Note that this is truly a corner case as it is very unusual to have goal conditions on exogeneous state variables in real problems.

@mikand Feel free to jump in on this, otherwise I will merge it as it is (which means that down the line, aries and tamer-val will need to align themselves on the PDDL semantics).

arbimo (approved):

alvalentini:
@arbimo I agree with you and the PDDL 2.2 semantics. I'm going to fix the tamer validator.

arbimo:
@alvalentini Perfect, we agree. I'll merge this one then (and add this aries' fix to my todos)

Thanks a lot to both of you!

PR #540, opened 2023-12-04
Doc underwater

ElisaTosello:
Add underwater use-case description


arbimo (approved):
Looks perfect! 
I just pushed a very minor rewording on a sentence that was a bit ambiguous to me. Could you just check that I didn't misinterpret it?

Otherwise this is good to go for me.

ElisaTosello:
Your change is perfect! The sentence is actually more understandable!

PR #541, opened 2023-12-04
Update 01_available_engines.rst

serivan:
Added lpg info


arbimo (approved):
LGTM! I just pushed LPG higher up in the list as it should typically be preferred to Aries and pyperplan for action-based planning.

PR #543, opened 2023-12-06
Add support to TILs in the PDDLWriter

alvalentini:


unified_planning/io/pddl_writer.py
```diff
@@ -686,7 +686,20 @@ def _write_problem(self, out: IO[str]):
             else:
                 out.write(f" (= {converter.convert(f)} {converter.convert(v)})")
         if self.problem.kind.has_actions_cost() or self.problem.kind.has_plan_length():
-            out.write(f" (= (total-cost) 0)")
+            out.write(" (= (total-cost) 0)")
+        for tm, le in self.problem.timed_effects.items():
+            for e in le:
+                out.write(f" (at {tm.delay}")
```
arbimo, line 692: If not an integer would output a fraction.

For instance, round trip reading/writing of the truck domain gives me : `(at 9197/10 (not (deliverable package1 l1)))`

unified_planning/io/pddl_writer.py
```diff
@@ -1061,22 +1075,24 @@ def _write_effect(
...
     if simplified_value.is_true():
-        out.write(f" {converter.convert(effect.fluent)}")
+        out.write(f" {fluent}")
     elif simplified_value.is_false():
-        out.write(f" (not {converter.convert(effect.fluent)})")
+        out.write(f" (not {fluent})")
+    elif initial:
+        value = converter.convert(simplified_value)
+        if effect.is_increase():
+            value = f"(+ {fluent} {value})"
```
arbimo, line 1086: This does not seem to be mentionned in the PDDL 2.2 references (that I believe assume literals in TIL). Do we have any example of this syntax usage ? (I am surprised because I would have assumed `increase`/`decrease` to be used there).

unified_planning/io/pddl_writer.py
```diff
@@ -1061,22 +1075,24 @@ def _write_effect(
...
+        elif effect.is_decrease():
+            value = f"(- {fluent} {value})"
+        out.write(f" (= {fluent} {value})")
```
arbimo, line 1089: Isn't an `else` missing? (together with an assert `is_assign()`

arbimo (changes requested):

unified_planning/io/pddl_writer.py
alvalentini: You are right!

unified_planning/io/pddl_writer.py
alvalentini: This is not in the standard, both in this way or with the `increase`/`decrease`.
Maybe it is better to raise an exception. What do you think?

arbimo: I have looked around a bit and only found one in a [short paper](https://kclpure.kcl.ac.uk/portal/en/publications/planning-with-numeric-timed-initial-fluents). They only consider assignment to numeric fluents, no increase/decrease.

I think their proposal to use `=` for assignments in timed effects is misleaded and that timed effects should use exactly the same syntax as in regular action effects. That is one of:

    (at <time-spec> <l>)
    (at <time-spec> (not <l>))
    (at <time-spec> (assign <sv> <v>))
    (at <time-spec> (increase <sv> <v>))
    (at <time-spec> (decrease <sv> <v>))


That way there is no new syntax introduced for timed effects. They only differ from action effects in what is allowed for <time-spec>: strictly positive number for TILs and `start`/`end` for actions.

I believe it is fine to introduce such very minor syntax extensions as long as (1) it solves a common problem, (2) it is perfectly unambiguous and (3) we are convinced that the syntax is future-proof. I think the above proposal matches all three.


(Note that we are still missing the `scale-up`/`scale-down` modifiers from PDDL 2.1)

alvalentini: I think your proposal is reasonable. I updated the code accordingly.

Where do you think we should document this extension?

unified_planning/io/pddl_writer.py
```diff
@@ -428,6 +429,16 @@ def _write_domain(self, out: IO[str]):
...
             ):
                 out.write(" :action-costs")
+            if self.problem_kind.has_timed_effects():
+                only_bool = True
+                for le in self.problem.timed_effects.values():
+                    for e in le:
+                        if not e.fluent.type.is_bool_type():
+                            only_bool = False
+                if not only_bool:
+                    out.write(" :timed-effects")
```
arbimo, line 439:
```suggestion
                    out.write(" :timed-initial-effects")
```

unified_planning/io/pddl_reader.py
```diff
@@ -101,7 +101,7 @@ def __init__(self):
             + ":requirements"
             + OneOrMore(
                 one_of(
-                    ":strips :typing :negative-preconditions :disjunctive-preconditions :equality :existential-preconditions :universal-preconditions :quantified-preconditions :conditional-effects :fluents :numeric-fluents :adl :durative-actions :duration-inequalities :timed-initial-literals :action-costs :hierarchy :method-preconditions :constraints :contingent :preferences"
+                    ":strips :typing :negative-preconditions :disjunctive-preconditions :equality :existential-preconditions :universal-preconditions :quantified-preconditions :conditional-effects :fluents :numeric-fluents :adl :durative-actions :duration-inequalities :timed-initial-literals :timed-effects :action-costs :hierarchy :method-preconditions :constraints :contingent :preferences"
```
arbimo, line 104:
```suggestion
                    ":strips :typing :negative-preconditions :disjunctive-preconditions :equality :existential-preconditions :universal-preconditions :quantified-preconditions :conditional-effects :fluents :numeric-fluents :adl :durative-actions :duration-inequalities :timed-initial-literals :timed-initial-effects :action-costs :hierarchy :method-preconditions :constraints :contingent :preferences"
```

unified_planning/io/pddl_writer.py
```diff
@@ -120,6 +120,7 @@
     "durative-actions",
     "derived-predicates",
     "timed-initial-literals",
+    "timed-effects",
```
arbimo, line 123:
```suggestion
    "timed-initial-effects",
```

arbimo (approved):
LGTM. I just suggested using `timed-initial-effect` has feature name to make it less ambiguous and avoid potential clashes.

Right now there is no place for it in the documentation but it should be mentioned in a future "advanced temporal modeling features".

PR #550, opened 2023-12-15
doc: various improvements to documentation

arbimo:
This provides several minor improvements to the documentation. It mostly polishes and fixes some minor things.

A notable changes is regarding the "Optimality" documentation that now refers to ANytime planning and suggest using it when no optimality guarantee is needed.

The other important change is to the contributors guide that is now much more complete.

The most controversial thing is the addition of a `justfile` at the root of the repository that should aim at documenting, making discoverable and easily accessible the various commands taht are useful to interact with the code base.
[just](https://github.com/casey/just) is command line utility that you can think of as a much better `make`. If you haven't used it I strongly suggest giving it a try. 
Even without using it the justfile is useful in centralizing the different commands (without going into the CI configuration as I usually have to do).

The objective of the PR is not to generalize the usage of just but if there is consensus that it is a desirable thing it could be be extended (e.g. for running the tests) and used in CI to avoid duplication.

### DIFF
--- a/docs/contributor_guide.rst
+++ b/docs/contributor_guide.rst
@@ -1,7 +1,30 @@
 Contributor's Guide
 ===================
 
+This page is dedicated to the technical information that you may find useful when contributing code or documentation to the UP repository.
+For information on the contribution process, please have a look to the `corresponding page <https://github.com/aiplan4eu/unified-planning/blob/master/CONTRIBUTING.md>`_.
 
+Common commands (Justfile)
+--------------------------
+
+A list of common commands and recipes are available in a `justfile <https://github.com/aiplan4eu/unified-planning/blob/master/justfile>`_.
+You can use it to find out common commands (e.g. running tests, reformatting the code) or use it directly with the `just <https://github.com/casey/just>`_ command line tool.
+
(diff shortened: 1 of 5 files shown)


docs/contributor_guide.rst
```diff
@@ -1,7 +1,30 @@
...
+For information on the contribution process, please have a look to the `corresponding page <https://github.com/aiplan4eu/unified-planning/blob/master/CONTRIBUTING.md>`_.
 
+Common commands (Justfile)
+--------------------------
+
+A list of common commands and recipes are available in a `justfile <https://github.com/aiplan4eu/unified-planning/blob/master/justfile>`_.
+You can use it to find out common commands (e.g. running tests, reformatting the code) or use it directly with the `just <https://github.com/casey/just>`_ command line tool.
+
+Whenever you see a command ``just RECIPE_NAME`` in this page, it means the command is available as a just recipe.
+You can run it directly with ``just`` or check the ``justfile`` for the actual commands to run.
```
mikand, line 14: Wow! I did not know `just`, I was missing out!

mikand (approved):
LGTM! Thanks!

PR #552, opened 2023-12-19
Small fixes

Framba-Luca:
This PR fixes 2 problems:

- a bug in the UsertypeFluentsRemover when used in a compiler's pipeline
- the error message given by the factory when an engine is not selected due to the missing optimality gurantees

### DIFF
--- a/unified_planning/engines/factory.py
+++ b/unified_planning/engines/factory.py
@@ -531,8 +531,14 @@ class Factory:
                 )
                 planners_features.append(x)
         if len(planners_features) > 0:
+            if optimality_guarantee is not None:
+                starting_line = f"No available engine supports all the problem features with optimality_guarantee: {optimality_guarantee.name}:"
+            elif anytime_guarantee is not None:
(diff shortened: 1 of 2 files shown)


unified_planning/engines/factory.py
```diff
@@ -531,8 +531,14 @@ def _get_engine_class(
...
                 planners_features.append(x)
         if len(planners_features) > 0:
+            if optimality_guarantee is not None:
+                starting_line = f"No available engine supports all the problem features with optimality_guarantee: {optimality_guarantee.name}:"
+            elif anytime_guarantee is not None:
+                starting_line = f"No available engine supports all the problem features with anytime_guarantee: {anytime_guarantee.name}:"
+            else:
+                starting_line = "No available engine supports all the problem features:"
             header = ["Engine"] + problem_features
-            msg = f"No available engine supports all the problem features:\n{format_table(header, planners_features)}"
```
alvalentini, line 535: Here we should add a column to indicate for each engine if it supports the optimality guarantee

alvalentini (changes requested):
@Framba-Luca

PR #557, opened 2024-01-11
Add a custom parameter to disable action pruning in the UP grounder

alvalentini:


Framba-Luca (approved):
LGTM!

mikand:
OK for me.
For future reference: we needed this to construct a grounding that remains valid even if the initial state is changed (this is not true if static fluents are used for simplification)

PR #559, opened 2024-01-16
perf: Optimize generation of initial state for default fluents.

arbimo:
This patch improves the performance of generating the initial state, essentially by caching the domain of each parameter.

On a non trivial problem I tested, it brings the execution time from ~21 seconds to ~1 second, bringing significant gains when this is the bottleneck (e.g. when converting to protobuf).


alvalentini (approved):
LGTM! Great improvement!

PR #560, opened 2024-01-25
Fix plan validation via command line interface.

jendrikseipp:
The `args.engine_names` attribute is not present in validation mode.

### DIFF
--- a/unified_planning/cmd/up.py
+++ b/unified_planning/cmd/up.py
@@ -155,7 +155,7 @@ def plan_validation(
     args: argparse.Namespace,
 ):
     problem = parse_problem(parser, args)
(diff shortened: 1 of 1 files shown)


alvalentini:
Hi @jendrikseipp, thanks for the PR and for signaling the bug!

It is actually possible to validate plans using more plan validators in parallel (similarly to parallel oneshot planning).
The behavior is that the first result is returned.

The bug is in the command line arguments parser, that do not allow the use of `engine_names` for plan validation. 
@Framba-Luca fixed it in this PR #561. Can you check if #561 fixes your problem? Thanks

jendrikseipp:
Thanks for the quick answer! I haven't tested it locally, but the code seems to address the problem correctly.

PR #562, opened 2024-01-25
fix: proto explicit init

arbimo:
Until now, all initial values (including implicit ones) written in the initial state when converting to protobuf. 
This is problematic as it means the problem reconstructed after export and importing to protobuf was not strictly equivalent: some implicit values were made explicit in process.

A bad consequence is that it made protobuf conversions much more expensive that it had too (even after the optimization of #559, 90% of the time spend remains in the initial state generation in at least some non trivial problems).

This PR only save to protobuf the explicitly set initial values. The defaults values of the fluent were already there.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ setup(
         "fast-downward": ["up-fast-downward==0.4.1"],
         "lpg": ["up-lpg==0.0.10"],
         "fmap": ["up-fmap==0.0.13"],
-        "aries": ["up-aries>=0.3.2"],
+        "aries": ["up-aries>=0.3.3"],
         "symk": ["up-symk>=1.0.1"],
         "engines": [
             "tarski[arithmetic]",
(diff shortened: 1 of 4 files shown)


alvalentini (approved):
LGTM! Just a note: the problem reconstructed is not exactly the same since it doesn't contain the ["general" defaults](https://github.com/aiplan4eu/unified-planning/blob/master/unified_planning/model/mixins/fluents_set.py#L46). But I don't think it is a problem. You can proceed with the merge.

arbimo:
Thanks for the review! I'll merge it as it is then.

> the problem reconstructed is not exactly the same since it doesn't contain the ["general" defaults](https://github.com/aiplan4eu/unified-planning/blob/master/unified_planning/model/mixins/fluents_set.py#L46). But I don't think it is a problem. You can proceed with the merge.

You are right, but this will be reflected on the fluent's default (even though the problem will not be exactly similar if you were to add new fluents). I think I added this step of promoting a global default to a fluent default precisely because I was not very at ease with adding global default in the protobuf encoding ^^

PR #563, opened 2024-01-25
Adding continuous effects to durative actions

Alelafar:
I added the **decrease_continuous_effect** and **increase_continuous_effect** functions in order to have a continuous change effect in a durative action and I added them also as two new ProblemKind. 

Moreover, I suitably modified the PDDL Writer in order to dump the new tool to the PDDL domain.


alvalentini:
Moved the changes in #657

PR #576, opened 2024-03-04
feat: Allow partially defined initial state

### LINKED ISSUE #526: Problem loading classical planning instances with action cost
**Describe the bug**
The library is too restrictive by requiring all fluents to be declared in the initial state. This causes the library not to work with exsiting benchmarks: the transport domain used in IPC'08, 11 and 14, for example.  
My use case only involves loading the domain and problem file and solving them with a planner (I'm using fast downward). I include a minimal script below.

The reason is that there is a fluent (road-length ?l1 ?l2 - location) that indicates the length of the road between two locations. This is used as the cost of the action drive. However, there is also a predicate (road ?l1 ?l2 - location) that specifies that there is a road between the two locations. So, the problem file does not specify the length of roads that do not exist.

Currently, the library complains that all values should be defined: Initial value not set for fluent: road-length(city-1-loc-1, city-1-loc-1)
However, having to enumerate all values would be cumbersome and may significantly increase the size of the instances. 

So, it would be desirable if the library could admit fluents that are not fully specified, trusting the user that the undefined values are not relevant.

**To Reproduce**
```
#!/usr/bin/env python3

from unified_planning.shortcuts import *
from unified_planning.io import PDDLReader

reader = PDDLReader()
domain_filename = "./downward-benchmarks/transport-opt11-strips/domain.pddl"
problem_filename = "./downward-benchmarks/transport-opt11-strips/p01.pddl"

problem = reader.parse_problem(domain_filename, problem_filename)

with  OneshotPlanner(names=['fast-downward']) as planner:
    final_report = planner.solve(problem)
```

**Result of the above script**
```
  File "/home/alvaro/teaching/phd-planning/src/./demo-load.py", line 14, in <module>
    final_report = planner.solve(problem)
  File "/home/alvaro/.local/lib/python3.10/site-packages/unified_planning/engines/mixins/oneshot_planner.py", line 80, in solve
    return self._solve(problem, heuristic, timeout, output_stream)
  File "/home/alvaro/.local/lib/python3.10/site-packages/unified_planning/engines/parallel.py", line 138, in _solve
    final_reports = self._run_parallel("solve", problem, timeout, None)
  File "/home/alvaro/.local/lib/python3.10/site-packages/unified_planning/engines/parallel.py", line 108, in _run_parallel
    raise res
unified_planning.exceptions.UPProblemDefinitionError: Initial value not set for fluent: road-length(city-1-loc-1, city-1-loc-1)
```

### LINKED ISSUE #569: Initial value not set.
**Describe the bug**
I'm trying to solve the transport domain instance 10 and the planner raise this error:
Screenshot 2024-02-21 at 10 27 26 AM

**To Reproduce**
Here is the code to reproduce
```
from unified_planning.io import PDDLReader
import up_symk
from unified_planning.shortcuts import *
from unified_planning.plans import SequentialPlan

domain  = './domain.pddl'
problem = './p10.pddl'
task = PDDLReader().parse_problem(domain, problem)

plannername   = 'symk-opt'
with OneshotPlanner(name=plannername) as planner:
    result = planner.solve(task)
    seedplan = result.plan if result.status in unified_planning.engines.results.POSITIVE_OUTCOMES else SequentialPlan([], task.environment)
```

**Additional context**
Here is the [planning task](https://github.com/aiplan4eu/unified-planning/files/14358151/instance.zip) used to trigger this error.

arbimo:
This PR adds support for undefined values in the initial state.

In essence, it introduces two new problem features that capture whether a problem has numeric/non-numeric state variables  with undefined values in the initial state.
If one of these features hold, then `problem.initial_value(...)` may return `None`.

fix #569
fix #526

todo: Missing the doc for the new features


mikand:
@arbimo This makes sense, but waht would be the semantics for a problem with uninitailized value? If you remember, engines and validators in PDDL have a wide variety of interpretation of waht is allowed and disallowed...

arbimo:
The interpretation in this PR (enforced by a test case) is the most conservative one: the evaluation of an expression is undefined if it refers to an undefined value, as per the PDDL spec.

My position on the matter is that we *have to* allow models with undefined values as their are simply to many benchmarks with those. This issue was already reported 3 times (+2 if you count myself and another PhD at LAAS).
For the semantics, we need to pick a "preferred" one and the one from the PDDL spec appears to be the least evil in the absence of a consensus in the community.

hstairs:
@arbimo I agree but let me try to summarize what I think we are implying here, at least for the numeric case which I think is the more critical. There are two cases. Numeric expressions in conditions and numeric expressions as right-hand side of effects (and let's treat inc and dec as assignment here). Conditions: any numeric expression containing an undefined numeric value in a state s is undefined is s, and the numeric condition that contains it evaluates to false in s. Any formula containing an undefined numeric condition (e.g., \phi = x +y >= 5) as a subformula is false (e.g., state s = {...x = undefined} \not \models \phi) . Numeric effect: any effect x = exp having exp undefined is undefined. An action having at least one undefined effect in s is not applicable. I think this is the closest semantics to PDDL2.1 and also the most conservative one, and hopefully I am not forgetting anything :)

alvalentini (approved):
@arbimo @hstairs 

I rebased the branch and I added in the documentation the semantic clarifications for the undefined value in the initial state. For me you can proceed with the merge.

PR #577, opened 2024-03-06
feat(htn): Add support for exporting hierarchical problems to HDDL

### LINKED ISSUE #506: PDDLWriter does not correctly create HDDL files
According to https://unified-planning.readthedocs.io/en/latest/interoperability.html#pddl-mapddl-hddl-reader-and-writer, the PDDLWriter class should be able to write a HDDL problem to disk. 

However, when I try to write a simple HDDL problem to disk (or even print it using the get_domain and get_problem method of the PDDLWriter class), the methods and tasks are not written to file (since there exists no code for it do to so).

arbimo:
This PR extends the `PDDLWriter` to correctly handle hierarchical problem, using the HDDL extension of PDDL. 

fix #506


alvalentini:
LGTM!

PR #580, opened 2024-03-14
Fix new typing errors

Framba-Luca:
This PR adds some type: ignore to ignore the typing problems that appear in the up.model.delta_stn.py file after the usage of an updated version of mypy in the CI


mikand:
I found a simpler solution in #582

PR #588, opened 2024-03-28
Updates for pddl_writer in order to adapt to val syntax

Alelafar:
Added flag for PDDLWriter class and ´:precondition ()´ for actions with no conditions. Similarly for durative actions.
Moreover, we add a new FailedValidationReason due to mutex conflict according to VAL syntax.

### DIFF
--- a/unified_planning/io/pddl_writer.py
+++ b/unified_planning/io/pddl_writer.py
@@ -337,7 +337,7 @@ class PDDLWriter:
     needs_requirements determines if the printed problem must have the :requirements,
     rewrite_bool_assignments determines if this writer will write
     non constant boolean assignment as conditional effects.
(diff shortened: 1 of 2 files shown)


unified_planning/io/pddl_writer.py
```diff
@@ -333,22 +333,27 @@ def walk_equals(self, expression, args):
...
+    action without preconditions instead of writing nothing or similar with conditions in durative actions.
     """
 
     def __init__(
         self,
         problem: "up.model.Problem",
         needs_requirements: bool = True,
         rewrite_bool_assignments: bool = False,
+        *,
+        empty_preco: bool = False,
```
alvalentini, line 350:
```suggestion
        empty_preconditions: bool = False,
```

alvalentini (changes requested):

alvalentini (approved):

PR #604, opened 2024-05-16
Fix ANML parser on MAJSP domain

arbimo:
This PR identifies a parser problem on an existing ANML domain.

The second commit adds a quick and dirty fix for it. However I have no knowledge of how it is actually supposed to work  (and did not really dig into it).
Hence it is likely the case that the problem is deeper that what is handled by the fix, which I really obtained by "randomly" changing things until it stops crashing...

@Framba-Luca could you have a look at it and see if it should be improved/generalized?

### DIFF
--- a/unified_planning/io/anml_reader.py
+++ b/unified_planning/io/anml_reader.py
@@ -944,8 +944,15 @@ class ANMLReader:
                         )
                     else:
                         operator = exp[1]
-                        if isinstance(operator, List):  # parameters list
-                            assert isinstance(exp[0], List) and isinstance(exp[2], List)
+                        if isinstance(operator, List) or isinstance(
+                            operator, ParseResults
+                        ):  # parameters list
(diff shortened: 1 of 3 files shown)


Framba-Luca:
Hi @arbimo , I've looked a bit into it and it seems the correct solution.

Indeed, lower in the code there is this check:
```python
  else:  # expression longer than 3, must be a parameters list
    for e in exp:
        assert isinstance(e, List) or isinstance(
            e, ParseResults
        ), f"expression {exp} is expected to be a parameters list, but it's not"
        pass
```
so parameters can be either `Lists` or `ParseResults`. 

Thanks for finding and fixing this!

arbimo:
@Framba-Luca Thanks for the feedback, I'll ask for it to be merged then ;)

alvalentini (approved):

PR #617, opened 2024-07-11
deps: require numpy version 1.X.Y

arbimo:
The recent release of numpy 2.0 seem to break some of our dependencies.
If you were to re-run the CI on master before this commit, it would fail due to incompatible depenndencies.

This introduces a work around to ensure that we remain on a 1.X.Y version of numpy. I expect this to be removable once the ecosystem fixes their required dependencies.﻿

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -8,3 +8,4 @@ jupyter
 tarski
 networkx
 pyparsing
+numpy<2  # work around to avoid incompatibilities in transitive dependencies
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 pyparsing
 networkx
 ConfigSpace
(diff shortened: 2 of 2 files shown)


mikand:
Thanks for this Arthur. Did you discover whcih pair of deps caused this (so that we can monitor them)?

arbimo:
No, but @nicola-gigante is I think the one who delved the most into this issue and may have more insights.

Note that we faced the same problem also for doc generation, where the dependencies are quite different.

nicola-gigante:
@mikand if you have access to the old CI logs, you should see the name of the offending module in the Python stack trace, I think.

alvalentini (approved):
Looks good also to me to keep these changes until the dependencies will be fixed!

PR #619, opened 2024-07-15
Fix bug in TimeTriggeredPlanValidator

alvalentini:


mikand:
LGTM, but codecov complains that a test is missing. Can we add one no-regressin testcase?

mikand (approved):
LGTM!

PR #620, opened 2024-07-18
Update pddl_planner.py - added errors='ignore' in byte string decode …

VladKlimen:
…in solve_

I tried to run some solver and the only problem was the exception that was thrown because of encoding failure of out_err_bytes, that anyway goes to logs and not critical to lack some non-unicode characters

### DIFF
--- a/unified_planning/engines/pddl_planner.py
+++ b/unified_planning/engines/pddl_planner.py
@@ -183,7 +183,7 @@ class PDDLPlanner(engines.engine.Engine, mixins.OneshotPlannerMixin):
                 proc_err: List[str] = []
                 try:
                     out_err_bytes = process.communicate(timeout=timeout)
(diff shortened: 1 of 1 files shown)


mikand:
Hi @VladKlimen,

thanks for this PR and for signaling the issue. I was wandering if ignoring the non-utf-8 characters is a good idea in general: Python offers other ways such as replacing: https://docs.python.org/3/library/codecs.html#error-handlers. Having pieces of outputs magically disappear from the logs might not be great, maybe replacing is a better option?

@arbimo @roeger Any thoutghs on this?

arbimo:
Hi, 
I am not particularly familiar with this part of the code but in general in would agree with Andrea and favor replacing the character with UTF-8 ? (�), as supported by the python function. For me that immediately makes the problem obvious and suggests an encoding problem.

PR #629, opened 2024-09-18
Autonomous processes events

hstairs:
In this branch we add support for PDDL+ specification. The branch includes extensions to the problem representation (so allowing to specify processes and events), pddl writer (so allowing to use domain independent off-the-shelf PDDL+ solvers), pddl reader (so allowing get PDDL+ specifications into up data structures). There is some design choice that would need discussion though.


unified_planning/model/action.py
```diff
@@ -764,3 +764,228 @@ def __repr__(self) -> str:
...
+"""
+Below we have natural transitions. These are not controlled by the agent and would probably need a proper subclass. Natural transitions can be of two kinds:
+Processes or Events.
+Processes dictate how numeric variables evolve over time through the use of time-derivative functions
+Events dictate the analogous of urgent transitions in timed automata theory
+"""
+
+
+class Process(Action):
+    """This is the `Process` class, which implements the abstract `Process` class."""
```
mikand, line 778: ? It seems to me it implements the abstract `Action` class :smile:

I thinkit is not too bad to have processes and events to be `Action`s, even though they are semantically different. In hindsight we might have called the abstract action class somethig else.
@alvalentini maybe we can consider a refactoring of this name, eventually...

unified_planning/model/action.py
```diff
@@ -764,3 +764,228 @@ def __repr__(self) -> str:
...
+    def add_derivative(
+        self,
+        fluent: Union["up.model.fnode.FNode", "up.model.fluent.Fluent"],
+        value: "up.model.expression.Expression",
+    ):
+        """
+        Adds the given `derivative effect` to the `process's effects`.
+
+        :param fluent: The `fluent` objective of the time derivative definition.
+        :param value: The given `fluent` is incremented by the given `value`.
```
mikand, line 928: This comment is wrong, and I think that how we model derivatives is a central issue that shall be very well documented.

unified_planning/model/action.py
```diff
@@ -764,3 +764,228 @@ def __repr__(self) -> str:
...
+        new_instantaneous_action._preconditions = self._preconditions[:]
+        new_instantaneous_action._effects = [e.clone() for e in self._effects]
+        new_instantaneous_action._fluents_assigned = self._fluents_assigned.copy()
+        new_instantaneous_action._fluents_inc_dec = self._fluents_inc_dec.copy()
+        new_instantaneous_action._simulated_effect = self._simulated_effect
+        return new_instantaneous_action
+
+    @property
+    def preconditions(self) -> List["up.model.fnode.FNode"]:
+        """Returns the `list` of the `Action` `preconditions`."""
```
mikand, line 864: All the comments in this class are copy-pasted, we need to replace `Action` with `Process`

unified_planning/model/action.py
```diff
@@ -764,3 +764,228 @@ def __repr__(self) -> str:
...
+                fluent_exp,
+                value_exp,
+                condition_exp,
+                kind=up.model.effect.EffectKind.INCREASE,
+                forall=tuple(),
+            )
+        )
+
+
+class Event(InstantaneousAction):
```
mikand, line 960: This is kind of dangerous: beware that `isinstance(x, InstantaneousAction)` would return `True` if x is an Event. I get that we save some code duplication, but here I would inherit from `Action` only... Then, maybe we can create a mixin to avoid the code duplication

mikand: For example, https://github.com/aiplan4eu/unified-planning/blob/master/unified_planning/model/mixins/actions_set.py#L60 is probably broken by this choice

unified_planning/model/problem.py
```diff
@@ -331,6 +331,10 @@ def _get_static_and_unused_fluents(
                     unused_fluents.clear()
                     for f in se.fluents:
                         static_fluents.discard(f.fluent())
+            elif isinstance(a, up.model.action.Process):
```
mikand, line 334: What about events here? Again, they are probably absorbed by the `isinstance(a, up.model.action.InstantaneousAction)`, so we need to fix this

unified_planning/model/problem.py
```diff
@@ -682,6 +686,9 @@ def _kind_factory(self) -> "_KindFactory":
         for goal in chain(*self._timed_goals.values(), self._goals):
             factory.update_problem_kind_expression(goal)
         factory.update_problem_kind_initial_state(self)
+        self.processes
+        if len(list(self.processes)) > 0:
+            factory.kind.set_time("PROCESSES")
```
mikand, line 691: Add also some flags for events!

unified_planning/model/problem.py
```diff
@@ -1003,6 +1010,8 @@ def update_problem_kind_action(
             if len(action.simulated_effects) > 0:
                 self.kind.set_simulated_entities("SIMULATED_EFFECTS")
             self.kind.set_time("CONTINUOUS_TIME")
+        elif isinstance(action, up.model.action.Process):
+            pass
```
mikand, line 1014: I think we should add `PROCESSES` to the kind

unified_planning/model/problem_kind.py
```diff
@@ -370,6 +371,12 @@ def intersection(self, oth: "ProblemKind") -> "ProblemKind":
 basic_temporal_kind.set_typing("FLAT_TYPING")
 basic_temporal_kind.set_time("CONTINUOUS_TIME")
 
+process_temporal_kind = ProblemKind(version=LATEST_PROBLEM_KIND_VERSION)
+process_temporal_kind.set_problem_class("ACTION_BASED")
+process_temporal_kind.set_time("PROCESSES")
+process_temporal_kind.set_time("CONTINUOUS_TIME")
```
mikand, line 377: no FLAT_TYPING here?

mikand (changes requested):
@hstairs Thanks fo this, it looks good but there are few things I would do before merging:
1. Add some tests in the examples showing the new feature
2. Add EVENTS to the kind and handle it
3. Do not inherit the `Event` class from `InstantaneousAction`, in many places we check with `isinstance(x, InstantaneousAction)`
4. I am ok with limiting ourselves with linear derivatives for now, but I would design the interface of processes to be generic, otherwise it is very hard to change it in the future. Also, the parser and the code should raise meaningful exceptions when a non-linear case is encountered, now they operate wrongly.

unified_planning/model/action.py
hstairs: One could call this transition, and I have hinted at something like that to Alessandro...but it is a major change no?

unified_planning/model/action.py
hstairs: Yep

unified_planning/model/action.py
hstairs: yep

unified_planning/model/problem.py
hstairs: yep

unified_planning/model/problem.py
hstairs: not sure we need to differentiate processes and events. I think a flag (likely not "processes") for both is fine

unified_planning/model/problem.py
hstairs: yep, connected with the previous

unified_planning/model/problem_kind.py
hstairs: Isn't FLAT_TYPING a restriction on the kind of types we can define (like no hierarchy)? If so, there is no restriction here, or I didn't understand this :)

unified_planning/io/pddl_reader.py
```diff
@@ -526,6 +552,8 @@ def _parse_exp(
                         solved.append(self._em.FluentExp(problem.fluent(exp.value)))
                     elif problem.has_object(exp.value):  # object
                         solved.append(self._em.ObjectExp(problem.object(exp.value)))
+                    elif exp.value == "#t":
+                        solved.append(self._em.Int(1))
```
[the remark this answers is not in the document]
hstairs: Having said that there is definitely a cleaner way to parse that!

unified_planning/model/problem.py
hstairs: One further comment: if we only have events we need to find a way to understand the semantics. I am not sure this case is well documented in the syntax of PDDL+. For instance, ENHSP considers temporal only a problem where there is at least one process. I don't see any use for events only (as they are understood in PDDL+) as events make sense if triggered by something which is changed as an effect of some passage of time.

Ezpeon:
Hello, I started looking a bit into this and I think I could help a bit with the refactoring that was previously mentioned in the other comments. 
I started another PR #646 where I will try to better organize the features (still work in progress, it does not currently pass mypy checks and other changes are needed, but pytest tests should still work). 
I will let you know when it is in a state where it would make sense to look more into it to decide if it is worth it to adopt the changes I made.

alvalentini (approved):
@hstairs For me now, it is ready to be merged! Can you check?

PR #669, opened 2025-03-21
wip(opt-sched): test interface for optional activites

arbimo:


unified_planning/model/presence.py
```diff
@@ -0,0 +1,33 @@
...
+    TimeExpression,
+)
+from abc import ABC
+from enum import Enum, auto
+from fractions import Fraction
+from typing import Union, Optional
+
+class Presence:
+    def __init__(self, container: str):
+        self._container = container
```
alvalentini, line 15: @arbimo Could the container be an Activity instead of a str here? Or you choose str for a specific reason?

arbimo: I have not given it too much thought about it and used the same pattern as for the timepoint expression (that also refers to a container), optimistically assuming I did a good choice at the time.

From what I remember the use of the string identifier was notably motivated by the avoidance of cyclic dependencies between the modules. 
Extensiblity is another one. For instance, the timepoint expression might either refer to a task (HTN) or an activity (scheduling)

I agree it is weak and a potential footgun and would be very pleased with a more type-safe solution.

unified_planning/test/examples/scheduling/flexible_jobshop.py
```diff
@@ -0,0 +1,148 @@
...
+
+            # exactly one activity constraint
+            problem.add_constraint(Or(act.present for act in task_activities[t]))
+            for i1, act1 in enumerate(task_activities[t]):
+                for i2, act2 in enumerate(task_activities[t]):
+                    if i1 != i2:
+                        # act1.add_constraint(Not(act2.present))
+                        # problem.add_constraint(Not(act2.present), scope=[act1.present])
+                        # problem.add_constraint(Implies(act1.present, Not(act2.present)))
+                        problem.add_constraint(Or(Not(act1.present), Not(act2.present)))
```
DavideLusuardi, line 125: @arbimo I have installed the latest version of Aries (branch `up-optional-activities`) and encountered some issues while using it.
- When I define the constraint `act1.add_constraint(Not(act2.present))` or `problem.add_constraint(Not(act2.present), scope=[act1.present])`, the plan generated by Aries includes both optional activities `job0_task0_machine3` and `job0_task0_machine1`.
- When I use the constraint `problem.add_constraint(Implies(act1.present, Not(act2.present)))`, Aries returns an `INTERNAL_ERROR`.
- When I define the constraint `problem.add_constraint(Or(Not(act1.present), Not(act2.present)))`, the plan appears to be correct.

Additionally, when I attempt to validate the plan using Aries as a validator, it reports that the plan is invalid.

I'm not sure if I made a mistake. Any insights would be greatly appreciated.
Thank you!

arbimo: Hi Davide,

this is most likely to the support in Aries itself being very preliminary, with many corners cut.
I'll try to fix it by the end of the week.

>     * When I define the constraint `act1.add_constraint(Not(act2.present))` or `problem.add_constraint(Not(act2.present), scope=[act1.present])`, the plan generated by Aries includes both optional activities `job0_task0_machine3` and `job0_task0_machine1`.

That's definitely a bug due to the incomplete implementation.

>     * When I use the constraint `problem.add_constraint(Implies(act1.present, Not(act2.present)))`, Aries returns an `INTERNAL_ERROR`.

This is most likely an unimplemented expression (implies?) you can check the logs of the planner to verify this. 
By default, (when no redirection of stdout/stderr is specified), Aries will save its log in a temporary file /tmp/aries-XXXX)

arbimo: @DavideLusuardi I have updated the corresponding branch and the problem should not be present anymore. All variants seemed to work. I'll try to validate on a more diverse set of problems, but not having a validator is a bit annoying :/


By the way if you want a baseline for comparison (notably to get the optimal solution cost), aries has "native" support for the flexible jobshop and should have state-of-the-art performance.

```
cargo run --release --bin scheduler -- fsp examples/scheduling/instances/flexible/bc/mt10c1.fjs
```
Should give a proved optimality of 927 in less than a second. Going through the UP, we are very far from this performance but it should be approachable once I get some time into optimizing the encoding (there is nothing particularly fancy in the native encoding). That's something I hope to do next month bu tin the mean time focusing on correctness is probably a better objective

arbimo:
FYI, I have added a alternative encoding for FSP that prefers direct disjunctive precedence constraints instead of statements on resources. The two are equivalent (useful for testing) but the new one should be much easier to solve, at least for Aries.

alvalentini:
Hi @arbimo! The PR looks good overall. There are just a few minor TODOs and some debug code to remove. It would be great to get this merged into `master`. What do you think? Do you have time to finalize it?

PR #677, opened 2025-05-05
Feat: Introduce Warm Start capability via `solve()` and `get_solutions()`

Shi-Raida:
## Summary

This pull request introduces the ability for users to provide an initial "warm start" plan to the search engine. This is achieved by adding an optional `warm_start_plan` parameter to the `solve()` and `get_solutions()` methods.

It's crucial that this warm start plan is provided per-request, as different `solve()` calls might require different starting points. Therefore, this parameter is added to the `solve()` / `get_solutions()` methods rather than being a planner configuration setting.

## Implementation Details

The implementation is divided into two main parts:

1.  **General Mechanism for Per-Solve Parameters (Commit 22b6d5b39f1f89c352305e274573b33417fc943f):**
    *   To support `warm_start_plan` and potentially other future per-solve parameters without breaking existing interfaces, a more flexible parameter passing mechanism was introduced.
    *   `solve()` and `get_solutions()` now accept arbitrary keyword arguments (`**kwargs`).
    *   New internal methods `_solve_with_params()` and `_get_solutions_with_params()` are introduced. Planners wishing to consume these additional parameters should override these new methods.
    *   These new methods receive the standard arguments plus the `**kwargs` dictionary containing any extra parameters passed to the public methods.

2.  **Addition of `warm_start_plan` Parameter (Commits 624b597b2f7b85ffd9ad84ecfa9ddcf5aa39304c, cdb29d75b3aa4a68b5d1d152fb041d47a001e3c1):**
    *   The `warm_start_plan` parameter definition is added to the relevant Protobuf messages.
    *   The public `solve()` and `get_solutions()` methods now explicitly check for the `warm_start_plan` keyword argument.
    *   The interval `_solve_with_params()` and `_get_solutions_with_params()` methods also explicitly check for the `warm_start_plan` keyword argument.
    *   If provided, `warm_start_plan` is included in the `**kwargs` passed down to `_solve_with_params()` / `_get_solutions_with_params()`, making all versions of the planner compatible with all future versions of UPF.

## Backward Compatibility

*   This change is fully backward compatible.
*   The base implementations of `solve()` and `get_solutions()` check if the planner overrides `_solve_with_params()` / `_get_solutions_with_params()`.
*   If the new methods are *not* overridden, the base methods automatically call the original `_solve()` / `_get_solutions()` methods, discarding any extra `**kwargs`.
*   Existing planners that do not need warm start capabilities require no modification and will continue to function as before.

## Future Considerations

This `**kwargs`-based approach provides a flexible foundation for adding other solve-specific parameters in the future without requiring disruptive signature changes across all planner implementations.


alvalentini:
Hi @Shi-Raida! Thanks for this contribution!
I like the idea of having a general mechanism for per-solve parameters. However, do we really need `warm_start_plan` as an explicit parameter? Could it instead be handled through `**kwargs`?

arbimo:
Hi @alvalentini 

For context: I have been working with @Shi-Raida on this feature and thus it would be better if I am not the one reviewing it.

I think the addition of per-solve arguments is indeed necessary and it is something we probably should have introduced earlier. A large part of the work there is to introduce it in a way that does not break existing planners integration.

Regarding the `warm_start_plan` argument, I would argue that it is in fact highly desirable. It is a feature that is ubiquitous in combinatorial optimization (MILP, CP, ...) and is extremely useful in substantially improving the performance of a solver by providing it with an initial (non-optimized) solution.
We have been experimenting with Aries to support a similar feature and we saw similar gains: using a fast and coverage oriented planner to provide warm-start and then using Aries for plan optimization gives us that are of substantially higher quality than what the original planner provided, while keeping a very high coverage and excellent time to the first solution.

In a library like unified planning, where one of the strength is the ability to combine  different planners in a single framework I think this would be a very desirable feature.

One of the limitation I see is that planners that support this kind of features are rare at the moment (Aries, LPG, ?). However I think it is a feature that is worth pushing: I suspect that with the advent of LLM and the resulting ease to create greedy domain-specific solvers, the gains of supporting optimization from initial solution will become more important.

Note: we have support for this already implemented in Aries (with release awaiting for upstream support). 

@Shi-Raida If we agree on the mechanism, the docs should be updated to present both the per-solve arguments and the warm-start one.

alvalentini:
Hi @arbimo! 

I agree that the per-solve arguments is something needed and we probably should have introduced earlier.

Regarding the `warm_start_plan` argument, I see a strong relationship with the `PlanRepairer` operation mode (that is supported by LPG). How do you see this relationship?

arbimo:
Indeed, there are some similarities. The main differences are that:

In `PlanRepairer`:
  - the plan is required
  - the plan is in general expected to be invalid (though it is not a strict requirement)
  - the interface is centered on satisfiability
 
With warm-starting:
 - the plan is expected to be valid
 - the plan is optional and just a hint and can be ignored by the planner
 - the interface is centered on optmization

Another (non-exclusive) option would be to have a `PlanOptimizer` operation mode where:
 - the plan must be provided and must be valid
 - the planner should only provide strictly better solution (but maybe we would like again another operation mode for requiring only an *optimal* solution?)

I kind of like warm-starting for the flexibility it provides, (giving additional help to the planner, without changing the sematincs of the method) and the fact that it is already widely supported in other fields.
A `PlanOptimizer`interface would probably be a nice addition as well but serves a slightly different purpose.

Shi-Raida:
Hi @alvalentini 

I just added presentation of the per-solve parameters on the `OneshotPlanner` documentation.

I can also add a presentation for the `warm_start_plan` parameter if we agree to explicitly expose it in the method signatures.

alvalentini:
Hi @Shi-Raida, 
After reading @arbimo's comment, I see the value of making the `warm_start_plan` parameter explicit.
It would be great if you could update the documentation to reflect this.

Additionally, it would be useful to update the [notebook](https://github.com/aiplan4eu/unified-planning/blob/master/docs/notebooks/tutorial/planner-integration.ipynb) to use the new `_solve_with_params` method.

Thanks!

Shi-Raida:
Hi @alvalentini,
I explicitly listed all per-solve parameters in the documentation (`heuristic`, `timeout`, `output_stream`, and `warm_start_plan`), and updated the notebook on the two commits above.

alvalentini (approved):
The PR looks good to me and is ready to be merged. @arbimo feel free to proceed!

arbimo (approved):

arbimo:
It looks good to me as well. But I am a bit worried that the CI wouldn't run on it (and the CI on the forked repo does not seem to pass).
There is no way for me to allow the workflow to run, which should be possible for external contributors.

I suspect the culprit is that we enable CI on push, which would only work for branches on this repository.

@Shi-Raida Could you add the `pull-request` trigger on the CI's `main.yml` ?

arbimo:
You can probably remove the requirement on the codecov secret (the corresponding job is only run on this repository)

arbimo:
Merged! @Shi-Raida thanks for polishing the contribution!

PR #678, opened 2025-05-06
UPState: add a raise if the values in the state are not constants

Framba-Luca:


unified_planning/model/state.py
```diff
@@ -64,7 +64,17 @@ def __init__(
                 f"The max_ancestor field of a class extending UPState must be > 0 or None: in the class {type(self)} it is set to {type(self).MAX_ANCESTORS}"
             )
         self._father = _father
-        self._values = values.copy()
+        self._values = {}
+        for fluent, value in values.items():
+            if not fluent.is_fluent_exp():
+                raise UPValueError(
+                    f"The fluent {fluent} is not a fluent, but a {type(fluent)}"
```
alvalentini, line 71: `type(fluent)` is a `FNode`. The message is not clear

unified_planning/model/state.py
```diff
@@ -64,7 +64,17 @@ def __init__(
...
+                )
+            if not value.is_constant():
+                raise UPValueError(
+                    f"The value {value} is not a constant, but a {type(value)}"
```
alvalentini, line 75: Same

alvalentini (changes requested):
@Framba-Luca

alvalentini (approved):

PR #679, opened 2025-05-08
Fix issues for unfactored ma-pddl writer

Olli1080:
Fix determination of constants -> also fixes objects 
Fix missing ? in front of agent


Olli1080:
You're welcome. Thank you for this project.
I have added myself as you requested in a new PR.

PR #687, opened 2025-07-18
Fix ProblemKind versioning

alvalentini:


alvalentini:
We have an issue with the current versioning management between the UP library and the engine. Specifically, the way versions and features are handled can lead to incorrect compatibility checks.

Below is a summary of the situation in the table:

| **Feature Situation**                          | **UP version < Engine version**                                                                        | **UP version > Engine version**                                                     |
|------------------------------------------------|------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------|
| **New features introduced in the newer version** | Engine is downgraded (loses new features); UP stays unchanged. <br> *Compatibility evaluated correctly.** | UP is downgraded (new features removed); Engine stays unchanged. <br> *New features ignored → false positive compatibility.* |
| **Features removed in the newer version**         | Engine is downgraded (removed features remain absent); UP stays unchanged. <br> *Compatibility evaluated correctly.* | UP is downgraded (removed features remain absent); Engine stays unchanged. <br> *Compatibility evaluated correctly.* |
| **Features replaced by others in the newer version** | Engine is downgraded (replaces features); UP stays unchanged. <br> *Compatibility evaluated correctly.**    | UP is downgraded (replaces features); Engine stays unchanged. <br> *Compatibility evaluated correctly.*                      |

*When UP version < Engine version, there is a problem with new features: if the engine tries to use features not supported by the older UP library, it will cause errors. Therefore, either:
- The engine must check the UP library version and only enable compatible features,
- Or the engine must bump its version so that the newer engine version is not considered compatible with the older UP library version.

To solve the problem when UP version > Engine version (which is currently causing PR test failures), instead of downgrading the UP library features, the better approach would likely be to perform an upgrade of the engine features to match the newer UP version.

mikand (approved):

arbimo:
Hi,

I am not sure I fully grasp the core of the problem. 
My understanding is that features should *not* be removed or replaced in the UP. But may be complemented by new more specific ones. For instance, we had the `CONTINUOUS_NUMBERS` which got split into `REAL_FLUENT` and `REAL_NUMBER_IN_ACTION_COSTS`, ...

**Assumption** : 
 - A planner that supports one of the more specific one should declare support for the more general one and the UP should include it in the problem features (for compatibility).
 - The UP continues to emit the old "catch all" feature

Under these assumptions:

 - If `UP_VERSION` < `ENGINE_VERSION, then there cannot be unsupported features in the problem and everything should be fine. We should ignore any new feature declared by the planner (the ones that are not present in the UP and thus cannot be produced by it)
- If `UP_VERSION` > `ENGINE_VERSION`, then there should be only two possibility:
   - the new "feature" is just a more specific characterization of a previous one. In this case, we should just drop it. This may cause some false positive but it is inevitable : the purpose of this feature is precisely to catch those false positive by allowing more fine-grained characterization of the feature set. Without it we can not hope to characterize it.
   - the feature is a true new feature. In this case it should be kept : the planner explicitly declares not supporting it by saying 

Did I get that right ? Not that the assumptions are what I think demanded from the planners and UP but may not actually hold in practice.

alvalentini:
In the case `UP_VERSION` > `ENGINE_VERSION`, I fully agree with you. However, the code currently in master does not follow this logic, particularly in the case "the feature is a true new feature". This PR aligns the implementation with that idea.

arbimo:
Thanks for the clarification, now I understand the logic behind the change! 
I saw no problem with the revised version and will merge it now. Thanks for fixing this!

Issue #689, opened 2025-07-24
SequentialSimulator + UPState terribly inefficient for large predicate / function arities


quickbeam123:
`SequentialSimulator` creates the initial state from `_problem.initial_values`, a property implemented via:
```
"""
        Gets the initial value of all the grounded fluents present in the `Problem`.

        IMPORTANT NOTE: this property does a lot of computation, so it should be called as
        seldom as possible.
        """
        res = self._initial_value
        for f in self._fluent_set.fluents:
            for f_exp in get_all_fluent_exp(self._object_set, f):
                value = self.initial_value(f_exp)
                if value is not None:
                    res[f_exp] = value
        return res
```

The upshot is that even ground predicates false in the initial state are made explicit!

This is detrimental already on some standard IPC benchmarks. Consider, e.g., `nomystery`, where an "addition table" is spelled out with initial facts such as `(sum level199 level0 level199)`. If 20K positive facts are specified this way, `Problem.initial_values` will go and explicitly list all the remaining ones as negative, making the total 8 million!

The solution is simple: 1) only store what is true in state (everything is implicitly false), 2) update `UPState` (mainly its `make_child` and `get_value` accordingly). This can be easily extended to numeric fluents, by tapping into `problem.fluents_defaults` or simply assuming the default is `0`. (We could then add `supported_kind.set_initial_state("UNDEFINED_INITIAL_NUMERIC")` to `SequentialSimulator` and its little slave `Grounder`.)

I can try putting together a PR with the corresponding fix. 

My question is: Would there be an interest in such performance improvement? Or am I missing some more advanced use case for `SequentialSimulator`/`UPState` where the explicit negative facts are needed for something?

quickbeam123:
I will have time to proceed next week. I will propose a destructive update which works for me for classical (numeric) planning and then let's see whether anything could break with any other potential use cases of `SequentialSimulator`/`UPState`, which I am perhaps not aware of.

PR #694, opened 2025-08-29
Implememt SparseState, a more efficient version of UPState

quickbeam123:
Fixes aiplan4eu/unified-planning#689 where additional details are already explained.

In a nutshell, `SparseState` does not need every ground fact to be explicitly represented, which is important especially with fluents with large arities (thus large groundings) from which only very few instances are true in every reachable state.

On the IPC benchmark `nomystery/p07.pddl` (Problem name: `transport-l10-t1-p9---int100n150-m25---int100c150---s1---e0`), the original `SequentialSimulator` takes 85.8 seconds (on my laptop) just to return from `get_initial_state` whereas with `SparseState` this goes down to 0.05s.

Note: arguably, the old `UPState` could be replaced by the new one, instead of keeping both.

### DIFF
--- a/unified_planning/model/state.py
+++ b/unified_planning/model/state.py
@@ -19,7 +19,8 @@ from operator import xor
 from typing import Dict, List, Optional, Tuple
 import unified_planning as up
 from unified_planning.exceptions import UPUsageError, UPValueError
-
+from unified_planning.model.mixins import FluentsSetMixin
+from unified_planning.environment import get_environment
 
 class State(ABC):
     """This is an abstract class representing a classical `State`"""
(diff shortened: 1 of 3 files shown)


unified_planning/model/state.py
```diff
@@ -159,3 +160,102 @@ def make_child(
...
+                )
+            if self._is_nondefault(fluent,value):
+                self._values[fluent] = value
+
+        self._hash: Optional[int] = None
+
+    def _is_nondefault(self, fluent: "up.model.FNode", value: "up.model.FNode"):
+        return (fluent.fluent() not in self._fluent_set.fluents_defaults or
+                self._fluent_set.fluents_defaults[fluent.fluent()] != value or
+                fluent.type.is_bool_type() and value.is_true())
```
alvalentini, line 201:
```suggestion
                self._fluent_set.fluents_defaults[fluent.fluent()] != value)
```

if `true` is the correct default for the fluent, it is already in the `fluents_defaults` dict

alvalentini (commented):
I like the idea of the `SparseState`, but I would suggest modifying the `UPState` so that it can optionally take a `FluentSetMixin` in the constructor.
If `None` is provided, the behavior stays the same as the current `UPState`; otherwise, it uses the defaults in the same way as `SparseState` does.
This way, we could just have the `UPState` class.

quickbeam123:
> Would you be available to merge the two implementations?

I can give it a shot next week.

About the failing tests: what's the best way to make sure locally that they will succeed on GitHub?

alvalentini:
To check the code locally, you mainly need to run these three commands:
- `python3 -m black --check --exclude=unified_planning/grpc/generated/ .` with `black==22.6.0`
- `python3 -m mypy unified_planning/`
- `python3 -m pytest unified_planning/tests/`

unified_planning/engines/compilers/usertype_fluents_remover.py
```diff
@@ -422,7 +422,7 @@ def new_fun(
                             assert (
                                 obj_exp == test_value
                             ), "Error, found True Value multiple times in the same state for a boolean fluent used to remove a UserType fluent"
-            state = UPState(original_state)
+            state = UPState(original_state,original_problem)
             # populate the ret_val list with the default returned value, when a non
```
quickbeam123, line 426: Not 100% sure about this one.

quickbeam123:
So, the two implementations are merged. Please let's have a look at the official tests.

From running locally, `pytest` and `black` should be fine. With `mypy` (v `1.17.1`) I am getting a bunch of errors even for master (and even when forcing it to read your `--config-file .mypy.ini`). Not sure what to think of that.

PR #734, opened 2026-03-23
test: add failing test for real fluent not shown in problem kind

arbimo:
@alvalentini @Framba-Luca This is more of a bug report than a complete pull request, but fixing it would certainly benefit from your expertise on this part of the library =)

In essence there is one problem in our test suite (`"robot_with_variable_duration"`) that has a real fluent but for which the corresponding feature is not detected.
My intuition is that this happens because the fluent is not used in the problem itself.

This PR just adds a failing test for this.


arbimo:
Oops. Made real sense at the time you answered but I forgot to close the PR!

Thanks for the prompt reply and sorry for the long one ;)

PR #736, opened 2026-03-25
build: swtich project management to uv

### LINKED ISSUE #553: Generalize the use of `just`
In PR #550 we introduced `just` (https://github.com/casey/just) for docuemntation and to simplify the main dev commands.
We should consider to generalize its use for CI and other comamnd-line tasks.

arbimo:
This is a tentative PR to switch project management from `pip+manual` to `uv`.

The main motivation behind this is that testing in the current setting is really hard: many of the tests require dependencies that are only specified in the CI configuration and getting a complete test environment locally is far from trivial.

The rational behind this PR:
 - have all dependencies (regular, optional, dev, dev-optional) declared in `pyproject.toml`
 - have all regular operations (formating, type-checking, testing, ...) be defined as commands in the justfile
   - those use uv internally to make sure they are invoked in the appropriate environment
 - have the CI just invoke the regular just commands (making it trivial to reproduce locally)

At this point this PR is mostly a proof-of-concept but, at least for my workflow, is already a great improvement on the status-quo =)

Things that would be required for this to be merged:

- [ ] complete planner dependencies (only a few were added)
  - [ ] check that all tests are run (some are currently skipped, presumably due to missing dependencies)
- [ ] check dependencies, those where added automatically from the requirements and some maybe relaxed (lower version or removed)
- [ ] Complete CI setup, I only translated a few jobs (but the fact that they all succeeded on the first try makes me happy)
- [ ] consider what would need to be done for building and distributing releases. This is allowed by `uv` but setting a version dynamically does not seem to be easy. 
  - (do we still distribute `--pre` packages ? if not this could be drastically simplified)


alvalentini:
See #766

PR #759, opened 2026-07-21
fix: Parse goal-prefixed ANML identifiers

### LINKED ISSUE #742: ANML I/O: Parsing error for identifiers starting with the 'goal' keyword
I have identified a bug in the ANML reader when using fluents or constants with names starting with goal (e.g., `goal_X`).

When a problem is dumped into an ANML file, constants are initialized as simple assignments without a temporal qualifier, such as: `goal_X := 4;`.

However, the ANML reader fails to parse this back in. Since 'goal' is a reserved keyword, the parser matches it and treats the remainder of the identifier (`_X`) as an invalid expression, resulting in a `NotImplementedError`.

The error does not occur if the assignment is manually prefixed with a temporal qualifier, for example:
`[ start ] goal_X := 4;`.
In this case, the parser seems to handle the identifier correctly, as well as if the identifier doesn't start with 'goal', i.e. `ggoal_X:=4;`.

Sanjays2402:
Closes #742

ANML parsed the `goal` declaration token as a prefix of identifiers such as `goal_X`, causing valid assignments to fail. Using a keyword boundary preserves `goal` declarations while allowing goal-prefixed identifiers; the regression is covered by the ANML reader tests.

### DIFF
--- a/unified_planning/test/test_anml_reader.py
+++ b/unified_planning/test/test_anml_reader.py
@@ -67,6 +67,13 @@ class TestANMLReader(unittest_TestCase):
         problem_2 = reader.parse_problem_string(problem_str, problem_filename)
         self.assertEqual(problem, problem_2)
 
+    def test_identifier_starting_with_goal_keyword(self):
+        problem = ANMLReader().parse_problem_string(
+            "constant integer goal_X; goal_X := 4;", "issue_742.anml"
+        )
+
+        self.assertEqual(problem.initial_value(problem.fluent("goal_X")), Int(4))
+
     def test_match_reader(self):
         reader = ANMLReader()
 
--- a/unified_planning/test/test_anml_reader.py
+++ b/unified_planning/test/test_anml_reader.py
@@ -67,12 +67,62 @@ class TestANMLReader(unittest_TestCase):
         problem_2 = reader.parse_problem_string(problem_str, problem_filename)
         self.assertEqual(problem, problem_2)
 
-    def test_identifier_starting_with_goal_keyword(self):
+    def test_identifiers_starting_with_keywords(self):
+        # Identifiers that merely start with a keyword must be read as a single
+        # name, not split into the keyword plus the rest of the name.
+        names = [
+            "goal_X",
+            "goalkeeper",
+            "typewriter",
+            "instanceof",
+            "constantly",
+            "fluently",
+            "actionable",
+            "whenever",
+            "durationless",
+            "startup",
+            "ending",
+            "notation",
+            "andromeda",
+            "orbit",
+            "truebool",
+            "falsehood",
+            "integerish",
+            "booleanish",
+            "floating",
+            "setting",
+            "allocation",
+            "infinityish",
+        ]
+        for index, name in enumerate(names):
+            with self.subTest(name=name):
+                problem = ANMLReader().parse_problem_string(
+                    f"constant integer {name}; {name} := {index};",
+                    "issue_742.anml",
+                )
+                self.assertEqual(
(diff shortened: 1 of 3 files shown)


alvalentini:
Thanks for the contribution! The fix is correct, but the problem is actually more general. It's described for `goal`, but the same thing happens with any other keyword used as a bare literal in the grammar. For example a fluent named `typewriter` gets read as type writer, and the same applies to `instance`, `constant`, `fluent`, `action`, `when`, and so on.

It would be good to handle all of them at once by applying `Keyword(...)` to the keyword matches instead of just goal. Would you be up for extending this PR to handle them?

Sanjays2402:
Sure, happy to extend it. You're right that scoping it to `goal` just papers over the one report, `typewriter` tokenising as `type writer` is the same bug wearing a different hat.

I'll wrap the keyword matches in `Keyword(...)` across the grammar rather than one-off, and add cases covering a few of them (`instance`, `constant`, `fluent`, `action`, `when`) so the regression test isn't just about `goal` either. Will push an update to this PR.

Sanjays2402:
Pushed the general fix.

Instead of patching `goal` alone, the keyword matches now go through a small `keyword()` helper that wraps them in pyparsing `Keyword(...)`, applied across the grammar: `type`, `instance`, `constant`, `fluent`, `action`, `goal`, `when`, `duration`, `forall`/`exists`, `and`/`or`/`xor`/`not`/`implies`, `true`/`false`, and the primitive type names (`boolean`, `integer`, `float`/`rational`, `infinity`). Symbolic tokens (operators, brackets, `:=`) stay on `one_of`, since word boundaries don't mean anything there.

Also dropped the incidental `Suppress("action")` string literal in favour of `TK_ACTION`, so that one isn't spelled twice.

The test is no longer goal-specific: it now checks a list of identifiers prefixed by the various keywords (`typewriter`, `instanceof`, `constantly`, `fluently`, `actionable`, `whenever`, `notation`, `orbit`, ...) parse as single names, plus a second case exercising keyword-prefixed type, object, fluent and action names in one problem. I checked each of the new names does fail on master, so the test would have caught this.

`test_anml_reader.py` is 18 passed / 22 subtests passed locally, and `test_anml_writer.py` is still green.

unified_planning/test/test_anml_reader.py
```diff
@@ -67,6 +67,63 @@ def test_basic(self):
...
+            };
+            [end] fluently(instanceof) == true;
+            """,
+            "keyword_prefixed_names.anml",
+        )
+
+        self.assertEqual(problem.user_type("typewriter").name, "typewriter")
+        self.assertEqual(problem.object("instanceof").name, "instanceof")
+        self.assertEqual(problem.fluent("fluently").name, "fluently")
+        self.assertEqual(problem.action("actionable").name, "actionable")
```
alvalentini:
```suggestion
        self.assertTrue(problem.has_type("typewriter"))
        self.assertTrue(problem.has_object("instanceof"))
        self.assertTrue(problem.has_fluent("fluently"))
        self.assertTrue(problem.has_action("actionable"))
```

alvalentini (changes requested):
Can you also add yourself in the CONTRIBUTORS file?

unified_planning/test/test_anml_reader.py
Sanjays2402: Applied, and added myself to CONTRIBUTORS in the same commit (b3b757b).

Full test_anml_reader is 18 passed locally.

alvalentini (approved):

PR #766, opened 2026-07-27
Build/tooling uv

alvalentini:
## Summary

Migrates the build and development tooling from `pip` / `setup.py` / `requirements*.txt` to **uv + just + hatchling**, with a committed lockfile, pre-commit hooks, Dependabot and ruff replacing black. CI is reworked around the same tools, and the dependency declarations are modernized on top: real lower bounds, Python 3.14 support, and a refreshed gRPC stack.

**21 commits, 227 files, +8423 / −3346.**

## How to review this

**193 of the 227 file touches are mechanical.** Two commits account for them, and both are safe to skim by message rather than by diff:

| commit | what | files |
|---|---|---|
| `f29b674f` | `style: apply ruff format` | 95 |
| `58b5e01e` | `chore: normalize trailing whitespace and end-of-file newlines` | 103 |

That leaves **34 files** with real content:

- **Added** — `pyproject.toml`, `uv.lock`, `.github/dependabot.yml`, `.github/workflows/pr.yml`, `.github/workflows/deps.yml`, `.github/actions/setup/action.yml`, `.github/actions/install-enhsp/action.yml`, `ci/stamp_dev_version.py`, `scripts/install_enhsp.py`
- **Removed** — `setup.py`, `MANIFEST.in`, `.mypy.ini`, `run_tests.sh`, `requirements.txt`, `dev-requirements.txt`, `plot-requirements.txt`, `docs/requirements.txt`, `scripts/generate_protobuf_bindings.sh`, `scripts/make_distrib.sh`, `.github/workflows/main.yml`
- **Modified** — `justfile`, `.pre-commit-config.yaml`, `.github/workflows/test.yml`, `.github/workflows/release.yml`, `.readthedocs.yaml`, `docs/conf.py`, `docs/contributor_guide.rst`, `unified_planning/__init__.py`, `unified_planning/io/{utils,anml_grammar,up_pddl_reader}.py`, `unified_planning/engines/compilers/tarski_grounder.py`, `unified_planning/grpc/generated/unified_planning_pb2{,_grpc}.py`

## What changed

**Build and packaging**
- `setup.py` → `pyproject.toml`; build backend is hatchling. A committed `uv.lock` gives reproducible CI and dev environments, with `uv sync --frozen` everywhere.
- Version stays static in `pyproject.toml`; `ci/stamp_dev_version.py` derives a unique `dev` version for master builds.
- Engine extras are declared once and the `engines` extra now references them, instead of restating all ten.

**Dependencies**
- Every direct requirement now declares a lower bound. The bounds match what `uv.lock` already resolves, so a default install is unaffected — the point is to make the floors auditable and to give the new `lowest-direct` CI job something real to test.
- **Python 3.14 is supported** and added to the post-merge matrix. That is what drives the gRPC bump: grpcio and grpcio-tools move 1.66.2 → 1.76.0, with `unified_planning/grpc/generated/` regenerated against them, since 1.76.0 is the oldest release with wheels for every interpreter we claim to support. The generated code enforces its own floors, hence the matching `protobuf>=6.31.1,<7` and the exact pins asserted by `just check-protobuf-version`.
- **`z3-solver` is now a real dependency** rather than a dev-group one: pysmt needs an SMT solver for the contingent execution environment, so on a plain install that code path could not run.
- **The tarski grounder no longer needs a `gringo` binary on `PATH`.** The `tarski` extra pulls the clingo Python bindings, which tarski drives through the `gringo.py` wrapper it ships; a system-wide `gringo` is still honoured. This is what lets the Windows job exercise the tarski extra at all, and removes the `gringo` apt install from every Linux job.
- `pyparsing>=3.0.0` is declared, so `unified_planning/io/` drops the per-call-site 2.x/3.x branches and the hand-written `Located` element.

**Developer workflow**
- The `justfile` is the single entry point: `just check` / `lint` / `typecheck` / `test` / `clean`, plus `check-imports`, `test-snippets`, `test-colab` — the last three wrap commands CI previously ran inline, so they are now reproducible locally. `just setup-ci` and `just install-enhsp` reproduce a CI test environment, ENHSP jar included.
- The ENHSP build moved out of the composite action into `scripts/install_enhsp.py`, stdlib-only Python rather than shell because the justfile runs recipes through `cmd.exe` on Windows. The release tag lives in the justfile and the action passes it through, so the two cannot drift.
- ruff replaces black (formatting only; lint rules deliberately stay off for now, see the note in `pyproject.toml`).
- pytest options moved into `pyproject.toml`, so a bare `pytest` behaves like `just test`. Previously a bare `pytest` aborted with `INTERNALERROR`, because it collected `scripts/test_imports.py` and hit that script's module-level `sys.exit()`.
- Locally, pre-commit runs the hygiene hooks plus `just lint` on commit, and `just typecheck` on **push**: types cross module boundaries, so a per-file mypy is meaningless, and blocking work-in-progress commits on a type error only teaches everyone `--no-verify`.

**CI**
- Two composite actions (`setup`, `install-enhsp`) remove the duplicated setup across all 7 jobs; `test.yml` drops from 452 to 222 lines. ENHSP's compiled build is cached, keyed on the runner OS and on the commit its tag resolves to.
- Concurrency groups: PR runs cancel superseded runs; master runs queue, which stops the rolling `dev` release job racing itself.
- `permissions: contents: read` by default with per-job escalation, and `secrets: inherit` replaced by an explicit `CODECOV_TOKEN`.
- Every dependency now comes from the lock, the planner engines included — `up-pyperplan`, `up-fast-downward`, `up-enhsp` and `up-fmap` were previously checked out at hard-coded SHAs that were invisible to Dependabot and drifted from the specifiers in `pyproject.toml`.
- The test jobs call `just setup-ci` instead of each carrying its own `uv sync --frozen --extra …` line, so the extra lists live next to the recipes that consume them.
- `basic-check` runs `just lint` and `just typecheck` directly rather than through pre-commit — the same commands the hooks run, with one less cache to keep warm.
- **New `deps` workflow.** Every other workflow installs with `uv sync --frozen`, which tests exactly one point in the resolution space; these two jobs cover the ends the lock hides. `lowest-direct` resolves every direct requirement to its declared floor on 3.10 and runs the tests there, on every PR — a code change can start relying on an API newer than a declared floor, and that is what the job is there to catch. `canary` re-resolves against the newest releases on PyPI weekly on 3.10 and 3.14, which is what notices a breaking upstream release here rather than in a user's install.

**Releases and versioning**
- **Master pushes no longer publish to PyPI.** Previously `release.yml` ran on every push to `master` and its `deploy` job uploaded to PyPI unconditionally, so every merge consumed a PyPI version. Publishing now happens **only on a `v*` tag**, through trusted publishing, and a matching GitHub release with generated notes is created alongside it.
- **Master pushes instead refresh a rolling `dev` GitHub pre-release** with the freshly built wheel and sdist. Master always has an installable artifact, without touching PyPI.
- **The version is no longer computed at runtime.** `unified_planning/__init__.py` used to shell out to `git describe --tags --dirty=-wip` on import and rewrite `VERSION` / `__version__` into `.devN` / `.postN` forms. It is now a single static `version` in `pyproject.toml`; `ci/stamp_dev_version.py` stamps `<base>.dev<commit-count>+g<sha>` into an uncommitted copy for master builds only, so each dev build is uniquely identifiable and ordered. That `+g<sha>` local segment also makes dev builds structurally unpublishable to PyPI, which rejects local versions — the two changes reinforce each other.

## Consequences for consumers

- `unified_planning.__version__` and `unified_planning.VERSION` no longer exist.
- Installing `unified-planning` now also installs `z3-solver`.
- `pyparsing>=3.0.0` is required; the 2.x fallback paths are gone.
- The `grpc` extra requires `grpcio>=1.76.0` and a protobuf 6 runtime.
- Python 3.14 is supported; the tarski extra no longer needs a system `gringo`.


alvalentini:
@arbimo In this PR, I continued the work you started in #736. I opened a new PR to keep things simple. I'd really appreciate your feedback. :)

arbimo:
Hi @alvalentini and thanks for continuing this work!

On the surface, it looks fantastic and I think would be a great improvement in developer experience!

Regarding the CI and pyproject (which are the main two things in this PR), my main interrogation is with respect to versions of dependencies.

- The pyproject.toml does not specifies min/max version for the main dependencies (pyparsing, networkx, ...) which is probably a bit dangerous (and has bitten before). Note that this was already the case before this PR and so it is not a regression and we might want to deal with it at a later time.

- The CI still relies on hand-picked versions for (notably) pyperplan, enhsp, ... I think it would make sense to pick them from the uv dependencies, since we really want to test against the released versions (again, this was already the case but it may be a good time to revisit)
- there are many formatting changes, which makes the PR a bit hard to parse. I am definitely in favor of ruff for formatting and Ok with a "breaking change". I am less convinced about the end of line trimming as I do not see much added value and it may generate noise (CI fails -> oh no I forgot to trim end of line -> fix and push -> CI pass).
- the last thing and perhaps most important thing is that we should make sure that when releasing with the new build, we do not break anything for users that install, e.g., through `pip`. This may for instance occur if we have a typo in a dependency group or change dependencies.  I don't think we have tests covering that (they all operate on pre-setup environments and we can't pull anything from pypi until it is released)

Somewhat adjacent to the PR:

- related to the previous one, I see claude as a coauthor on the reformatting commits. I think it should be removed from those ones if indeed they are just the result of invoking the tool (having it featured as the one of the most active developers in the repo based on the number of line changes is not something I look forward to).
- not specific to this PR but we probably need a policy on how to acknowledge AI usage in the commits. Seeing a tool as a co-author is not something I am very fond of and I believe more a result of marketing from AI firms. (but we are not adding other tools (e.g. ruff) as co-authors and just mention them in the commit message. On the other hand, I like seeing at a glance when the commit was AI generated.


I have not run anything locally and probably won't have time because I am leaving very soon. My previous experience from the ealier PR was very positive though but I won't give an approve right away. My position is that if you and @mikand think this is the direction we want to follow, then getting it into master soon is probably a good idea: this is the kind of changes that only gets really tested when it is into master because that where developers live. But I would be careful with extra test before cutting a release from this new project definition (even though I haven't seen anything fishy).

alvalentini:
Thanks a lot for the detailed feedback @arbimo!

I completely agree with the first two points, and I'm already working on addressing them.

Regarding the formatting changes, my reasoning is that trimming trailing whitespace is enforced by pre-commit (just like ruff), so in practice it shouldn't generate much additional noise once everyone has pre-commit installed. Because of that, I'd be inclined to keep it.

For the release/testing point, I completely agree that we should be careful. Do you have any suggestions on how we could best verify that we don't break the installation experience for users before cutting a release?

And yes, you're absolutely right about the last two points. The Claude co-author was automatically added by the tool, but I agree it doesn't make much sense, so I'll remove it.

arbimo:
Hi @alvalentini,

Perfect, the plan sounds good to me.

> Regarding the formatting changes, my reasoning is that trimming trailing whitespace is enforced by pre-commit (just like ruff), so in practice it shouldn't generate much additional noise once everyone has pre-commit installed. Because of that, I'd be inclined to keep it.

I am not sure it is safe to assume that most people will have the pre-commit hooks installed. They are not installed by default and I suspect most one-off contributors would not have them enabled (I personally don't have it either, mostly because I often reclone the repo and do not think of enabling them).

If we are to enable "trimming withespace", they should probably be part of `just format` and `just lint` as well so they can be reproduced easily.
Also from a quick look, I don't think that `ruff` is enabled in hte 

> 
> For the release/testing point, I completely agree that we should be careful. Do you have any suggestions on how we could best verify that we don't break the installation experience for users before cutting a release?

No, sorry :/

> 
> And yes, you're absolutely right about the last two points. The Claude co-author was automatically added by the tool, but I agree it doesn't make much sense, so I'll remove it.

Perhaps a good alternative would be to have an `Assisted-by: ...` flag that maintains visibility will attributing authorship (that the solution is being pushed by several open-source projects).

alvalentini:
@arbimo 

> I am not sure it is safe to assume that most people will have the pre-commit hooks installed. They are not installed by default and I suspect most one-off contributors would not have them enabled (I personally don't have it either, mostly because I often reclone the repo and do not think of enabling them).

Ok, so to keep things simple, I propose that CI runs `just lint`, not the `pre-commit` hooks. That way the only thing CI enforces about formatting is what `just format` fixes. Note that `ruff` already handles trailing whitespace and end-of-file newlines for Python files, so nothing is lost there.

`pre-commit` then becomes an optional local convenience: you can install it with `uv run pre-commit install` if you want it, nothing breaks if you don't. On top of the ruff check it also trims whitespace in the non-Python files (markdown, yaml, rst, the PDDL fixtures), but only for whoever opts in — never as a CI gate.

> Also from a quick look, I don't think that ruff is enabled in hte

What do you mean here?

> Perhaps a good alternative would be to have an `Assisted-by: ...` flag that maintains visibility will attributing authorship (that the solution is being pushed by several open-source projects).

Good suggestion!

arbimo:
> Ok, so to keep things simple, I propose that CI runs `just lint`, not the `pre-commit` hooks. ...

 Sounds good to me 👍
> 
> > Also from a quick look, I don't think that ruff is enabled in hte
> 
> What do you mean here?

Nothing, just a left over from something deleted.

unified_planning/engines/factory.py
```diff
@@ -400,12 +400,12 @@ def configure_from_file(self, config_filename: Optional[str] = None):
 
             module_name = config.get(s, "module_name")
             assert module_name is not None, (
-                "Missing 'module_name' value in definition" "of '%s' engine" % name
+                "Missing 'module_name' value in definitionof '%s' engine" % name
```
mikand:
```suggestion
                "Missing 'module_name' value in definition of '%s' engine" % name
```

unified_planning/engines/factory.py
```diff
@@ -400,12 +400,12 @@ def configure_from_file(self, config_filename: Optional[str] = None):
...
             )
 
             class_name = config.get(s, "class_name")
             assert class_name is not None, (
-                "Missing 'class_name' value in definition" "of '%s' engine" % name
+                "Missing 'class_name' value in definitionof '%s' engine" % name
```
mikand:
```suggestion
                "Missing 'class_name' value in definition of '%s' engine" % name
```

unified_planning/io/utils.py
```diff
@@ -14,23 +14,16 @@
 #
 
 
-import pyparsing
```
mikand, line 17: Does this file make sense now? All the functions are just proxies of standard pyparsing APIs...

pyproject.toml
```diff
@@ -0,0 +1,151 @@
...
+name = "unified-planning"
+version = "1.3.0"
+description = "Unified Planning Framework"
+readme = "README.md"
+requires-python = ">=3.10"
+license = "Apache-2.0"
+authors = [{ name = "AIPlan4EU Project", email = "aiplan4eu@fbk.eu" }]
+keywords = ["planning", "logic", "STRIPS", "RDDL"]
+classifiers = [
+    "Development Status :: 3 - Alpha",
```
mikand: Maybe we are now more than alpha?

mikand (approved):
LGTM 🎉 
I left some minor comments, but this can be merged. I agree with @arbimo that we need to test it "in practice" before making a release.

PR #775, opened 2026-07-31
fix(compilers): simplify and prune preconditions of zero-parameter actions during grounding

DavideLusuardi:
## Summary
- `GrounderHelper.ground_action`'s zero-parameter shortcut clones the
  action instead of routing it through `create_action_with_given_subs`,
  since there's nothing to substitute. That also skipped
  `check_and_simplify_preconditions` / `check_and_simplify_conditions`: a
  zero-parameter action with a contradictory precondition (`b and not b`)
  survived grounding un-pruned, while the identical action with an added
  (unused) parameter was correctly dropped. Un-simplified preconditions
  (unflattened `And`, uncleared tautologies) had the same gap.
- Extended `normalize_ground_action` (renamed from
  `normalize_ground_action_effects`) to also run
  `check_and_simplify_preconditions` / `check_and_simplify_conditions` on
  the cloned action, mirroring `create_action_with_given_subs` exactly.
  Verified a zero-parameter `SensingAction` still comes through as a
  `SensingAction` with its `observed_fluents` intact.

### DIFF
--- a/unified_planning/engines/compilers/grounder.py
+++ b/unified_planning/engines/compilers/grounder.py
@@ -34,7 +34,7 @@ from unified_planning.model.problem_kind_versioning import LATEST_PROBLEM_KIND_V
 from unified_planning.engines.compilers.utils import (
     lift_action_instance,
     create_action_with_given_subs,
-    normalize_ground_action_effects,
+    normalize_ground_action,
     split_all_ands,
 )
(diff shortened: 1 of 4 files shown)


alvalentini (commented):
@DavideLusuardi Can we use `create_action_with_given_subs` with `subs={}` instead of `normalize_ground_action` ?

PR #785, opened 2026-08-12
fix: three latent correctness bugs (compiler closure, shared dataclass default, total-cost metric)

alvalentini:
## Problem

Three independent bugs, none of which any existing test catches. Each is fixed in its own commit with its own regression test, so they can be reviewed — or cherry-picked onto a patch release — separately.

**1. Rebuilt durative simulated effects all called the last one's function.**
`create_action_with_given_subs` rebuilds a `DurativeAction`'s simulated effects one timing at a time, and the closure it builds captured the loop variable rather than its value. That closure is stored in the `SimulatedEffect` and attached to the action, so it outlives the loop — by the time it runs, every timing invoked the last timing's function. The result is a silently wrong compiled problem on every path through this helper: grounder, quantifiers remover, disjunctive and conditional effects removers.

Two simulated effects at different timings are legal (`simulated_effects` is keyed by `Timing`, and `check_conflicting_simulated_effects` only compares within a single timing), but nothing in the suite used more than one. The only grounder test with simulated effects goes through the `InstantaneousAction` branch, which is safe because it is an `if` rather than a `for` — which is why two otherwise identical blocks behaved differently.

**2. All default-constructed `MethodInstance`s shared one `Decomposition`.**
`decomposition` defaulted to a single class-level `Decomposition()`, so every instance saw the same `subtasks` dict and mutating one leaked into all the others. Neither `MethodInstance` nor `Decomposition` had any test coverage.

**3. Three defects in `UPPDDLReader`'s total-cost handling.**
`use_plan_length` was assigned `False` in five places and read once, but never initialised — so a domain declaring `(:metric minimize (total-cost))` with no actions raised `UnboundLocalError` straight out of `parse_problem`. Separately, `cost.value` is an `FNode`, which leaves `__eq__` as identity, so `cost.value != 1` was never false: the flag was cleared for every action that had a cost at all, making `MinimizeSequentialPlanLength` unreachable. A domain whose actions all cost 1 parsed as `MinimizeActionCosts`.

## Fix

- `engines/compilers/utils.py`: bind `old_se` as a default argument. Renamed, since both closures share a scope and their signatures now differ (mypy `no-redef`).
- `plans/hierarchical_plan.py`: `field(default_factory=Decomposition)`.
- `io/up_pddl_reader.py`: initialise `use_plan_length = True`, and compare the constant via `is_int_constant()` / `constant_value()`.

With the reader fixed, a uniformly unit-cost domain now yields the plan-length metric, which is what the surrounding code was written to do. **No existing fixture changes behaviour**: `parking_action_cost` has no instantaneous actions, `citycar`'s costs are not all 1, and `tpp_metric`'s are expressions. It also makes the metric survive a PDDL round trip, since `PDDLWriter` already emits unit increase effects for it.

## Test plan

- [x] 5 new tests, each verified to **fail against the pre-fix sources** and pass after — checked by loading the parent commit's version of each file through an import hook, so they are genuine regression guards rather than tautologies
- [x] `just test` — 527 passed, 3 skipped (unchanged)
- [x] `just lint`
- [x] `just typecheck`
- [x] pre-commit hooks on each commit
- [x] CI

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/unified_planning/engines/compilers/utils.py
+++ b/unified_planning/engines/compilers/utils.py
@@ -304,14 +304,16 @@ def create_action_with_given_subs(
             for f in old_se.fluents:
                 new_fluents.append(f.substitute(subs))
 
-            def fun(_problem, _state, _):
-                return old_se.function(_problem, _state, c_subs)
+            # _old_se bound as a default: the closure outlives the iteration, so capturing
+            # the loop variable would make every timing call the last one's function.
+            def durative_fun(_problem, _state, _, _old_se=old_se):
+                return _old_se.function(_problem, _state, c_subs)
 
             # this rebuilds a simulated effect the user already defined (and got
             # warned about), so the deprecation warning is silenced here
             with warnings.catch_warnings():
                 warnings.simplefilter("ignore", DeprecationWarning)
-                new_simulated_effect = SimulatedEffect(new_fluents, fun)
+                new_simulated_effect = SimulatedEffect(new_fluents, durative_fun)
             # We try to add the new simulated effect, but a compiler might generate conflicting effects,
             # so the action is just considered invalid
             try:
--- a/unified_planning/engines/compilers/utils.py
+++ b/unified_planning/engines/compilers/utils.py
@@ -304,10 +304,11 @@ def create_action_with_given_subs(
             for f in old_se.fluents:
                 new_fluents.append(f.substitute(subs))
 
-            # _old_se bound as a default: the closure outlives the iteration, so capturing
-            # the loop variable would make every timing call the last one's function.
-            def durative_fun(_problem, _state, _, _old_se=old_se):
-                return _old_se.function(_problem, _state, c_subs)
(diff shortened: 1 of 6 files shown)


unified_planning/engines/compilers/utils.py
```diff
@@ -304,14 +304,16 @@ def fun(_problem, _state, _):
             for f in old_se.fluents:
                 new_fluents.append(f.substitute(subs))
 
-            def fun(_problem, _state, _):
-                return old_se.function(_problem, _state, c_subs)
+            # _old_se bound as a default: the closure outlives the iteration, so capturing
+            # the loop variable would make every timing call the last one's function.
+            def durative_fun(_problem, _state, _, _old_se=old_se):
```
mikand: Maybe we can use a more descriptive name than `_old_se` for the parameter? Like `_inner_simulated_effect` or similar?

mikand (approved):
LGTM

PR #798, opened 2026-08-20
perf: speed up grounding, name lookups, and expression promotion

DavideLusuardi:
## Summary
- Index `*SetMixin` name lookups (`NameIndex`) so building a problem is no longer O(N^2).
- Fast-path `FNode` in expression promotion and cache `FNode.__repr__`'s dispatch table.
- Add hash-join pruning of action parameters against static fluents, replacing per-parameter-only pruning for the common case, and fix it to intersect every matching argument position of a static fluent instead of just the first.
- Skip deep-cloning actions in `Grounder._compile` when they're immediately discarded, and skip deep-cloning a candidate grounded action's effects until after its preconditions are known to be feasible.

## Hash-join static-fluent pruning

**Goal:** given one action and the type domains of its parameters, compute the exact list
of parameter tuples that satisfy all of the action's static-boolean-fluent conditions --
via a relational hash-join instead of the existing per-parameter (independent) pruning.
When the join isn't safe for an action (unbounded/unenumerable parameter type, a candidate
count above a safety cap, an unrecognized action type, or any unexpected exception), that
one action falls back to the existing pruning, so the join is strictly non-regressive.

**High-level shape:** it's a classic database join. Each static fluent in the action's
conditions is a "table" of rows -- the argument tuples for which that fluent is true -- and
the algorithm joins these tables together on whatever action-parameter columns they share,
incrementally narrowing a running result set. Parameter columns no fluent ever constrains
are left free and cross-produced in only at the end.

**Why the old pruning misses this:** it narrows each parameter's domain independently to
the objects seen at a matching argument position of some true static-fluent atom. That's
blind to a static fluent that correlates several parameters together (e.g. a lookup table
relating two of them): every value that's valid for a parameter *somewhere* survives, even
when no single combination of survivors is jointly valid, so the full cross product still
gets enumerated (and then rejected later, when preconditions are checked).

**Example:** action `deliver(?r: Robot, ?l: Location, ?p: Package)` with domains
`Robot={r1,r2}`, `Location={l1,l2}`, `Package={p1,p2}` (8 combinations, unpruned) and static
conditions `robot_can_carry(?r,?p)` true for `(r1,p1), (r1,p2), (r2,p1)`, and
`package_at(?p,?l)` true for `(p1,l1), (p2,l2)`.

Per-parameter pruning can't shrink anything here: `r1`, `r2`, `l1`, `l2`, `p1`, `p2` each
appear in at least one true tuple, so all 8 combinations -- including invalid ones like
`(r2, l2, p2)`, since r2 never carries p2 -- would still be cross-produced.

The join instead folds in `robot_can_carry` first, producing partial bindings
`{r:r1,p:p1}, {r:r1,p:p2}, {r:r2,p:p1}`; then folds in `package_at`, joining on their
shared `p` column (`p1` only pairs with `l1`, `p2` only with `l2`), leaving exactly
`(r1,l1,p1), (r1,l2,p2), (r2,l1,p1)` -- the 3 combinations that are jointly consistent with
both conditions.

**Along the way:** the existing (fallback) pruning is also fixed to intersect *every*
matching argument position of a static fluent instead of stopping at the first -- needed
for a symmetric predicate applied as `sym(?x, ?x)` -- and static-fluent/domain lookups are
cached, since both pruning paths were otherwise recomputing them per action/parameter.

### DIFF
--- a/unified_planning/engines/compilers/grounder.py
+++ b/unified_planning/engines/compilers/grounder.py
@@ -36,7 +36,7 @@ from unified_planning.engines.compilers.utils import (
     create_action_with_given_subs,
     split_all_ands,
 )
-from typing import Dict, List, Optional, Set, Tuple, Iterator, cast
+from typing import Any, Dict, List, Optional, Set, Tuple, Iterator, cast
 from itertools import product
 from functools import partial
 
@@ -73,7 +73,17 @@ class GrounderHelper:
             * `b (o3)`
             * `b (o4)`
             If this map is `None`, the `unified_planning` grounding algorithm is applied.
-        :param prune_actions: If true, the grounder prunes actions exploiting the simplification of static fluents.
+        :param prune_actions: If `True` (the default), actions are pruned exploiting the simplification of
+            static fluents: the action's eligible static fluents are hash-joined together (correctly
+            handling hierarchical typing), so a static predicate that jointly constrains several of the
+            action's parameters together (e.g. a lookup table relating them) is pruned down to the handful
+            of parameter tuples actually consistent with it, rather than enumerating (and then rejecting)
+            their full cross product. If the join isn't applicable/safe for a given action -- an unbounded
+            parameter type, a candidate count above an internal safety cap, or any unexpected condition --
+            that one action falls back to the weaker (but always sound) per-parameter-independent pruning:
+            each parameter is narrowed, independently of the others, to the objects that appear at a
+            matching argument position of some true static fluent in the action's precondition. If
+            `False`, no pruning at all is done.
         """
         assert isinstance(problem, Problem)
         self._problem = problem
@@ -90,13 +100,23 @@ class GrounderHelper:
         # When the grounded instance of the Action is None, it means that the resulting grounding
         # of that action is a meaningless Action.
         # An Action is meaningless when:
-        # - it has no effects
         # - his conditions create a contradiction
         # - the action has conflicting effects
         # - for a DurativeAction, its duration interval simplifies to an empty interval
+        # Note: an action with no effects is NOT rejected; it is grounded as-is.
         self._grounded_actions: Dict[
             Tuple[str, Tuple[FNode, ...]], Optional[Action]
         ] = {}
+        # Caches reused across every lifted action grounded by this instance. The problem is
+        # documented as immutable for the lifetime of this class (grounding relies on it, same
+        # as Simplifier's own "problem must not be modified" contract below), so these are safe
+        # to compute once instead of once per action / per (parameter, static-fluent) pair.
+        self._static_fluents_cache: Optional[Set["up.model.fluent.Fluent"]] = None
+        self._true_arg_tuples_cache: Dict[
+            "up.model.fluent.Fluent", List[Tuple[FNode, ...]]
+        ] = {}
+        self._domain_items_cache: Dict[Type, List[FNode]] = {}
+        self._valid_params_cache: Dict[Tuple[FNode, int], Set[FNode]] = {}
         env = problem.environment
         if prune_actions:
             self._simplifier = Simplifier(env, problem)
@@ -167,7 +187,8 @@ class GrounderHelper:
         * The ``grounded_action`` is the `Action` created by grounding the ``original_action`` with the given ``parameters``;
             the ``grounded_action`` can be ``None`` if the grounding of the ``original_action`` with the given parameters
             creates an invalid or meaningless `Action` (invalid if it has conflicting `Effects`,
-            meaningless if it has no `effects` or contradicting `conditions`).
+            meaningless if its conditions create a contradiction, or, for a `DurativeAction`,
+            if its duration interval simplifies to an empty interval).
         """
         for old_action in self._problem.actions:
             for grounded_params in self.get_possible_parameters(old_action):
@@ -205,52 +226,22 @@ class GrounderHelper:
                 # objects_list = [[r1, r2], [l1, l2]]
                 # the product of *objects_list will be:
                 # [(r1, l1), (r1, l2), (r2, l1), (r2,l2)]
-                ground_size = 1
-                domain_sizes = []
-                for t in type_list:
-                    ds = domain_size(self._problem, t)
-                    domain_sizes.append(ds)
-                    ground_size *= ds
-                items_list: List[List[FNode]] = []
-                for size, type in zip(domain_sizes, type_list):
-                    items_list.append(
-                        [domain_item(self._problem, type, j) for j in range(size)]
-                    )
+                if self._prune_actions:
+                    joined = self._compute_join_pruned_parameters(action, type_list)
+                    if joined is not None:
+                        return iter(joined)
+                    # else: the join wasn't applicable/safe for this action -- fall through
+                    # to the per-parameter-independent cross-product path below, for this
+                    # action only.
 
-                problem_static_fluents = self._problem.get_static_fluents()
-                if self._prune_actions and isinstance(
-                    action, up.model.action.InstantaneousAction
+                items_list: List[List[FNode]] = [
+                    self._get_domain_items(t) for t in type_list
+                ]
+                if self._prune_actions and (
+                    isinstance(action, up.model.action.InstantaneousAction)
+                    or isinstance(action, up.model.action.DurativeAction)
                 ):
-                    no_and_list = split_all_ands(action.preconditions)
-                    bool_conditions = []
-                    for c in no_and_list:
-                        if (
-                            c.is_fluent_exp()
-                            and c.fluent().type.is_bool_type()
-                            and c.fluent() in problem_static_fluents
-                        ):
-                            bool_conditions.append(c)
-                    items_list = self._purge_items_list(
-                        items_list=items_list,
-                        params=action.parameters,
-                        conds=bool_conditions,
-                    )
-                elif self._prune_actions and isinstance(
-                    action, up.model.action.DurativeAction
-                ):
-                    condlist = []
-                    for _, cl in action.conditions.items():
-                        condlist.extend(cl)
-
-                    no_and_list = split_all_ands(condlist)
-                    bool_conditions = []
-                    for c in no_and_list:
-                        if (
-                            c.is_fluent_exp()
-                            and c.fluent().type.is_bool_type()
-                            and c.fluent() in problem_static_fluents
-                        ):
-                            bool_conditions.append(c)
+                    bool_conditions = self._static_bool_fluents(action) or []
                     items_list = self._purge_items_list(
                         items_list=items_list,
                         params=action.parameters,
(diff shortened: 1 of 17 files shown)


alvalentini (commented):
Reviewed all 6 commits at `fe69aa9`. The performance work is solid, and the fallback-by-construction design of the join is the right call: the per-action bail-out to the existing pruning means a bad join can never produce a wrong grounding. Full test suite green locally (532 passed, 3 skipped, 63 subtests).

Three things I'd want addressed before merge:

1. **`name_index.py`** -- the stated rationale for the in-place-rename gap doesn't hold; `has_action`/`action(name)` do regress relative to master for that case. The gap itself may well be acceptable, but the justification (and the `test_renamings` reference) needs correcting.
2. **`grounder.py`** -- `_JOIN_MAX_CANDIDATES` is checked only after the candidate set has been fully materialised, so it doesn't bound peak memory.
3. **`grounder.py`** -- the two `# TODO: can we enter here?` markers sit on correctness guards that tests in this PR already exercise.

Plus seven non-blocking items inline. The most substantive is that `force_fallback_only` in `test_grounder.py` doesn't force the fallback when the join collapses to zero rows, which weakens the equivalence test the whole non-regression argument rests on.

Leaving this as a comment rather than a change request.

unified_planning/model/mixins/name_index.py
```diff
@@ -0,0 +1,106 @@
...
+class NameIndex(Generic[T]):
+    """A lazily-rebuilt ``name -> element`` index over a list of named elements.
+
+    The index is guarded by an identity+length token against the list it was built from, so
+    it self-heals (at the cost of one O(n) rebuild) against every way the underlying list can
+    change without going through :func:`note_appended`: wholesale reassignment (every
+    ``clone()``/``_clone_to`` path in this library replaces the list outright) or in-place
+    removal (``list.remove(...)``, used to drop a fluent). The one case this does not catch is
+    renaming an already-indexed element in place without changing the list's identity or
+    length (`Action.name` has a public setter) -- that was already unguarded before this index
```
alvalentini: **The stated rationale for the in-place-rename gap doesn't hold.**

I'm not arguing against accepting the gap, but this sentence is what a future reader will rely on, and it's checkably false. Before this PR there was no cache at all: the linear scan in `has_action`/`action(name)` always saw the element's current name. Checked against `master`:

```python
pb.has_action("a")              # populates the index
pb.action("a").name = "zzz"     # in-place rename

pb.has_action("zzz")   # master: True   -> this branch: False
pb.has_action("a")     # master: False  -> this branch: True
pb.action("zzz")       # this branch: raises UPValueError
```

So two public methods change their answer for a scenario master handled correctly. It's narrow -- every rename inside the library happens before `add_action`, so only user code is exposed -- but it is a regression. Whether and how to close it is your call; the comment should just say "accepted regression because X" rather than "not a regression".

unified_planning/test/test_problem.py
```diff
@@ -654,6 +662,148 @@ def test_interpreted_functions_simple(self):
...
+        # renames an action after reading it via problem.action(...), and a subsequent
+        # add_object call must not treat the rename as a fresh collision. Documented here so
+        # a future change to this behavior is deliberate, not a silent regression.
+        renamed = problem.clone()
+        target = renamed.actions[0]
+        old_name = target.name
+        self.assertTrue(renamed.has_action(old_name))  # force the index to build now
+        target.name = "renamed_action"
+        self.assertTrue(renamed.has_action(old_name))  # stale: rename isn't tracked
+        self.assertFalse(renamed.has_action("renamed_action"))
```
alvalentini: **The `test_renamings` reference doesn't support this.**

In `test_pddl_io.py::test_renamings` the action is renamed to `move-move` and the object added afterwards is named `block-4`. There is no name collision under either the old or the new behaviour, and that test passes both ways -- so it isn't relying on the stale index. The reference should be dropped, or replaced with one that actually exercises the gap.

(See also my comment on `name_index.py`: the "already unguarded before this index existed" claim doesn't hold either -- master's linear scan did see the new name.)

unified_planning/engines/compilers/grounder.py
```diff
@@ -262,6 +253,392 @@ def get_possible_parameters(self, action: Action) -> Iterator[Tuple[FNode, ...]]
...
+                    bindings = [
+                        {**left, **right} for left in bindings for right in rows
+                    ]
+                bound_columns |= fluent_columns
+
+            if not bindings:
+                # Zero rows survived: no combination can ever satisfy every fluent folded in
+                # so far, so no later fluent can matter either -- stop early.
+                break
+            if len(bindings) > self._JOIN_MAX_CANDIDATES:
```
alvalentini: **`_JOIN_MAX_CANDIDATES` doesn't bound peak memory -- the cap is checked after materialisation.**

Both merge branches build their full result before line 503 tests the cap:

- hash-join: `joined_rows` accumulates unbounded;
- cross-product: the whole `[{**left, **right} for left in bindings for right in rows]` comprehension is built at once.

With `bindings` near the cap and a fluent contributing many rows, the intermediate can exceed the cap by orders of magnitude before the check fires -- so the safety valve doesn't guard the failure mode it exists for.

The correct shape is already used for the free-column expansion further down (`len(bindings) * free_size > self._JOIN_MAX_CANDIDATES` -- arithmetic, evaluated before anything is built). The same applies here:

- cross-product: pre-check `len(bindings) * len(rows) > cap` before the comprehension;
- hash-join: break out of the accumulation loop once `len(joined_rows)` exceeds the cap.

unified_planning/engines/compilers/grounder.py
```diff
@@ -262,6 +253,392 @@ def get_possible_parameters(self, action: Action) -> Iterator[Tuple[FNode, ...]]
...
+            # pattern; each one that's consistent becomes one candidate binding *for this
+            # fluent alone* (not yet folded together with any other fluent's bindings).
+            rows: List[Dict[int, FNode]] = []
+            for true_args in self._fluent_true_arg_tuples(static_fluent.fluent()):
+                binding: Dict[int, FNode] = {}
+                is_consistent = True
+                for (kind, payload), arg_value in zip(pattern, true_args):
+                    if kind == "c":
+                        if arg_value != payload:
+                            # TODO: can we enter here?
```
alvalentini: **This TODO is answered by a test in this PR.**

The constant-position mismatch branch is reachable, and `test_join_constant_argument` covers it: the pattern is `at(?x, l1)` while `at` is also true for `(i1, l2)`, so position 1 compares `l2` against the constant `l1` and rejects the tuple.

Worth replacing the TODO with a pointer to that test -- an open "is this reachable?" question sitting on a correctness guard invites someone to remove the guard.

unified_planning/engines/compilers/grounder.py
```diff
@@ -262,6 +253,392 @@ def get_possible_parameters(self, action: Action) -> Iterator[Tuple[FNode, ...]]
...
+                            # supertype of the action parameter it's being matched against.
+                            # Omitting this intersection would bind an object that's ill-typed for the action
+                            # parameter.
+                            is_consistent = False
+                            break
+                        # The same parameter appearing at more than one position of this one
+                        # fluent (e.g. `sym(?x, ?x)`) must agree with itself at every
+                        # occurrence, not just be individually well-typed at each one.
+                        if col in binding and binding[col] != arg_value:
+                            # TODO: can we enter here?
```
alvalentini: **Same here -- this TODO is answered by a test in this PR.**

`test_static_bool_condition_prunes_on_every_matching_position` reaches it: the pattern for `sym(?x, ?x)` is `[("p", 0), ("p", 0)]`, and against the true tuple `sym(o0, o1)` column 0 is bound to `o0` and then re-bound to `o1`. This check is precisely what makes `sym(x, x)` correctly unsatisfiable.

unified_planning/engines/compilers/grounder.py
```diff
@@ -262,6 +253,392 @@ def get_possible_parameters(self, action: Action) -> Iterator[Tuple[FNode, ...]]
                 res = iter(self._grounding_actions_map.get(action, []))
         return res
 
+    # Safety valve for the join: bail out to the per-parameter cross-product fallback for
+    # one action if its running candidate count exceeds this. The reference dataset this
+    # algorithm was validated against never got close to it (largest was ~52k).
+    _JOIN_MAX_CANDIDATES = 1_000_000
```
alvalentini: **Consider making the cap a constructor parameter.**

As a class attribute, the only way to vary it is to mutate class state globally -- which `_ground_names` in `test_grounder.py` does, with a `finally` restore. That isn't thread-safe if the suite ever runs in parallel, and it leaves a user with a pathological domain no way to lower the cap without patching the library.

`GrounderHelper(problem, ..., join_max_candidates=1_000_000)` would let the tests pass it explicitly instead.

unified_planning/engines/compilers/grounder.py
```diff
@@ -262,6 +253,392 @@ def get_possible_parameters(self, action: Action) -> Iterator[Tuple[FNode, ...]]
...
+
+    def _compute_join_pruned_parameters(
```
alvalentini, line 303: **Distinguish an expected bail-out from an unexpected exception.**

The documented bail-outs (unrecognised action type, cap exceeded) can reasonably stay silent, but a bare `except Exception` makes an internal bug indistinguishable from them: a typo in a helper degrades to the per-parameter fallback with no signal anywhere.

```python
try:
    return self._compute_join_pruned_parameters_unsafe(action, type_list)
except Exception:
    warn(f"static-fluent join failed for {action.name}, "
         "falling back to per-parameter pruning", UserWarning)
    return None
```

`test_join_unsafe_actually_prunes_and_does_not_silently_fail` is the right mitigation and shows the risk was understood, but it only covers that one shape.

unified_planning/engines/compilers/grounder.py
```diff
@@ -262,6 +253,392 @@ def get_possible_parameters(self, action: Action) -> Iterator[Tuple[FNode, ...]]
...
+        cached = self._domain_items_cache.get(type)
+        if cached is None:
+            cached = [
+                domain_item(self._problem, type, j)
+                for j in range(domain_size(self._problem, type))
+            ]
+            self._domain_items_cache[type] = cached
+        return cached
+
+    def _fluent_true_arg_tuples(
```
alvalentini, line 580: **Worth a peak-memory number for this cache.**

For a static fluent with a `True` default this materialises every grounding (a binary predicate over N objects is N**2 tuples) and holds it for the lifetime of the `GrounderHelper`. Master recomputed -- worse in time -- but retained nothing. Could you add a before/after peak-memory figure on a large problem? Same question for `_domain_items_cache` and `_valid_params_cache`.

unified_planning/engines/compilers/grounder.py
```diff
@@ -274,39 +651,34 @@ def _purge_items_list(
...
-        if default_value is not None and default_value.is_false():
-            # if default is false, check only explicit instead of all values
-            for key, value in self._problem.explicit_initial_values.items():
-                if key.fluent() == sf.fluent() and value.is_true():
-                    ret_val.add(key.args[sp])
-        else:
-            for key, value in self._problem.initial_values.items():
-                if key.fluent() == sf.fluent() and value.is_true():
-                    ret_val.add(key.args[sp])
+        cache_key = (sf, sp)
```
alvalentini: **Cache key is over-specified.**

`cache_key` is `(sf, sp)` where `sf` is the full atom, but the body below only depends on `sf.fluent()` and `sp`. Keying on `(fluent, sp)` gives strictly more hits -- across different atoms over the same fluent, and across actions -- with a smaller key.

unified_planning/test/test_grounder.py
```diff
@@ -885,3 +980,262 @@ def test_pyperplan_grounder_mockup_problem(self):
...
+    check every candidate still goes through regardless of which one produced it.
+    """
+
+    def setUp(self):
+        unittest_TestCase.setUp(self)
+        self.problems = get_example_problems()
+
+    @staticmethod
+    def _ground_names(problem, force_fallback_only=False):
+        original_cap = GrounderHelper._JOIN_MAX_CANDIDATES
```
alvalentini: **`force_fallback_only` doesn't actually force the fallback in all cases.**

Setting the cap to `0` fails whenever the join collapses to zero rows, because of the order of the two checks in the fold loop: `if not bindings: break` fires first, execution leaves the loop, and the final guard computes `len([]) * free_size == 0`, which does not exceed `0`. The function then returns `[]` **from the join**, not from the fallback.

So for exactly the actions where the join prunes hardest, `test_join_equivalent_to_fallback_pruning_on_example_problems` compares join-with-cap-0 against join, rather than join against the fallback. Since that test carries the non-regression argument for the whole feature, I'd make it airtight: either add an explicit `_force_fallback` flag, or swap the order of the two checks so the cap is evaluated before the early break.

(This is independent of moving the cap into the constructor -- that alone wouldn't fix it.)

unified_planning/test/test_grounder.py
```diff
@@ -885,3 +980,262 @@ def test_pyperplan_grounder_mockup_problem(self):
...
+        gh = GrounderHelper(problem)
+        type_list = [p.type for p in action.parameters]
+        result = gh._compute_join_pruned_parameters_unsafe(action, type_list)
+        assert result is not None
+        self.assertEqual(
+            set(result),
+            {rel(items[0], items[1]).args, rel(items[2], items[3]).args},
+        )
+
+    def test_join_equivalent_to_fallback_pruning_on_example_problems(self):
```
alvalentini, line 1196: **This test passing establishes that no example problem exercises the join's advantage.**

The assertion is `joined_names == fallback_names` across every example problem, so if a single one benefited from the join the two lists would differ and this would fail. That's a valuable non-regression check, but it also means the interesting path has unit coverage on hand-written synthetic problems only.

Worth adding one example problem with a static predicate that jointly constrains two action parameters, so the join also goes through the integration paths (writers, plan validation, round-trip).

unified_planning/model/problem.py
```diff
@@ -249,14 +249,30 @@ def __hash__(self) -> int:
         return res
 
     def clone(self):
+        new_p = self._clone_without_actions()
+        new_p._actions = [a.clone() for a in self._actions]
```
alvalentini, line 263: **This also drops quality metrics, which the name doesn't say.**

The grounder hides it by calling `clear_quality_metrics()` immediately afterwards and rebuilding them, so it works here -- but the next caller won't necessarily know. Since the rollout to the other compilers lives in a separate PR, this matters more rather than less: consider renaming (`_clone_without_actions_and_metrics`) or asserting the contract, and double-checking that every new caller there handles metrics.

alvalentini (approved):

